PR #13, opened 2015-09-22
Use fixed-sized arrays

tomaka:
Causes linking error with the MSVC nightly.
Wait for https://github.com/rust-lang/rust/issues/27438 to be fixed before merging.


tomaka:
Too outdated.

PR #48, opened 2015-10-17
Implement decoding mp3

tomaka:
cc @bendykst

Need to change the library to use crates.io before merging.


tomaka:
I didn't notice at first, but the example sound stops playing because a `Err(Mad(BufLen))` is triggered.

tomaka:
Hum. Libmad is under GPL, so this code is not legal.

emersion:
Note that it's unclear whether or not linking to libmad is allowed: https://en.wikipedia.org/wiki/GNU_General_Public_License#Libraries

Also, you're not using libmad directly, but `simple-mad` which is [licensed under MIT](https://github.com/bendykst/simple-mad.rs/blob/master/LICENSE).

mhsjlw:
Any word on this?

tomaka:
The inner source code is still GPL: https://github.com/bendykst/simple-mad.rs/blob/master/simplemad_sys/libmad-src/COPYRIGHT

est31:
Can this be closed, now that rodio has mp3 support thanks to minimp3?

PR #62, opened 2015-10-26
Deny unsafe code in the library

tomaka:


tomaka:
`lazy_static` doesn't work with this :-/

PR #65, opened 2016-01-15
bump cpal dependency version

psych0d0g:
include latest cpal lib


tomaka:
Note that this doesn't change anything, as Cargo will pick version 0.2.8 even if you request 0.2.7.

psych0d0g:
then we should change rodio version number aswell so cargo can pick up the changed version

PR #75, opened 2016-09-04
Fix a private-in-public error

petrochenkov:
Not sure if this crate is abandoned or not, but it was broken by a bugfix in rustc (see https://tools.taskcluster.net/task-inspector/#VPC0Q4iFQKO1tX2H-Ubm1w/0 and rust-lang/rust#34537 for more details).
Here's a fix (I also fixed one deprecation warning).


tomaka:
Thanks

PR #77, opened 2016-10-04
Use lewton

### LINKED ISSUE #74: Switch to lewton
https://github.com/est31/lewton

est31:
Fixes #74.


src/decoder/vorbis.rs
```diff
@@ -4,34 +4,42 @@ use std::vec;
...
 
-        let (data, rate, channels) = match decoder.packets().filter_map(Result::ok).next() {
-            Some(p) => (p.data, p.rate as u32, p.channels as u16),
-            None => (Vec::new(), 44100, 2),
+        let mut data = match stream_reader.read_dec_packet_itl().ok().and_then(|v| v) {
+            Some(d) => d,
+            None => Vec::new(),
+        };
+
+        match stream_reader.read_dec_packet_itl().ok().and_then(|v| v) {
```
tomaka: Why call `read_dec_packet_itl()` three times? Is it an arbitrary choice?

src/decoder/vorbis.rs
```diff
@@ -44,12 +52,12 @@ impl<R> Source for VorbisDecoder<R> where R: Read + Seek {
 
     #[inline]
     fn get_channels(&self) -> u16 {
-        self.current_channels
+        self.stream_reader.ident_hdr.audio_channels as u16
```
tomaka: I thought that in the vorbis format, it was possible that the number of channels and samples rate changes between two frames.

src/decoder/vorbis.rs
est31: Ah sorry, that was left from when I was debugging. Will remove

src/decoder/vorbis.rs
est31: No, the number of channels is constant over the whole stream: https://www.xiph.org/vorbis/doc/Vorbis_I_spec.html#x1-120001.2.1

tomaka:
Thanks!

PR #81, opened 2017-01-23
Fix volume, dropping of Sinks, and add play/pause functions to Sinks

Xaeroxe23:
In addition to the changes mentioned in the title I ran rustfmt on the files I edited to make the formatting consistent and easier to read.


Xaeroxe23:
With the large amount of files changed by rustfmt I realize this pull request could probably get annoying to read quickly, so there were 2 files the real work of this request was focused in, lib.rs, and engine.rs.  everything else is just rustfmt changes.

tomaka:
Err.. you should submit one PR per change.
Unfortunately I won't accept a PR that does changes and runs rustfmt at the same time. Running rustfmt on the code isn't urgent, it can always be done later.

Xaeroxe23:
Understandable, sorry I got a little carried away

PR #82, opened 2017-01-23
Play pause

Xaeroxe23:
Enhances the sink with ability to pause and resume playback of the current source queue.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -35,7 +35,14 @@ pub struct Engine {
 struct EndPointVoices {
     format: Format,
     next_id: AtomicUsize,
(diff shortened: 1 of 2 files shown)


src/engine.rs
```diff
@@ -35,7 +35,15 @@ pub struct Engine {
 struct EndPointVoices {
     format: Format,
     next_id: AtomicUsize,
-    pending_sounds: Mutex<Vec<(usize, QueueIterator)>>,
+    pending_sounds: Mutex<Vec<HandleHandle>>,
+}
+
+// Handle to a handle I guess.
```
tomaka: Same two remarks as #84

PR #83, opened 2017-01-23
Volume, play/pause, and fix sink dropping

Xaeroxe23:
Fixed the set_volume function of a Sink and adds a get_volume function.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -1,8 +1,7 @@
 use std::collections::HashMap;
 use std::thread::Builder;
 use std::sync::mpsc::{self, Sender, Receiver};
-use std::sync::atomic::AtomicUsize;
-use std::sync::atomic::Ordering;
+use std::sync::atomic::{AtomicUsize, AtomicBool, Ordering};
 use std::sync::Arc;
 use std::sync::Mutex;
 
@@ -38,16 +37,15 @@ struct EndPointVoices {
     pending_sounds: Mutex<Vec<HandleHandle>>,
 }
 
-// Handle to a handle I guess.
-// The tuple type was getting a bit unwieldy so I added this.
+// Handle to a handle
 struct HandleHandle {
     handle_id: usize,
     queue_iterator: QueueIterator,
     volume: Arc<Mutex<f32>>,
 
-    paused: Arc<Mutex<bool>>,
+    paused: Arc<AtomicBool>,
 
-    dead: Arc<Mutex<bool>>,
+    dead: Arc<AtomicBool>,
 }
 
 impl Engine {
@@ -125,10 +123,10 @@ impl Engine {
                 if sounds.len() == 0 {
                     return Ok(());
                 }
-                sounds.retain(|s| !*s.dead.lock().unwrap());
+                sounds.retain(|s| !s.dead.load(Ordering::Relaxed));
                 let samples_iter = (0..).map(|_| {
                     sounds.iter_mut().filter_map(|s| {
-                        if *s.paused.lock().unwrap() {
+                        if s.paused.load(Ordering::Relaxed) {
                             return None;
                         }
                         Some(s.queue_iterator.next().unwrap_or(0.0) * (*s.volume.lock().unwrap()))
@@ -162,10 +160,10 @@ impl Engine {
         let volume = Arc::new(Mutex::new(1.0));
 
         // If paused is set to true then don't play from this handle.
-        let paused = Arc::new(Mutex::new(false));
+        let paused = Arc::new(AtomicBool::new(false));
 
         // If dead is set to true then this Handle should be removed.
-        let dead = Arc::new(Mutex::new(false));
+        let dead = Arc::new(AtomicBool::new(false));
 
         // `next_sounds` contains a Vec that can later be used to append new iterators to the sink
         let next_sounds = Arc::new(Mutex::new(Vec::new()));
@@ -181,9 +179,7 @@ impl Engine {
                 handle_id: handle_id,
                 queue_iterator: queue_iterator,
                 volume: volume.clone(),
-
                 paused: paused.clone(),
-
                 dead: dead.clone(),
             }
         );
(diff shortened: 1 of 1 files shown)


src/engine.rs
```diff
@@ -35,7 +35,15 @@ pub struct Engine {
 struct EndPointVoices {
     format: Format,
     next_id: AtomicUsize,
-    pending_sounds: Mutex<Vec<(usize, QueueIterator)>>,
+    pending_sounds: Mutex<Vec<HandleHandle>>,
+}
+
+// Handle to a handle I guess.
```
tomaka: Same two remarks as #84

Xaeroxe23: Ok, there is no such thing as an AtomicF32 though so I assume the volume level should still be a Mutex<f32>

Xaeroxe23:
The other two have been merged into this request and the requested changes have been made.

Xaeroxe23:
@tomaka Ready for review.

tomaka:
Please load the volume in a temporary variable outside of the iterator (for example between lines 126 and 127) so that we only lock the mutex once from time to time. That should be okay-ish I think.

src/lib.rs
```diff
@@ -103,6 +103,34 @@ impl Sink {
         self.handle.append(source);
     }
 
+    /// Gets the volume of the sound.
+    ///
+    /// The value `1.0` is the "normal" volume (unfiltered input). Any value other than 1.0 will
+    /// multiply each sample by this value.
+    #[inline]
+    pub fn get_volume(&self) -> f32 {
```
tomaka: Please rename this to just `volume()`.

src/engine.rs
```diff
@@ -206,18 +250,42 @@ impl Handle {
         self.next_sounds.lock().unwrap().push((source, Some(tx)));
     }
 
+    /// Gets the volume of the sound played by this sink.
+    #[inline]
+    pub fn get_volume(&self) -> f32 {
```
tomaka: Same here.

tomaka:
Sorry, I didn't review your volume PR carefully. This time it should be good after these changes.

Xaeroxe23:
Ok, so I agree it is not desirable for the sampling thread to block, locking can lead to blocking.  However I do want the volume to be fetched for every sample as that makes the set_volume function feel the most responsive.  It's not possible to make the f32 atomic.  So the approach I'm currently thinking of would use a [channel](https://doc.rust-lang.org/std/sync/mpsc/fn.channel.htmll)<f32>, we'd store the Sender in the Handle and the Receiver in the HandleHandle, then whenever the Handle wishes to change the volume level it will send an f32 through the channel, HandleHandle would store an f32 called volume and in the sample iterator (lines 128-133) we would call try_recv to fetch any new volume level sent by Handle, and store it in HandleHandle.  This will **not** be contained in a while loop, it will only attempt to receive one volume level per sample, which prevents the volume thread from being DDOSed by a large influx of volume change requests.

I'll create and push an implementation of this on my lunch break, let me know if you find any fault in it.

tomaka:
I don't think it's better to use a channel.

The iterator that builds the samples is a code that is very easy to optimize for the compiler. The fact that we're using the `Relaxed` ordering means that in theory it can fetch the value of "paused" only from time to time and build the samples through SIMD for example.

If you're introducing a channel read or a mutex in there, you're not only paying the price of the channel or mutex read, but you're also totally killing any potential for a compiler optimisation on this code.

In fact I have already doubts about the fact that `paused` doesn't do kill all optimisations already, but I thought it would be "good enough" while waiting for a more robust rewrite of the engine's code.

Xaeroxe23:
I have a difficult time imagining how the engine could have the features in this request in a more optimized manner, unless we were to make changes to cpal.  That being said, currently on line 126 of engine.rs I am dropping the sound if it was declared dead, and that seems to be responsive enough in my own testing, so on second thought I'm not opposed to loading temporary volume variables here.

tomaka:
> I have a difficult time imagining how the engine could have the features in this request in a more optimized manner, unless we were to make changes to cpal.

You could put the `Arc<Mutex<f32>>` of the volume in the source of samples itself, alongside with a local copy of the volume. The source of samples would then load the value from the mutex to its local copy from time to time.

Xaeroxe23:
Ok, that makes sense.  I'd like to see if I can accomplish that, should "from time to time" just be approximately every 5ms?

Xaeroxe23:
Some context: I'm using rodio in a game engine project and I want to give users of my engine more control over the playback of their sounds, those features that I need to do that are in this PR.  I don't have anywhere near the same level of experience manipulating audio devices as you do but I gave it my best attempt.  I'm trying to avoid having an audio engine that is just "good enough for now" I'd like it to be at this ideal performance level you describe.  So I'm willing to tackle the idea of finding a good place and way to do more infrequent fetching of the variables into a local copy.

Xaeroxe23:
Alternatively if you want to do a rewrite of the engine I could just add my temporary variable after line 126 as you originally suggested and just be done with it.

tomaka:
I'd like to rewrite some parts of the engine but I don't know if it's going to happen in the next month or in the next year.

Basically the ideal solution would be to add a filter (similar to the ones in `src/source`) whose constructor takes a source and an `Arc<Mutex<f32>>`.
The filter would calculate approximately the number of samples that are equal to 5 ms, and re-read from the mutex once that number of samples has been processed.
The actual duration of 5ms could perhaps be customized with another constructor.

Then the engine could wrap all sound sources passed to it around that filter.

Xaeroxe23:
Ok, do you also feel we should be doing something similar with the AtomicBools or is the current implementation sufficient for those?

tomaka:
@Xaeroxe23 Yeah I think that would be better.
Once you have written one of the filters, adding the others should be very easy after some copy-pasting.

Xaeroxe23:
It just occurred to me dead doesn't modify the sample output though so that's only really applicable to paused.

Xaeroxe23:
Resolved in another PR, closing.

PR #84, opened 2017-01-23
Sink drop

Xaeroxe23:
Fixes sinks to correctly stop playing sounds when they are dropped.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -35,7 +35,14 @@ pub struct Engine {
 struct EndPointVoices {
     format: Format,
     next_id: AtomicUsize,
-    pending_sounds: Mutex<Vec<(usize, QueueIterator)>>,
+    pending_sounds: Mutex<Vec<HandleHandle>>,
+}
+
+// Handle to a handle I guess.
+// The tuple type was getting a bit unwieldy so I added this.
+struct HandleHandle {
+    handle_id: usize,
+    queue_iterator: QueueIterator,
 }
 
 impl Engine {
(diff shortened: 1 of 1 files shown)


src/engine.rs
```diff
@@ -35,7 +35,15 @@ pub struct Engine {
...
-    pending_sounds: Mutex<Vec<(usize, QueueIterator)>>,
+    pending_sounds: Mutex<Vec<HandleHandle>>,
+}
+
+// Handle to a handle I guess.
+// The tuple type was getting a bit unwieldy so I added this.
+struct HandleHandle {
+    handle_id: usize,
+    queue_iterator: QueueIterator,
+    dead: Arc<Mutex<bool>>,
```
tomaka: `dead` should be a `Arc<AtomicBool>` instead of a `Mutex<bool>`.
You can use `Ordering::Relaxed` if you're not familiar with the ordering stuff.

src/engine.rs
```diff
@@ -35,7 +35,15 @@ pub struct Engine {
 struct EndPointVoices {
     format: Format,
     next_id: AtomicUsize,
-    pending_sounds: Mutex<Vec<(usize, QueueIterator)>>,
+    pending_sounds: Mutex<Vec<HandleHandle>>,
+}
+
+// Handle to a handle I guess.
+// The tuple type was getting a bit unwieldy so I added this.
```
tomaka: This comment should be removed.

tomaka:
Sorry for the misunderstanding, but since your three PRs have things in common it's ok to submit them together. It would avoid you having to solve merge conflicts.

Anyway, looks good except for the small problems.

src/engine.rs
Xaeroxe23: Ok, would

//Handle to a handle

work better?

tomaka: Yes

PR #85, opened 2017-01-24
Added implementation as source filters

Xaeroxe23:
Implementation of https://github.com/tomaka/rodio/pull/83 using source filtering with periodic updates.  The killable implementation needed to be accessible from the engine so it was not possible to make that one simply an audio filter.  I added that functionality to the QueueIterator because those are what the engine manages in its current form.

### DIFF
--- a/src/engine_filters/pauseable.rs
+++ b/src/engine_filters/pauseable.rs
@@ -1,0 +1,81 @@
+use std::sync::Arc;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::time::Duration;
+
+use Sample;
+use Source;
+
+/// Filter that allows another thread to pause the stream.
+#[derive(Clone, Debug)]
+pub struct Pauseable<I> where I: Source, I::Item: Sample {
+    input: I,
+
+    //Local storage of the paused value.  Allows us to only check the remote occasionally.
+    local_paused: bool,
+
+    //The paused value which may be manipulated by another thread.
+    remote_paused: Arc<AtomicBool>,
+
+    //The frequency with which local_paused should be updated by remote_paused
+    update_frequency: u32,
+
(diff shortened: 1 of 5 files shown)


src/engine_filters/pauseable.rs
```diff
@@ -0,0 +1,81 @@
...
+    //The frequency with which local_paused should be updated by remote_paused
+    update_frequency: u32,
+
+    //How many samples remain until it is time to update local_paused with remote_paused.
+    samples_until_update: u32,
+}
+
+impl<I> Pauseable<I> where I: Source, I::Item: Sample {
+    pub fn new(source: I, remote_paused: Arc<AtomicBool>, update_ms: u32) -> Pauseable<I> {
+        let update_frequency = (update_ms * source.get_samples_rate())/1000;
```
tomaka: In theory the samples rate of the source can change while it is playing.
However this never happens in practice and making this correct needs a lot of work, so please just add a comment here saying something like `// TODO: handle the fact that the samples rate can change`.

src/engine_filters/volume_filter.rs
```diff
@@ -0,0 +1,82 @@
...
+    //The frequency with which local_volume should be updated by remote_volume
+    update_frequency: u32,
+
+    //How many samples remain until it is time to update local_volume with remote_volume.
+    samples_until_update: u32,
+}
+
+impl<I> VolumeFilter<I> where I: Source, I::Item: Sample {
+    pub fn new(source: I, remote_volume: Arc<Mutex<f32>>, update_ms: u32) -> VolumeFilter<I> {
+        let update_frequency = (update_ms * source.get_samples_rate())/1000;
```
tomaka: Same here.

tomaka:
Looks good. R+ after the two comments are added.

Xaeroxe23:
Comments have been added

PR #88, opened 2017-01-28
QueueIterators for detached sinks are dropped when sources have stopped

### LINKED ISSUE #87: Memory leak
It looks like `Sink::detach` will leak the associated `QueueIterator` (even when no sounds are, or can be played anymore) until the stream for the associated endpoint is dropped.

This means if you create and detach too many sinks it degrades sound quality.

Eroc33:
Fixes #87
Edit: Not sure why but this seems to be preventing the first sound on an endpoint from playing, so it's probably not ready to be merged yet.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -104,7 +104,7 @@ impl Engine {
             let epv = end_point_voices.clone();
 
             let sounds = Arc::new(Mutex::new(Vec::new()));
-            future_to_exec = Some(stream.for_each(move |mut buffer| -> Result<_, ()> {
+            future_to_exec = Some(stream.for_each(move |mut buffer| -> Result<(), ()> {
                 let mut sounds = sounds.lock().unwrap();
 
                 {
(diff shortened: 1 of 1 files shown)


src/engine.rs
```diff
@@ -104,18 +104,27 @@ impl Engine {
...
                 {
                     let mut pending = epv.pending_sounds.lock().unwrap();
                     sounds.append(&mut pending);
                 }
 
+
                 if sounds.len() == 0 {
                     return Ok(());
                 }
                 sounds.retain(|s| !s.1.local_dead);
```
Xaeroxe23: I'd recommend merging these two calls to retain, we can take advantage of some short circuiting optimization if we do.

src/engine.rs
```diff
@@ -104,18 +104,27 @@ impl Engine {
...
+                //Drop if it's not playing a real source, and it's sink is detached
+                sounds.retain(|s|{
+                    if !s.1.is_playing_real_source{
+                        Arc::strong_count(&s.1.next)>1
```
Xaeroxe23: [strong_count](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.strong_count) is considered unstable.  We should look for some alternate implementation.

Xaeroxe23: Also, that logic needed to be inverted.  This was the cause of the issue you mentioned:

Edit: Not sure why but this seems to be preventing the first sound on an endpoint from playing, so it's probably not ready to be merged yet.

Xaeroxe23:
https://github.com/tomaka/rodio/pull/89 This is my alternative implementation.  I like this because it doesn't use strong_count which is unstable.

Xaeroxe23:
I'm also not sure why this version was experiencing the problem with not playing the first sound.  I tried debugging that myself but #89 doesn't seem to suffer from that issue thankfully.

Eroc33:
Closing in favour of #89

PR #95, opened 2017-02-07
Rewrite engine

tomaka:
This PR rewrites the engine of `rodio` to be more simple.

Before this PR, the engine would handle playing/pausing, stopping sounds, volume, queueing sounds and mixing.
Each of these capabilities has been extracted to a dedicated module.

The engine now only handles mixing and sending the mix output to the cpal voice.
For each endpoint the engine creates one voice (thanks to cpal) and one mixer. When you send a sound to the engine, the engine adds to the mixer. That's it. That's how it works.
The implementation of the mixer lives in its own module, so the code is nicely split and (I think) quite clean compared to before.

The `Sink` type that lives at the root now handles playing/pausing, stopping, queue and volume itself, and passes a single source to the engine.


tomaka:
Opened #96 for a potential future improvement.

PR #100, opened 2017-02-09
Implement decoding Flac

### LINKED ISSUE #37: Add a FLAC decoder

tomaka:
Close #37 

For some reason the sound is wrong and I don't really understand why.
The speed of the playback is correct, so the samples rate and channels must be good.
The channels and samples rate match what my hardware expect, so it shouldn't be a bug in the converter.

As far as I know the FLAC format uses only integer samples, so I don't think I'm misinterpreting the data.


tomaka:
Oh it's because the channels are not interleaved. Will fix that.

PR #103, opened 2017-02-17
Implement stop

icefoxen:
Also tiny doc fix to Stoppable.

Tested this and it appears to work, so this is becoming a pull request.  The reasoning is that without this you can't stop a playing sound without dropping the `Sink`, which becomes awkward when starting to play a sound again requires creating a new one.  Suddenly your structures keeping track of `Sink`s and `Source`s become mutable for some operations but not others.


src/sink.rs
```diff
@@ -79,7 +80,7 @@ impl Sink {
     ///
     /// No effect if not paused.
     #[inline]
-    pub fn play(&self) {
+    pub fn resume(&self) {
```
tomaka: Please add this as a separate method and put `#[deprecated]` on `play`. I don't want to break the API randomly like this.

src/sink.rs
```diff
@@ -99,6 +100,18 @@ impl Sink {
...
 
+    /// Stops playback.  To resume, a new sound must be added with `append()`
+    pub fn stop(&self) {
+        self.stopped.store(true, Ordering::Relaxed);
+    }
+
+    /// Gets whether a sound is stopped.
+    ///
+    /// A stopped sound cannot be resumed, a new one must be started with `append()`.
+    pub fn is_stopped(&self) -> bool {
```
tomaka: I don't think that's a good idea. If I add a 2-seconds sound to a sink then call `is_stopped` 3 seconds later, I'd expect it to return true while it will return false.

icefoxen: Reasonable enough.  Do you want me to change the behavior of `is_stopped()` or just remove it altogether?

tomaka: Just remove it I think

tomaka:
Superceded by #128

PR #109, opened 2017-05-01
Better API for amplify, and fix Sample::lerp

tomaka:


tomaka:
Closing because tests are failing and it looks more complicated than I thought.

Issue #40, opened 2015-10-01
Add sound spatialization


tomaka:

tomaka:
Some quick reference here: https://webaudio.github.io/web-audio-api/#Spatialization

Xaeroxe:
I'd like to take a crack at implementing this.  If I'm successful would you like a PR @tomaka ?

tomaka:
@Xaeroxe Sure.

Xaeroxe:
PR opened: https://github.com/tomaka/rodio/pull/125

Xaeroxe:
PR merged.  I believe we can close this now.

PR #125, opened 2017-07-01
Add spatial audio

Xaeroxe:
This addresses https://github.com/tomaka/rodio/issues/40

Things done in this PR
- Add ChannelVolume filter which allows a sound to be played to different channels at different volumes
- Add cgmath dependency to make working in 3 dimensional space easier
- Add Spatial sound filter, which is the back end for SpatialSink
- Add SpatialSink, which is a special sink type with support for playing sounds in 3 dimensional space
- Add spatial example for demonstration and testing purposes.

@tomaka indicated in #40 that he wanted our system to be similar to [this](https://webaudio.github.io/web-audio-api/#Spatialization).  I've made some modifications to that system, this implementation only has 1 sound fall-off algorithm and instead of azimuth with a single point in space this system uses multiple points in space to represent each "ear" and then applies the following formula:

```Rust
let left_diff_modifier = ((left_distance - right_distance) / max_diff + 1.0) / 4.0 + 0.5;
let right_diff_modifier = ((right_distance - left_distance) / max_diff + 1.0) / 4.0 + 0.5;
```

These modifiers are then multiplied into the final volume for each ear.

The purpose of this is to amplify the difference between what the two ears hear, which was the purpose of the azimuth calculation in the system used by web-audio.


tomaka:
Thanks for the PR!

- I don't want to publicly depend on `cgmath` (or any other crate). Using `cgmath` privately within rodio is fine (although I don't think it's necessary), but when it comes to parameters of public functions I don't want to expose any of the types from an external crate. Therefore please turn `Vector3<f32>` into `[f32; 3]`.

- While the `SpatialSink` is a nice addition, I think the primary implementation (ie. the formula) should be in a filter and the `SpatialSink` should simply be a convenient wrapper just like the `Sink` currently is.

Xaeroxe:
How would you feel about rodio having it's own Vector3 structure?  It's nice to have functions like `magnitude()` and a `Sub` implementation as it makes expressing Vector math so much easier.

If the primary implementation is in a filter then the calculations to determine the volume levels have to be done by the audio engine, which needs to be low latency.  The volume level calculations are computationally a bit more heavy than most of the other filters I think.  However I could certainly try moving it into a filter to see if that impacts audio quality at all.

kvark:
It appears that all you really receive from the user is positions, so even `cgmath::Vector3` is a little misleading. Accepting `[f32; 3]` seems to be less ambiguous here.
In the future, if you need more type diversity on the API side, and considering https://github.com/brendanzab/cgmath/pull/418 is accepted, I'd recommend exposing the `mint` types.

Xaeroxe:
Using vectors to represent points isn't really that uncommon in my experience, [Unity does it](https://docs.unity3d.com/ScriptReference/Transform-position.html), [UE4 does it](https://docs.unrealengine.com/latest/INT/API/Runtime/Core/Math/FTransform/index.html), [XNA and Monogame do it too](https://msdn.microsoft.com/en-us/library/microsoft.xna.framework.graphics.spritebatch.draw.aspx).

Xaeroxe:
If we just want to fix the public API @tomaka then I could make it accept a `f32; 3]` and then just convert to a `cgmath::Vector3` internally.  That seems to be one option you were getting at in your comment.

Xaeroxe:
Cgmath also has this type: https://docs.rs/cgmath/0.14.1/cgmath/struct.Point3.html which would probably be better suited to this honestly.  I'm so accustomed to used vector types to represent points I didn't even look for that.

kvark:
Frankly speaking, I don't think we should be inspired by the type systems of existing engines, since they are built on languages that are not as expressive. We can do better.

Xaeroxe:
It's not like they didn't have the option to make a Point type, operator overloading definitely isn't something Rust invented.  They just chose not to because ultimately the definition turns out to be completely redundant with that of the vector.  The only advantage to having a Point type is that it's more semantically clear.  They're the same structure with several similar operations.  I'm getting sidetracked though, and it's losing relevance since I made this PR use the Point type anyways.

Xaeroxe:
@tomaka I'm looking at my `SpatialSink` code and I don't see much of a reason to move it into a filter, because if I do then I have to transmit the positions across thread boundaries unnecessarily.  It also doesn't really behave much like a filter either, as it doesn't modify every sample, it just computes some volumes whenever the positions change then sends it to`ChannelVolume` which does the actual sample modification itself.  Having that be a filter would probably be detrimental to performance.  I consider this ready your review, let me know if you have any more thoughts on this or disagree with me on anything I've said.

tomaka:
Right now your PR does:

> user sets position  ->  spatial sink  ->  (thread boundary)  ->  channel volume

I think it could be:

> user sets position  ->   (thread boundary)  -> spatial filter  ->  channel volume

In order words the "spatial filter" would instantly call `set_volume`, without any sort of mutex or anything.
The mutex, thread boundary and period_access thingies would just happen outside of the spatial filter.

I don't see what could be detrimental to performances, unless I'm missing something. The only difference is that you'd put 36 bytes of data in the mutex instead of 16, but that doesn't really matter.

Xaeroxe:
Ok yeah that does make more sense.  Thanks for the explanation.  I'll make it happen.

Xaeroxe:
@tomaka does this look any better?

tomaka:
Yes, thanks!

Now for more minor changes:

- In the filter I think the set_emitter_position, set_left_ear_position, set_right_ear_position and update_volumes functions should be merged in just one. You shouldn't have to call `update_volumes` manually.
- Similarly, there's no reason to put the three properties in three different `Arc<Mutex<>>`. Just use a single Arc/Mutex and put either a tuple or a struct in it.
- `set_volume` should simply panic if out of range. If you really don't want to panic, an alternative is using a `Result`. A `bool` is wrong, though.

Xaeroxe:
@tomaka The ChannelVolume change you requested is now in place, there were two ways for me to interpret your other request so I've made commits with commit message Version 1 and Version 2.

Version 1: The `SpatialSink` interface doesn't change, however all positions are stored in a single structure under the same `Mutex` and the functions for `Spatial` are combined.

Version 2: In addition to the changes in Version 1, the `SpatialSink` and `Spatial` interface changes to use a new structure called `SoundPositions` which bundles all 3 coordinates into a single structure that implements `Debug`, `Clone`, and `Copy`.  This structure has its fields publicly exposed.

If you like version 1 better I'll revert Version 2 then squash my commits.
If you like version 2 better I'll just squash my commits.

tomaka:
Thanks!

Version 2 is not bad, but I think I prefer version 1.

Xaeroxe:
@tomaka As you wish, Version 2 has been reverted and commits have been squashed.

Xaeroxe:
I also determined that the only time the positions were used in the Spatial filter was when the volumes were being calculated, so it's actually not necessary to store them in the Spatial filter because the positions were already provided by the function arguments.

tomaka:
Oh it didn't stuck me that you stored them.

Anyway, thanks for the PR!

Xaeroxe:
Yay!  Thanks! By the way I just wanted to say @tomaka I'm super impressed with this code base.  It's such a pleasure to work in and the constructs you've provided are very powerful ways of managing audio effects.  Can't wait to use it in Amethyst.

PR #129, opened 2017-07-21
fix 'index out of bounds'

chyvonomys:
`SamplesBuffer` for `current_frame_length` returns `None`, so `Buffered` will never be `Frame::End` and this code will panic:
```
extern crate rodio;

fn main() {
    use rodio::Source;
    rodio::buffer::SamplesBuffer::new(1, 44100, vec![0.0]).buffered().count();
}
```


tomaka:
Thanks!

PR #133, opened 2017-08-20
Add empty() function to Sink

Xaeroxe:
This addresses https://github.com/tomaka/rodio/issues/104


src/sink.rs
```diff
@@ -40,13 +42,14 @@ impl Sink {
...
             stopped: Arc::new(AtomicBool::new(false)),
+            empty: Arc::new(AtomicBool::new(true)),
             detached: false,
         }
     }
 
     /// Appends a sound to the queue of sounds to play.
     #[inline]
-    pub fn append<S>(&self, source: S)
+    pub fn append<S>(&mut self, source: S)
```
tomaka: `append` really should keep `&self` instead of `&mut self`

src/sink.rs
```diff
@@ -70,7 +73,8 @@ impl Sink {
                 }
             })
             .convert_samples();
-
+        self.empty = Arc::new(AtomicBool::new(false));
```
tomaka: Why not do `self.empty.store(false, Ordering::Relaxed);` instead?

src/spatial_sink.rs
```diff
@@ -53,7 +53,7 @@ impl SpatialSink {
 
     /// Appends a sound to the queue of sounds to play.
     #[inline]
-    pub fn append<S>(&self, source: S)
+    pub fn append<S>(&mut self, source: S)
```
tomaka: Same here. Revert to `&self`.

Xaeroxe:
So I'm going to address all your review comments here:

The idea behind this PR is that when a source returns `None` from `next()` that it stores "true" in the arc bool it contains.  So every source append must generate a new `Arc<AtomicBool>` and update the sink to point to that new `Arc<AtomicBool>` as a result during the append operation we require mutability.  Simply storing `false` in the existing `Arc<AtomicBool>` will break the check for when the most recently added source ends.

tomaka:
Ah yeah that's true.
However another solution could be to use an `AtomicUsize`. When a sound is added, the counter is incremented. When a sound is finished, it is decremented. When the counter is equal to 0, the sink is empty.

Xaeroxe:
That's much better.  Thanks, I'll do that.

Xaeroxe:
@tomaka Ready for your review.

PR #138, opened 2017-10-07
make sure buffer doesn't contain garbage

### LINKED ISSUE #132: Sound repeats forever when detaching sink (on Windows)
On Linux, if I detach the sink, it keeps playing the sound clip until it finishes. On Windows, a short segment (<100 ms) of the sound clip I'm playing starts repeating forever when I detach the sink. I've even tried stopping the sink before detaching it, but it has no effect. The only workaround I've found is to sleep for some time (e.g. 100 ms) after stopping it, before detaching it, but that's a bit inelegant.

```
extern crate rodio;

use std::fs::File;
use std::io::BufReader;
use std::time::Duration;

fn main() {
	let endpoint = rodio::get_default_endpoint().unwrap();
	let mut sink = rodio::Sink::new(&endpoint);
	sink.set_volume(0.1);

        // Using beep.wav from rodio/examples
	let source = File::open("res/beep.wav").ok()
		.and_then(|file| rodio::Decoder::new(BufReader::new(file)).ok()).unwrap();

	sink.append(source);
	std::thread::sleep(Duration::from_millis(100));

	// A stop, followed by a short sleep, prevents the bug from occuring.
	// sink.stop();
	// std::thread::sleep(Duration::from_millis(100));

	sink.detach();
	std::thread::sleep(Duration::from_secs(10)); // Sleep so we can hear the annoying sound for a while.
}
```

chyvonomys:
This fixes #132.

Given `buffer` is obtained from `IAudioRenderClient::GetBuffer` and may contain garbage (previous sound for example). `IAudioRenderClient::ReleaseBuffer` currently always returns whole buffer (in `cpal`) no matter if we filled it completely or partially or none, so we have to be sure it is filled with correct data.

My test code:
```rust
extern crate rodio;

use std::time::Duration;
use rodio::Source;

fn main() {
    let endpoint = rodio::get_default_endpoint().unwrap();
    let mut sink = rodio::Sink::new(&endpoint);
    sink.set_volume(0.1);

    let source = rodio::source::SineWave::new(440).take_duration(Duration::from_secs(10));

    println!("start 10 sec of sound");
    sink.append(source);

    std::thread::sleep(Duration::from_secs(1));

    println!("detach after 1 sec");
    sink.detach();

    std::thread::sleep(Duration::from_secs(9));

    println!("silence");

    std::thread::sleep(Duration::from_secs(100));
}
```

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -141,18 +141,18 @@ fn new_voice(endpoint: &Endpoint, events_loop: &Arc<EventLoop>)
     let future_to_exec = stream.for_each(move |mut buffer| -> Result<_, ()> {
         match buffer {
             UnknownTypeBuffer::U16(ref mut buffer) => {
-                for (o, i) in buffer.iter_mut().zip(mixer_rx.by_ref()) {
-                    *o = i.to_u16();
+                for d in buffer.iter_mut() {
+                    *d = mixer_rx.next().map(|s| s.to_u16()).unwrap_or(0u16);
                 }
             },
             UnknownTypeBuffer::I16(ref mut buffer) => {
-                for (o, i) in buffer.iter_mut().zip(mixer_rx.by_ref()) {
-                    *o = i.to_i16();
+                for d in buffer.iter_mut() {
+                    *d = mixer_rx.next().map(|s| s.to_i16()).unwrap_or(0i16);
                 }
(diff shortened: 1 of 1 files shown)


tomaka:
Thanks!

PR #140, opened 2017-10-11
Replace usages of deprecated functions in cpal

camjackson:
I get deprecation warnings from cpal when using rodio. This adds `pub use` of the non-deprecated functions, and also replaces some internal/example usages of deprecated methods.

I've run the tests and the examples and everything still works, although the FLAC playback is really choppy on my machine for some reason.

### DIFF
--- a/examples/basic.rs
+++ b/examples/basic.rs
@@ -5,7 +5,7 @@ use std::thread;
 use std::time::Duration;
 
 fn main() {
-    let endpoint = rodio::get_default_endpoint().unwrap();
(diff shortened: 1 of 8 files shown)


tomaka:
Thanks! These functions were deprecated this morning.

camjackson:
Haha wow, just lucky timing on my part I guess :)

While I'm here, thanks for all the amazing work you do on Rust OSS @tomaka! And I really enjoyed your talk at RustFest in Zurich the other week! 😄

PR #146, opened 2017-10-29
Optional decoders

chyvonomys:
This PR adds ability to turn off unwanted decoders and their dependencies. (`wav/hound`, `vorbis/lewton`, `flac/claxon`). Can be useful if you use `rodio` with one format only and don't need any extra. It is also valid to turn off all decoding capabilities. By default all three are turned on.
(Similar thing is possible in `image` crate, I used it there and thought it may be useful for `rodio` as well)

### DIFF
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -17,8 +17,13 @@ mod wav;
 /// Source of audio samples from decoding a file.
 ///
 /// Supports WAV, Vorbis and Flac.
+#[cfg(any(feature = "wav", feature = "flac", feature = "vorbis"))]
 pub struct Decoder<R>(DecoderImpl<R>) where R: Read + Seek;
 
+#[cfg(not(any(feature = "wav", feature = "flac", feature = "vorbis")))]
+pub struct Decoder<R>(::std::marker::PhantomData<R>);
+
+#[cfg(any(feature = "wav", feature = "flac", feature = "vorbis"))]
 enum DecoderImpl<R>
     where R: Read + Seek
 {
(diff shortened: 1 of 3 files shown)


src/decoder/mod.rs
```diff
@@ -30,29 +36,36 @@ impl<R> Decoder<R>
     /// Builds a new decoder.
     ///
     /// Attempts to automatically detect the format of the source of data.
-    pub fn new(data: R) -> Result<Decoder<R>, DecoderError> {
-        let data = match wav::WavDecoder::new(data) {
-            Err(data) => data,
+    pub fn new(_data: R) -> Result<Decoder<R>, DecoderError> {
+        #[cfg(feature = "wav")]
```
tomaka: I'd prefer if the variable names didn't have a _ prefix, and you just put a `#![allow(unused_variables)]` in the function instead.

src/lib.rs
```diff
@@ -108,13 +113,15 @@ mod sink;
...
 pub mod buffer;
+#[cfg(any(feature = "wav", feature = "flac", feature = "vorbis"))]
 pub mod decoder;
 pub mod dynamic_mixer;
 pub mod queue;
 pub mod source;
 
 /// Plays a sound once. Returns a `Sink` that can be used to control the sound.
 #[inline]
+#[cfg(any(feature = "wav", feature = "flac", feature = "vorbis"))]
```
tomaka: It's not great to remove this function if you don't enable any decoder. I think the `DecoderError` should just be exposed in any situation instead.

chyvonomys:
Yea, I agree. I made necessary changes to have `play_once` and `Decoder` in place.
Tested on `play_wav` example. It panics with `UnrecognisedFormat` as expected now.

tomaka:
Is the split between two different versions of `Decoder` necessary?
I feel like it should just work with the empty enum.

chyvonomys:
Yes, otherwise compiler complains about unused `R`.
Other option would be to have one more variant for `DecoderImpl`, but I don't feel like that is better.

tomaka:
I see, thanks!

PR #149, opened 2017-12-24
Fix empty / wrong audio output on ALSA

### LINKED ISSUE #135: No sound output when using default Endpoint (ALSA)
So this is a bit strange: when running the cpal's ["beep" example](https://github.com/tomaka/cpal/blob/master/examples/beep.rs) (which uses `get_default_endpoint`) everything works as expected but running rodio's ["basic" example](https://github.com/tomaka/rodio/blob/master/examples/basic.rs) (which also uses the re-exported `get_default_endpoint`) no sound is playing.

Using the first endpoint from `get_endpoints_list` works as expected (in both examples).

I tried [changing the format selection](https://github.com/tomaka/rodio/blob/4ca81f7c5f51bfdda60029c87c8430361653be6b/src/engine.rs#L106) to use the first format like in the cpal example but that didn't help.

Apparently something's not right with the event loop because after I inserted a `println` [here](https://github.com/tomaka/rodio/blob/4ca81f7c5f51bfdda60029c87c8430361653be6b/src/engine.rs#L142) it's only printed exactly once.

edit: (In both examples the default endpoint is named "default" and the first endpoint from `get_endpoints_list` is "pulse".)

fschutt:
I am not sure if this fix can be merged without careful review.

- The update from `cpal 0.5.1` to `cpal 0.6.0` fixes #135 - you at least get **some** sound on Linux now. I don't know the reason for this or what happened between `0.5.1` and `0.6.0`.
- The code change now does this: On the initialization of a new voice, it sets the formats `samples_rate` to the songs samples_rate. If you don't do this, you get stuttering (i.e. between every sample, there is an audible pause, the song is slowed down * 2).

Now I don't quite know as much about audio processing as I'd like to know, but I think what's happening is that the `.with_max_samples_rate()` function sets a samples rate that is incompatible with the actual song being played (resulting in audible tearing). It might be that this should be fixed at a different point in the code, but I'm not completely sure, I haven't digged through the architecture of `rodio` too much.

I originally notice this here: https://github.com/fschutt/LudumDare40/blob/master/src/audio.rs#L74 - the song would only play correctly if I set the samples rate again.

Now an obvious case would be that the song has a weird samples rate that isn't supported by the system, however I am not sure if it is `rodio`-s job to fix this. Of course you could also go ahead and first filter the supported audio formats by the sources sample_rate. But I am not sure if I should do this.

With this PR, the sound on Linux (at least on ALSA) works perfectly for me, but I am not sure if this bug is present on other operating systems.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ documentation = "http://docs.rs/rodio"
 
 [dependencies]
 claxon = { version = "0.3.0", optional = true }
-cpal = "0.5.1"
+cpal = "0.6.0"
 hound = { version = "1.0.0", optional = true }
 lazy_static = "0.1.12"
 lewton = { version = "0.5", optional = true }
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -101,7 +101,7 @@ fn start<S>(engine: &Arc<Engine>, endpoint: &Endpoint, source: S)
 
         match end_points.entry(endpoint.name()) {
             Entry::Vacant(e) => {
-                let (mixer, voice) = new_voice(engine, endpoint);
+                let (mixer, voice) = new_voice(engine, endpoint, &source);
                 e.insert(Arc::downgrade(&mixer));
                 voice_to_start = Some(voice);
                 mixer
@@ -110,7 +110,7 @@ fn start<S>(engine: &Arc<Engine>, endpoint: &Endpoint, source: S)
                 if let Some(m) = e.get().upgrade() {
                     m.clone()
                 } else {
-                    let (mixer, voice) = new_voice(engine, endpoint);
+                    let (mixer, voice) = new_voice(engine, endpoint, &source);
                     e.insert(Arc::downgrade(&mixer));
                     voice_to_start = Some(voice);
                     mixer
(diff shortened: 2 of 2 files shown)


tomaka:
Thanks for the PR.

Unfortunately your fix is highly-specific to ALSA, so I can't merge this.

> however I am not sure if it is rodio-s job to fix this

It *is* rodio's job to fix this.

fschutt:
> Unfortunately your fix is highly-specific to ALSA, so I can't merge this.

So .. does it break things on Windows or Mac? Yes, this fix is probably specific to ALSA, but where should I put it? Once the voice has been built, there's no way to adjust the `samples_rate` as far as I can see. And leaving it in the state that it is now isn't helpful, because right now the sound simply doesn't work at all, so there's that. I would at least do an upgrade to `cpal 0.6.0`.

I could also see a refactoring, so that the voice gets built in two steps, so that the ALSA-specific thing can be adjusted only on ALSA. But on the other hand, I am not sure why this fix is specific to ALSA - actually, there shouldn't be sound on any OS. Or it's a bug in cpal.

fschutt:
Alright, I didn't see your explanation, I'm closing this. I'll try to debug the conversion module then.

PR #162, opened 2018-04-19
Update cpal dependency to 0.8

### LINKED ISSUE #160: "basic" example crashes with PulseAudio on LInux
Running the "basic" example produces a crash on my machine with the following output.
```
$ cargo run --example basic
    Finished dev [unoptimized + debuginfo] target(s) in 0.23 secs
     Running `target/debug/examples/basic`
thread 'main' panicked at 'sample rate could not be set: "Invalid argument"', libcore/result.rs:945:5
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

I think this is related to an [issue](https://github.com/tomaka/cpal/issues/173) in `cpal`, where the maximum sample rate reported by ALSA is "garbage" (`4294967295`) when using PulseAudio.

The issue has has been solved in `cpal >= 0.8` by using a better default, so updating the dependency version might be sufficient.

frazar:
Fixes #160.

This pull request includes:
- a bit of refactoring, in order to follow some changes in the naming conventions used in `cpal` 0.8. I also tried to adopt the rodio internals the new naming convention as closely as possible. Let me know if this is desirable or not.
- `cpal` 0.8 introduces a hard distinction between input and output devices, where only the latter is now used in `rodio`.
- [removal](https://github.com/frazar/rodio/commit/255db6aa7195c55df60c487fd8bedebede7433b3) of a `rodio` piece of code choosing the best `Format` for a given `Device` in favour of the `default_output_format()` method provided by `cpal`

There are a number of __breaking changes__ in the user-facing API of `rodio`, in particular
- renamed `Endpoint` to `Device`
- splitted `default_endpoint()` into `default_output_device()` and `default_input_device()`
- renamed `endpoints()` to `devices()`
- introduced `output_devices()` and `input_devices()`

I tried to test the changes by compiling the crates that depdend on `rodio >= 0.6`, [as listed on crates.io](https://crates.io/crates/rodio/reverse_dependencies). In order for these crates to run, I had to adapt to the breaking changes introduced by this pull requests.
They are, 
- [`ggez`](https://crates.io/crates/ggez): compiles fine, tests pass, and examples finally run without being affected by #160 (as opposed to before).
- [`impose`](https://crates.io/crates/impose): compiles, but has no tests
- [`audact`](https://crates.io/crates/audact): compiles, tests pass
- [`macromusic`](https://crates.io/crates/macromusic): compiles, tests dont pass neither before or after my changes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ documentation = "http://docs.rs/rodio"
 
 [dependencies]
 claxon = { version = "0.3.0", optional = true }
-cpal = "0.7.0"
+cpal = "0.8"
 hound = { version = "3.3.1", optional = true }
 lazy_static = "1.0.0"
 lewton = { version = "0.5", optional = true }
--- a/examples/basic.rs
+++ b/examples/basic.rs
@@ -5,24 +5,24 @@ use std::thread;
 use std::time::Duration;
 
 fn main() {
-    let endpoint = rodio::default_output_device().unwrap();
+    let device = rodio::default_output_device().unwrap();
 
     let file = std::fs::File::open("examples/beep.wav").unwrap();
-    let mut beep1 = rodio::play_once(&endpoint, BufReader::new(file)).unwrap();
+    let mut beep1 = rodio::play_once(&device, BufReader::new(file)).unwrap();
     beep1.set_volume(0.2);
     println!("Started beep1");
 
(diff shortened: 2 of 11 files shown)


tomaka:
Thanks!

frazar:
Can we have a 0.7 release of `rodio` now? :smile:

PR #170, opened 2018-06-22
Prepare 0.8 release

Xaeroxe:
This PR adds a few mentions about MP3 decoding and releases 0.8


tomaka:
Thanks! The release should be automatic.

PR #172, opened 2018-07-26
Update lewton to 0.8.0

est31:
See the [changelog](https://github.com/RustAudio/lewton/blob/bbd45f0e7b5a8be95d7b3ee4374d7833444dfef0/CHANGELOG.md) for changes in lewton.

The update increases the minimum required Rust version of lewton to 1.20.0 (from something like 1.9.0 or something even older than that, not entirely sure). I say this update is okay as a) after deleting `Cargo.lock`, I wasn't able to compile this crate with Rust 1.20, so presumably one of rodio's (transitive?) dependencies already requires a newer Rust version and b) 1.20 was released in August 2017 so is almost one year old by now.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ claxon = { version = "0.3.0", optional = true }
 cpal = "0.8"
 hound = { version = "3.3.1", optional = true }
 lazy_static = "1.0.0"
-lewton = { version = "0.5", optional = true }
+lewton = { version = "0.8", optional = true }
 cgmath = "0.14"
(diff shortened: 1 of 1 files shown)


tomaka:
Thanks!

Is there a reason why your changelog is "upside down"? I imagine most people reading it are interested in the latest changes, not the changes from months ago.

est31:
The reason is probably that I don't have that many changelogs :). I have flipped the order: https://github.com/RustAudio/lewton/commit/21dd2b4a7b6f6d09b96ce5b39e7496a350c7028c

PR #174, opened 2018-08-16
Update lewton to 0.9

est31:
See the [changelog](https://github.com/RustAudio/lewton/blob/fa554eabc8ff723779362caed01f723d4efd6097/CHANGELOG.md#release-090---august-16-2018) for changes in lewton.

No change in the MSRV of lewton. It continues to require 1.20.0.


tomaka:
Thanks!

Issue #64, opened 2016-01-08
Relicense under dual MIT/Apache-2.0


cmr:
# Why?

The MIT license requires reproducing countless copies of the same copyright
header with different names in the copyright field, for every MIT library in
use. The Apache license does not have this drawback, and has protections from
patent trolls and an explicit contribution licensing clause. However, the
Apache license is incompatible with GPLv2. This is why Rust is dual-licensed as
MIT/Apache (the "primary" license being Apache, MIT only for GPLv2 compat), and
doing so would be wise for this project. This also makes this crate suitable
for inclusion in the Rust standard distribution and other project using dual
MIT/Apache.
# How?

To do this, get explicit approval from each contributor of copyrightable work
(as not all contributions qualify for copyright) and then add the following to
your README:

```
## License

Licensed under either of
 * Apache License, Version 2.0 ([LICENSE-APACHE](LICENSE-APACHE) or http://www.apache.org/licenses/LICENSE-2.0)
 * MIT license ([LICENSE-MIT](LICENSE-MIT) or http://opensource.org/licenses/MIT)
at your option.

### Contribution

Unless you explicitly state otherwise, any contribution intentionally submitted
for inclusion in the work by you shall be dual licensed as above, without any
additional terms or conditions.
```

and in your license headers, use the following boilerplate (based on that used in Rust):

```
// Copyright (c) 2015 t developers
// Licensed under the Apache License, Version 2.0
// <LICENSE-APACHE or
// http://www.apache.org/licenses/LICENSE-2.0> or the MIT
// license <LICENSE-MIT or http://opensource.org/licenses/MIT>,
// at your option. All files in the project carrying such
// notice may not be copied, modified, or distributed except
// according to those terms.
```

And don't forget to update the `license` metadata in your `Cargo.toml`!
# Contributor checkoff
- [x] @tomaka

Xaeroxe23:
- [x] @Xaeroxe23

Xaeroxe23:
Contributor checkoff update:

- [ ] @icefoxen 
- [ ] @est31 
- [ ] @petrochenkov 
- [ ] @psych0d0g

est31:
I license all my past, present, and future contributions to this crate, under the dual `MIT/Apache-2.0` license as above, without any additional terms or conditions.

petrochenkov:
I agree with relicensing my contributions to https://github.com/tomaka/rodio to the dual MIT/Apache-2.0 license.

[...]

PR #175, opened 2018-08-16
Require all new contributions to be MIT/Apache2

est31:
Since the last update in issue #64 , new contributors have added their contributions. As we don't know how many additional contributors might send PRs to this repo, I suggest we start requiring Apache2/MIT from any new contributors. The gfx project as well as the Rust RFCs repo have been following a similar approach.


tomaka:
Thanks for taking care of this!

PR #182, opened 2018-08-29
repeat: implement Clone

mathstuf:
It seems the compiler is not deriving `Clone` automatically even though
it is just `Buffered` wrappers around the original data.


tomaka:
Thanks

PR #188, opened 2018-09-21
added `.stop()` to SpacialSink

AntonHermann:
As pointed out in this issue [https://github.com/tomaka/rodio/issues/184](184), `SpatialSink` is missing the `stop()` method `Sink` has. Ad-Hoc, I couldn't think of a reason to let this out on purpose, so here is my proposed fix :)

PS: Are there any contributing guidelines for new contributors?
I'm looking for an open source project to contribute to and I really like `rodio`s design and would be very happy if i could help

### DIFF
--- a/src/spatial_sink.rs
+++ b/src/spatial_sink.rs
@@ -113,6 +113,12 @@ impl SpatialSink {
         self.sink.is_paused()
     }
 
+    /// Stops the sink by emptying the queue.
+    #[inline]
+    pub fn stop(&self) {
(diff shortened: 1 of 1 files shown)


tomaka:
Thanks for the PR! Looks good.

> PS: Are there any contributing guidelines for new contributors?
I'm looking for an open source project to contribute to and I really like rodios design and would be very happy if i could help

There is no contributor guidelines, but I'd be glad to get help!

PR #199, opened 2018-11-03
Fix reading WAV of different encodings.

### LINKED ISSUE #195: 24 bit integer and 32 bit float WAV files load but do not play sound.
It seems that 24 bit and 32 bit float WAV files load but do not play sound in Rodio. In fact, inspecting the actual samples when decoding one of these wavs reveals that all the samples are set to `0i16`.

I tested this via WAVs exported from LMMS and Audacity. Only the 16 bit integer ones would play, the rest would load but have no sound when played, although they do seem the run for the correct amount of time. Here is my `main.rs`, which was copy pasted from the `music_wav.rs` example that this repo has.

```rust
extern crate rodio;

use std::io::BufReader;

fn main() {
    let device = rodio::default_output_device().unwrap();
    let sink = rodio::Sink::new(&device);

    println!("Playing audacity 16bit wav");
    let file = std::fs::File::open("examples/audacity16bit.wav").unwrap();
    sink.append(rodio::Decoder::new(BufReader::new(file)).unwrap());
    sink.sleep_until_end();

    println!("Playing audacity 32bit float wav");
    let file = std::fs::File::open("examples/audacity32bit.wav").unwrap();
    sink.append(rodio::Decoder::new(BufReader::new(file)).unwrap());
    sink.sleep_until_end();

    println!("Playing lmms 16bit wav");
    let file = std::fs::File::open("examples/lmms16bit.wav").unwrap();
    sink.append(rodio::Decoder::new(BufReader::new(file)).unwrap());
    sink.sleep_until_end();

    println!("Playing lmms 24bit wav");
    let file = std::fs::File::open("examples/lmms24bit.wav").unwrap();
    sink.append(rodio::Decoder::new(BufReader::new(file)).unwrap());
    sink.sleep_until_end();

    println!("Playing lmms 32bit float wav");
    let file = std::fs::File::open("examples/lmms32bit.wav").unwrap();
    sink.append(rodio::Decoder::new(BufReader::new(file)).unwrap());
    sink.sleep_until_end();
}
```

This reads from a folder named examples, which I have attached below.

[examples.zip](https://github.com/tomaka/rodio/files/2522188/examples.zip)

I am on Windows 10 Home, 64 bit

a2aaron:
Fixes #195. This PR makes it so that when getting samples from a wav we actually check the encoding of the wav. This is necessary because hound returns `Err` whenever we attempt to read with the wrong encoding type. Note that the different encodings aren't exposed beyond `wav.rs`, as we simply just convert `f32` and `i32` values in `i16`s, which causes some precision loss but wasn't audible when I tested it on files exported from LMMS and Audacity.

This PR also adds a test in the `tests/` folder that just ensures that a wav that isn't silent doesn't get read as silence.


teknico:
Loss of precision not being audible might not be a useful criterion for some applications. The library seems to support `f32` throughout, so it would seem possible to leave `f32` samples as they are, and convert 24 bit ones to `f32`.

tomaka:
Looks good, thanks.

Issue #212, opened 2018-12-01
MP3 playback clipping.


GlenCFL:
MP3 playback using the [mp3 music example](https://github.com/tomaka/rodio/blob/master/examples/music_mp3.rs) seems to be clipping something like a second of audio from the end of the audio stream. This is particularly apparent on short audio clips, with a 1 second duration MP3 not outputting any sound and a 2 second duration MP3 cutting out midway through playback.

This is on Windows 10 x64 using rodio 0.8.1.

germangb:
Could you file an issue in the `minimp3-rs` crate? The following line might return an early EOF

https://github.com/germangb/minimp3-rs/blob/master/src/lib.rs#L78

germangb:
Seems the `minimp3` crate is returning an early EOF. Could you file a bug in https://github.com/germangb/minimp3-rs/issues ?

Thanks!

PR #213, opened 2018-12-07
Bump minimp3 version

germangb:
- Update `slice-deque` dependency just in case (gnzlbg/slice_deque#57)
- Fix mp3 playback clipping issue #212
- Update minimp3 C library


tomaka:
This is normally not needed. Anyone can run `cargo update` to import bugfixes from `minimp3`.
And if someone doesn't run `cargo update`, they will also not get the rodio update containing this version bump.
But it also doesn't hurt to merge.

PR #217, opened 2019-01-28
Add fadeout

mathstuf:


src/source/fadeout.rs
```diff
@@ -0,0 +1,95 @@
...
+where
+    I: Source,
+    I::Item: Sample,
+{
+    type Item = I::Item;
+
+    #[inline]
+    fn next(&mut self) -> Option<I::Item> {
+        if self.remaining_ns <= 0.0 {
+            return self.input.next();
```
tomaka: So once we reach 0.0, we will play full volume again?

mathstuf: Ooh, good point.

mathstuf:
Updated. Now pins volume to `0.` when the fadeout time has expired.

OvermindDL1:
A fadeout seems exceptionally situation specific rather then generic.

Why not instead have something that just does standard audio envelope manipulation?  I.E. it would be defined with an `attack`, a `decay`, a `sustain`, and a `release` in that order.  When applied to a known-length source the attack adjust the beginning of it, trivial to do a fade-in (as well as more complex things), the decay then is the transition from the end of the attack to the sustain, the sustain is then the 'settled on' volume or other adjustments (this should work on a lot more than just volume depending on the callback passed in of course), then the release can either be done right before it ends, or potentially right after (to extend it, which is more traditional but both are often supported).

Fade-out and fade-in would both be trivially and efficiently built on that without requiring a specialized multitude of structs for each and every potential sound manipulation that could be done via normal audio envelope's.  For example, a fade-out would just be a release (or some step before it, prerelease perhaps? would only work on sources with known lengths though all the others would work on sources with unknown lengths) with the rest set to 0 times and no-ops as is traditional.  The code would execute just as efficiently and it would give significant power.  You can then define `fadein`/`fadeout` functions on sinks as wrappers to the envelope (as well as a full `envelope` function, as well as other common envelope effects).

mathstuf:
That seems beyond my knowledge of audio manipulation (in that I know what those words mean in general, but not in this context). As long as a method for fading out that takes a `Duration` exists (possibly with some `Ramp::Linear` argument for a little bit more fanciness, but even that seems a bit much for something that many other audio libraries also just have a simple duration parameter for), I'm OK with it. To that end, when such an API does exist, the implementation of this method can then call that API. But, I don't think it's something I would feel confident implementing at this time.

If you really want to go far, something built on [`splines`](https://docs.rs/splines/3.0.0/splines/index.html) seems possible as well and isn't restricted to a trapezoid shape (just going on my intuition here).

OvermindDL1:
@mathstuf Actually it's all really simple things, much more so then most people think.  I'd recommend going through this audio tutorial when you get a half hour or hour or so, it's all in-browser but teaches what they are and how they work.  The envelope section is directly related here but adding all the traditional audio effects would be very useful:  https://learningsynths.ableton.com/

mathstuf:
Finally got some time to go back and look at this :) . So looking at that, I think that's slightly different. Are you suggesting to have a source which applies some envelope to another source? That seems a bit different here where we don't know how long `sustain` is since the `fadeout` portion is triggered by an external event. I think another `envelope` source modifier makes sense (maybe this can then be rebased on top of that, but without linear types enforcing the attack to decay to sustain to release flow, the API feels like its going to be unweildy.

I also have no use case for such features, so I don't want to sink more time into it.

dyc3:
I can clearly see the advantage of having an envelope based source, but I think it should be a separate source. The fade in/out sources should stay because they are easy to use (if anything, refactored to work on top of the envelope source).

Any chance we can get this merged in? I have a project that needs this feature badly.

paulirotta:
Nice PR. Is there a pattern for using it to fade down the end of an audio stream? Since the fade runs over the entire remaining time, do you somehow trigger it X ms before the end, or..?

Thanks in advance. If someone wants to try it with the cpal 0.12 pre-release version of rodio see https://github.com/paulirotta/rodio/tree/fadeout

mathstuf:
My main use case is to fade out a looping piece of music at the end of a level. I suppose you could compute the length of the audio doing something like:

```rust
if let Some(duration) = source.total_duration() {
    let source = source.fadeout(source, fade_out_duration);
    source.periodic_access(duration - fade_out_duration, |s| s.start())
} else {
    // Probably Box<Source> to align with the above.
}
```

sighclone:
I do agree with @dyc3 that having a separate fade_out is beneficial for simplicity's sake and ease of use.

For now at least, for those who are looking, I'll just put this here...
#248 Allows us to perform fade out like so:

````rust
let quiet = SineWave::new(0.0);
let my_sound_source = SineWave::new(440.0)    // 440Hz
    .take_duration(Duration::from_secs_f32(3.0))
    .amplify(1.0)
    .fade_in(Duration::from_secs_f32(1.0))
    .take_crossfade_with(quiet, Duration::from_secs_f32(2.0));    // fade_in and crossfade are provided in rodio
    /*
        fade in starts at 1.0 seconds
        fade out starts after 2.0 seconds, i.e, for the last 1.0 seconds of the total 3.0 seconds
    */
sink.append(my_sound_source);
````

dvdsk:
This is implemented. See the method: `Source::fade_out`

I do not know when it happened and why this PR was not used/merged.

sighclone:
> A fadeout seems exceptionally situation specific rather then generic.
> 
> Why not instead have something that just does standard audio envelope manipulation? I.E. it would be defined with an `attack`, a `decay`, a `sustain`, and a `release` in that order. When applied to a known-length source the attack adjust the beginning of it, trivial to do a fade-in (as well as more complex things), the decay then is the transition from the end of the attack to the sustain, the sustain is then the 'settled on' volume or other adjustments (this should work on a lot more than just volume depending on the callback passed in of course), then the release can either be done right before it ends, or potentially right after (to extend it, which is more traditional but both are often supported).

@OvermindDL1 
Hi, any idea if `attack`, `decay`, `sustain` and `release` options are now available?

OvermindDL1:
I don't know, I don't currently use it as I ended up using pulse audio and then pipe wire directly as I really wanted some of their features that this library did not expose...

dvdsk:
> Hi, any idea if `attack`, `decay`, `sustain` and `release` options are now available?

we have: https://docs.rs/rodio/latest/rodio/source/trait.Source.html#method.linear_gain_ramp with:
 
- duration: [Duration](https://doc.rust-lang.org/nightly/core/time/struct.Duration.html),
- start_value: [f32](https://doc.rust-lang.org/nightly/std/primitive.f32.html),
- end_value: [f32](https://doc.rust-lang.org/nightly/std/primitive.f32.html),
- clamp_end: [bool](https://doc.rust-lang.org/nightly/std/primitive.bool.html),

sighclone:
Hi @dvdsk,
It does seem possible to implement a basic ADSR envelope, perhaps easily, built _on top of_ the linear gain ramp method (at least for a proof of concept), so I should also be able to build it by modifying the method too. I'm willing to take a look over the weekend. When is the next release scheduled for?

dvdsk:
> I'm willing to take a look over the weekend.

Please do, more sources are almost always welcome. Take a look at: https://github.com/RustAudio/rodio/blob/master/CONTRIBUTING.md before you start, it contains some useful advise.

> When is the next release scheduled for?

There is none scheduled, the plan is to apply all the breaking changes we have planned for a while (see #614). Then write a porting/upgrade guide and release that. But you can of course rely on rodio as a git dependency to get new features early.

Issue #171, opened 2018-07-13
Cannot restart stopped sink


mbillingr:
Hi there! Thank you for this awesome library. While exploring the API I stumbled over the following behavior, which I believe is a bug:

A `Sink` no longer plays newly appended sounds after `.stop()` has been called on the `Sink`.

Looking at the sources, it seems that `.stop()` sets `Controls.stopped` to `true`, but I could not find anything that would set it back to `false` other than constructing a new `Sink`.

tomaka:
Hi! By design we cannot restart a stopped source. The `next()` method of the `Iterator` produces a `None`, which indicates that the source is over.

mbillingr:
Thank you for the response! 
I am aware of the `Iterator` based design behind sources. However, I have meant to refer to sinks instead. Chances are that I made a mistake, though. Here is what I did in code:

```rust
    let device = rodio::default_output_device().unwrap();
    let sink = rodio::Sink::new(&device);

    let source_a = rodio::source::SineWave::new(440);
    let source_b = rodio::source::SineWave::new(880);

    sink.append(source_a);  // played correctly
    thread::sleep_ms(1000);
    sink.stop();
    thread::sleep_ms(500);
    sink.append(source_b);  // only short "click" sound
    thread::sleep_ms(1000);
```

If I do not use `sink.stop()` but limit the length of `source_a` with `.take_duration()` `source_b` is played correctly.

tomaka:
Once you have stopped a sound, it's not possible to restart it even with `append`.
I guess this should either be properly documented, or `append` should return a `Result`.
Also you're not supposed to hear a click.

tomaka:
> To clarify, if I stop one sound and then append a different one I'm not supposed to hear it? In other words I cannot reuse a Sink?

Yes!

TheEdward162:
So can we talk about why the API is designed in such way? It seems counter-intuitive that you get an unusable sink after you call .stop() on it. Am I supposed to create a new sink every time I want to restart playback? If so, why even expose .stop() method, why not just Drop, since the Sink is useless after calling it anyway?

mbillingr:
@TheEdward162 These are excellent questions.

Working around the `stop` == `drop` issue in ggez had me do some ugly things with [context and default devices](https://github.com/ggez/ggez/pull/437/files#diff-7cb74807e11b65d561faf0b6135433c6R224), that will probably bite someone in the future. This could have been easily avoided with different stop semantics.

[...]

PR #221, opened 2019-03-20
Immediately apply a Sink's state to the next Source in it's Queue.

willstott101:
**The Problem**
When progressing through Sources in a Sink's queue, it takes roughly 2.5ms (on a stereo stream, 5ms on mono) to query and apply the Sink's atomic state.

**Symptoms**
This means stopping a Sink with lots of Sources queued will result in a fuzz noise as a very short section of each queued Source is played one after the other.
This can also cause an almost imperceptibly small moment of sound when adding a Source to an already-paused Sink.
This can cause a very short (potentially very loud) moment of sound when switching to the next Source in a Sink's Queue when a volume is set. In fact abusing a Sink by queuing lots of very short Sources, whilst the Sink has a volume set, completely garbles the sound.

**My Solution**
~Existing implementations of [throttling functions](https://lodash.com/docs/4.17.11#throttle) sometimes have options to enable leading/trailing edge execution. I chose to create a new function which enables leading-edge execution for the throttle callback, as this seems the simplest and completely backwards compatible solution.~ _I've changed periodic_access to trigger it's first callback the first time the source is iterated._

I've also fixed the calculation of the access interval to take into account a Source's channel count. So that access times on mono and stereo streams are consistent.

### DIFF
--- a/src/sink.rs
+++ b/src/sink.rs
@@ -33,13 +33,20 @@ struct Controls {
 }
 
 impl Sink {
-    /// Builds a new `Sink`.
+    /// Builds a new `Sink`, beginning playback on a Device.
     #[inline]
     pub fn new(device: &Device) -> Sink {
-        let (queue_tx, queue_rx) = queue::queue(true);
+        let (sink, queue_rx) = Sink::new_idle();
         play_raw(device, queue_rx);
+        sink
+    }
 
-        Sink {
+    /// Builds a new `Sink`.
+    #[inline]
+    pub fn new_idle() -> (Sink, queue::SourcesQueueOutput<f32>) {
+        let (queue_tx, queue_rx) = queue::queue(true);
+
+        let sink = Sink {
             queue_tx: queue_tx,
             sleep_until_end: Mutex::new(None),
             controls: Arc::new(Controls {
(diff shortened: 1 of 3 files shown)


willstott101:
I hadn't realised that periodic_access was based on it's consumption rate and only applies to a playing Source.

This is just a quick proof of concept of adding a periodic access which is guaranteed to execute on the first sample, but otherwise is identical.

willstott101:
I've tried to tidy up my patch a bit. I've also edited my original PR comment to be a better summary of this PR.

sashahilton00:
Any reason this isn't being merged?

tomaka:
The fact that `periodic_access` doesn't access `Self` the first time feels more like a bug to me. I feel like we should just change `periodic_access` instead of adding `periodic_access_leading`.

willstott101:
Yeah I think that's much cleaner. I could just see use cases for both so wasn't sure if I ought to remove it. I'll update this shortly

willstott101:
I've added some tests and expanded the periodic access docs.

I'd understand if you'd rather not have the one-time access rate calculation documented though. Perhaps I ought to patch it to be calculated every access? That seems like the best compromise between flexibility and avoiding unnecessary computation.

willstott101:
Hmm, looking again at #171 makes me think I should break up the `pause` and `stop` `Sink` tests. More thoroughly testing `stop`. I would prefer to do that after this PR is merged though.

PR #223, opened 2019-04-05
Is it possible to provide a method to seek the sample?

heibor:
Is it possible to provide a method to seek the sample? Then we can jump to special time when play audio.


tomaka:
I feel like this is supposed to be an issue and not a PR. Closing.

PR #224, opened 2019-04-23
Use nalgebra for Spatial source

azriel91:
Hiya, I've swapped out `cgmath` for `nalgebra 0.18.0` (current latest) in an effort to reduce the number of dependencies that my project compiles.

The benefits of doing this, from a Rust ecosystem point of view, is that [we're converging on `nalgebra`](https://users.rust-lang.org/t/cgmath-looking-for-new-maintainers/20406/2) as the recommended math library (with full support from cgmath's author @brendanzab).

Alternatively we could just bump cgmath's version, but I'd prefer to switch the crate.

I've tested this on Ubuntu 18.10, and the volume fade in/out appears (sound-appears) to be the same when comparing before and after this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
+
 - Remove exclusive `&mut` borrow requirements in `Sink` & `SpatialSink` setters.
+- Use `nalgebra` instead of `cgmath` for `Spatial` source.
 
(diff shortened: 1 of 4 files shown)


azriel91:
@tomaka Hiya, any chance this could be merged and published?

azriel91:
Thank you very much 🙇

Issue #190, opened 2018-09-27
Can we implement `Source::total_duration` for non-wav sources?


icefoxen:
The question was brought up by this `ggez` issue:  https://github.com/ggez/ggez/issues/477

It does indeed appear to be correct that only the WAV decoder implements this method.  `rodio`'s stream-based API does seem like it might be problematic, but the method only exists when the stream is `Seek`, so I think we can manage it.

Looking at the various upstream crates, does the functionality exist to do this without having to decode the entire stream in the process?

 * `lewton`: https://docs.rs/lewton/0.9.1/lewton/audio/fn.get_decoded_sample_count.html
 * `minimp3`: Doesn't look like it, though I am pretty darn sure that the C library provides the capability to do this.  Just need to write a function that skims through the blocks in the stream reading the metadata without bothering to decode the individual blocks.
 * `claxon`: I think [this](https://docs.rs/claxon/0.4.1/claxon/metadata/struct.StreamInfo.html) has the relevant info, but it doesn't seem reliable (the `samples` field is an `Option`) and I'm not sure if there's a nicer way to get to it

ryankopf:
How is this still an issue?

gulis1:
My application needs `Source::total_duration` implemented for vorbis files. Can I work on a PR to implement this, using the `ogg_metadata` crate?

dvdsk:
hi @gulis good news, we actually have total_duration implemented for the `symphonia` decoders. It should just work. Let me know if it does not!

gulis1:
Hello @dvdsk, ty for answering so fast!!

I'm tried enabling the `symphonia` and `symphonia-all` features, but `rodio::Decoder` returns the following error:
```
thread 'main' panicked at src/main.rs:13:60:
called `Result::unwrap()` on an `Err` value: IoError("end of stream")
```
Without enabling the Symphonia features, the vorbis file (`/assets/music.ogg` from this repo) plays, but `Source::total_duration` does not work.

This is the code I'm using for the test:

```rust
use std::error::Error;
use std::io::BufReader;
use rodio::{OutputStream, Sink};

fn main() -> Result<(), Box<dyn Error>> {

    let (_stream, stream_handle) = OutputStream::try_default()?;
    let sink = Sink::try_new(&stream_handle)?;
       
    let file = std::fs::File::open("assets/music.ogg").unwrap();
    sink.append(rodio::Decoder::new(BufReader::new(file)).unwrap());
    sink.sleep_until_end();

[...]

PR #226, opened 2019-05-15
Implement `total_duration` for `FlacDecoder`

whentze:
As suggested in #190, this uses the `samples` field from Claxon's StreamInfo struct.
While that may be `None`, it appears to always be `Some` when decoding regular files (as opposed to potentially infinite streams).

I also added a test for this functionality.


tomaka:
Looks good, thanks!

PR #233, opened 2019-05-31
Add support for more FLAC files

cubetastic33:
Bump version of claxon in `Cargo.toml` to add support for FLAC files with LPC order > 12 (relevant issue [here](https://github.com/ruuda/claxon/issues/18))


cubetastic33:
I realized there could be breaking changes with just this change, and that the current version of radio itself pulls up a newer version of claxon.

PR #240, opened 2019-07-18
Update cpal -> 0.10

alexheretic:
* Provide device prevous method using `cpal::default_host()`
* Naively handle new cpal errors

If instead we'd want to reexport the `default_host` method, we'd have to change the `play_raw` behaviour, and that wouldn't be trivial as far as I can see.

This is a breaking change, but not very.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ documentation = "http://docs.rs/rodio"
 
 [dependencies]
 claxon = { version = "0.4.2", optional = true }
-cpal = "0.8"
+cpal = "0.10"
 hound = { version = "3.3.1", optional = true }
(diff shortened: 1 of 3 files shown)


est31:
> This is a breaking change, but not very.

Anything beyond that the origin of some reexported types has had a semver breaking change?

alexheretic:
> Anything beyond that the origin of some reexported types has had a semver breaking change?

No. But rodio re-exports cpal pretty freely, so I imagine that breaks in cpal generally mean a break in rodio. The `0.10` update is a very api breaking one, in general trying to propagate all errors to the caller.

alexheretic:
Thank you, it's good to see this project being maintained!

PR #244, opened 2019-08-09
Try all supported output formats when the default fails

### LINKED ISSUE #239: Panic reported when calling Sink::new
I have a bug report panic (Windows 10, Arctis Pro Wireless headset):
```
2019-07-17T19:48:00Z (main) called `Result::unwrap()` on an `Err` value: FormatNotSupported - stack backtrace:
...
6: std::panicking::rust_panic_with_hook (0x7ff7998d7dd5)
at /rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\/src\libstd\panicking.rs:478
7: std::panicking::continue_panic_fmt (0x7ff7998d78e4)
at /rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\/src\libstd\panicking.rs:381
8: std::panicking::rust_begin_panic (0x7ff7998d77c9)
at /rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\/src\libstd\panicking.rs:308
9: core::panicking::panic_fmt (0x7ff7998e54fc)
at /rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\/src\libcore\panicking.rs:85
10: core::result::unwrap_failed<cpal::CreationError> (0x7ff7995b28d1)
at /rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\src\libcore\macros.rs:18
11: core::result::Result<cpal::StreamId, cpal::CreationError>::unwrap (0x7ff7995b4385)
at /rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\src\libcore\result.rs:800
12: rodio::engine::new_output_stream (0x7ff7995b4385)
at C:/Users/alex/.cargo/registry/src/github.com-1ecc6299db9ec823/rodio-0.9.0/src/engine.rs:147
13: rodio::engine::start (0x7ff7993cbbeb)
at C:/Users/alex/.cargo/registry/src/github.com-1ecc6299db9ec823/rodio-0.9.0/src/engine.rs:112
14: rodio::engine::play_raw (0x7ff7993cbbeb)
at C:/Users/alex/.cargo/registry/src/github.com-1ecc6299db9ec823/rodio-0.9.0/src/engine.rs:51
15: rodio::sink::Sink::new (0x7ff7993cbbeb)
at C:/Users/alex/.cargo/registry/src/github.com-1ecc6299db9ec823/rodio-0.9.0/src/sink.rs:40
...
```

So `Sink::new(&device)` is panicking here.

alexheretic:
This pr attempts to improve/fix situations that cause #239.

When `build_output_stream` fails with the default format, try each supported format to see if it will work before panicking.

I'm not too experienced with audio stuff so I'm not totally sure if this is the way to go about it. But on the other hand, the new code only executes in the case the current code would have panicked, so in a way it "can't hurt" to try.

Resolves #239

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -188,7 +188,7 @@ fn find_working_output_stream(
         .into_iter()
         .flat_map(|sf| {
             let max_rate = sf.max_sample_rate;
-            let min_rate = sf.max_sample_rate;
+            let min_rate = sf.min_sample_rate;
             let mut formats = vec![sf.clone().with_max_sample_rate()];
             if HZ_44100 < max_rate && HZ_44100 > min_rate {
                 formats.push(sf.clone().with_sample_rate(HZ_44100))
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -138,21 +138,28 @@ where
 }
 
 // Adds a new stream to the engine.
-// TODO: handle possible errors here
 fn new_output_stream(
-    engine: &Arc<Engine>, device: &Device,
+    engine: &Arc<Engine>,
+    device: &Device,
 ) -> (Arc<dynamic_mixer::DynamicMixerController<f32>>, StreamId) {
-    // Determine the format to use for the new stream.
-    let format = device
-        .default_output_format()
-        .expect("The device doesn't support any format!?");
+    let (format, stream_id) = {
+        // Determine the format to use for the new stream.
+        let default_format = device
+            .default_output_format()
+            .expect("The device doesn't support any format!?");
 
-    let stream_id = engine
-        .events_loop
-        .build_output_stream(device, &format)
-        .unwrap();
-    let (mixer_tx, mixer_rx) =
-        { dynamic_mixer::mixer::<f32>(format.channels, format.sample_rate.0) };
+        match engine
+            .events_loop
+            .build_output_stream(device, &default_format)
+        {
+            Ok(sid) => (default_format, sid),
+            Err(err) => find_working_output_stream(engine, device)
+                .ok_or(err)
+                .expect("build_output_stream failed with all supported formats"),
+        }
+    };
+
+    let (mixer_tx, mixer_rx) = dynamic_mixer::mixer::<f32>(format.channels, format.sample_rate.0);
 
     engine
         .dynamic_mixers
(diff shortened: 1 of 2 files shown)


src/engine.rs
```diff
@@ -162,3 +169,57 @@ fn new_output_stream(
...
+        .supported_output_formats()
+        .expect("No supported output formats")
+        .collect();
+    supported.sort_by(|a, b| b.cmp_default_heuristics(a));
+
+    supported
+        .into_iter()
+        .flat_map(|sf| {
+            let max_rate = sf.max_sample_rate;
+            let min_rate = sf.max_sample_rate;
```
est31: min_sample_rate

src/engine.rs
```diff
@@ -138,21 +138,28 @@ where
 }
 
 // Adds a new stream to the engine.
-// TODO: handle possible errors here
```
est31: They aren't really handled well with Result so I wouldn't remove this TODO yet :)

src/engine.rs
alexheretic: nice one thanks

src/engine.rs
alexheretic: Aren't they all handled now consistently with rodio's fairly panic happy API? This does remove the unwrap that clearly could fail. Now we have panics that we generally wouldn't expect to happen, similar to elsewhere.

alexheretic:
> I wonder: why do you do this 44100 sample rate computation? What's the reason?

It's just from my limited experience that 44100 is commonly used so maybe will work, yet also may not be the max supported. So in the cases where it's in the `(min, max)` range I thought it wouldn't hurt to try it.

It's an annoying API to iterate through formats that may or may not work since it only provides min & max. Maybe we could add a method to `SupportedFormat` that returns a bunch of standard formats from min to max in a better way?

est31:
Yeah optimally we'd go through a list of formats [like cpal does](https://github.com/RustAudio/cpal/blob/e4ea5e9d0696ae3a7d6b6752ea92d3dcc0db3d35/src/host/alsa/mod.rs#L313) but I guess this would also involve even more attempts. Idk, I guess this version is okay for now.

PR #247, opened 2019-08-23
Rework to use cpal-next

alexheretic:
~Reworks to use cpal with the event loop removed https://github.com/RustAudio/cpal/pull/301.~

Reworks to use current cpal master, removing the engine and static parts of the library, now the `cpal::Stream` cannot be shared between threads.

If we go this way we'll need to take another look at the docs.


alexheretic:
I've reworked this against cpal master. Static/global playback is a dead end with the new cpal code, so I replaced the old API with a wrapped device variant.

```rust
// old
rodio::play_once(&cpal_device, source);

// new
rodio_device.play_once(source);
```

alexheretic:
So yes the `cpal::Stream` is immovable, latest code has a movable handle type to go with it and be more flexible. Though the immovable stream needs to not be dropped.

```rust
let (stream, handle) = rodio::OutputStream::try_default()?;
handle.play_once(source)?
```

alexheretic:
I'm still keeping this up to date with cpal git, awaiting the next release (https://github.com/RustAudio/cpal/issues/429). This change is a big improvement to debug build performance, at least on my hardware.

est31:
The cpal release is out!

alexheretic:
Great this is ready for review after a long wait for cpal 0.12. The main change is:

> So yes the `cpal::Stream` is !Send, but latest code has a Send handle type to go with it and be more flexible. Though the immovable stream needs to not be dropped.
> 
> ```rust
> let (stream, handle) = rodio::OutputStream::try_default()?;
> handle.play_once(source)?
> ```

We can probably do more to document the immovable `OutputStream` + more flexible `OutputStreamHandle`.

PR #249, opened 2019-09-13
Generate samples at a multiple of channels

Cocalus:
I ran into an issue where pausing and unpausing would swap the channels of a stereo source.  It was because the number of samples the pause source generated while paused wasn't a multiple of 2. This has two fixes for that issue. 

One Pausable now always generates Zero's in a multiple of the number of channels.
Two Periodic only changes things at a multiple of the number of channels. This might help other similar issues.

I also fixed Delay since it has the same potential issue.

I suspect channel ordering issues can also occur if current_frame_len isn't a multiple of the channels, when called before iterating. Also if the iterator returns a None that's not aligned with current_frame_len. Those could cause the issue with Repeat, Stop and various ways of chaining Sources.

### DIFF
--- a/src/source/pausable.rs
+++ b/src/source/pausable.rs
@@ -4,26 +4,46 @@ use Sample;
 use Source;
 
 /// Internal function that builds a `Pausable` object.
-pub fn pausable<I>(source: I, paused: bool) -> Pausable<I> {
+pub fn pausable<I>(source: I, paused: bool) -> Pausable<I>
+where
+    I: Source,
+    I::Item: Sample,
+{
+    let paused_channels = if paused {
+        Some(source.channels())
+    } else {
+        None
+    };
     Pausable {
         input: source,
-        paused: paused,
+        paused_channels: paused_channels,
+        remaining_paused_samples: 0,
     }
 }
 
 #[derive(Clone, Debug)]
 pub struct Pausable<I> {
     input: I,
-    paused: bool,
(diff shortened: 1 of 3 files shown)


src/source/pausable.rs
```diff
@@ -4,26 +4,46 @@ use Sample;
...
+    I: Source,
+    I::Item: Sample,
+{
     /// Sets whether the filter applies.
     ///
     /// If set to true, the inner sound stops playing and no samples are processed from it.
     #[inline]
     pub fn set_paused(&mut self, paused: bool) {
-        self.paused = paused;
+        // Minimize calls to channels by only calling on state change
```
est31: This comment is wrong. You always call the function when paused is true, even if it was true before. In general, `set_paused` shouldn't be called often so minimizing calls shouldn't be needed. I'd say you should just remove the comment.

src/source/pausable.rs
```diff
@@ -4,26 +4,46 @@ use Sample;
...
+{
+    let paused_channels = if paused {
+        Some(source.channels())
+    } else {
+        None
+    };
     Pausable {
         input: source,
-        paused: paused,
+        paused_channels: paused_channels,
```
est31: Please use init shorthand here.

Cocalus:
I think I fixed those issues. Long term using something like the sample crates Frame trait would avoid these issues at the type level, but that's a big change to the API.

PR #250, opened 2019-09-20
.gitignore the IntelliJ IDEA project files

burjui:


est31:
IDE/editor related files are best added to the user-specific gitignore, not to projects.

burjui:
Thanks for pointing that out, I didn't know about `$GIT_DIR/info/exclude`

est31:
I rather meant using this mechanism but I guess your method works too:

https://git-scm.com/docs/gitignore/1.7.12#_description

> Patterns which a user wants git to ignore in all situations (e.g., backup or temporary files generated by the user’s editor of choice) generally go into a file specified by core.excludesfile in the user’s ~/.gitconfig. Its default value is $XDG_CONFIG_HOME/git/ignore. If $XDG_CONFIG_HOME is either not set or empty, $HOME/.config/git/ignore is used instead.

burjui:
It would make sense to add `.idea` to `$HOME/.config/git/ignore`, but I would still need to add `<project name>.iml` in every project's `$GIT_DIR/info/exclude`. Just ignoring `*.iml` globally would be too much, as this file extension can be potentially used by other software.

OvermindDL1:
> Just ignoring *.iml globally would be too much, as this file extension can be potentially used by other software.

Not really, once a file is tracked then it is fine, and it can always be manually added as well.

burjui:
Yes, but it's too easy to forget to add the file you created in the repo, if it's ignored by default and is not shown by `git status`. I guess for people with normal attention and memory this is not a problem, but it happened to me before, so I just prefer to play safe. Anyway, it's just a single shell command (`echo project_name.iml > $GIT_DIR/info/exclude`), which will be saved in shell's history, so I won't be even typing the full command in the future thanks `Ctrl+R` magic.

PR #259, opened 2019-11-18
dependencies: update lazy_static, hound and minimp3

tirz:
Upgrade dependencies:
 - lazy_static from 1.0.0 to 1.4.0
 - hound from 3.3.1 to 3.4.0
 - minimp3 from 0.3.2 to 0.3.3

Also reorganize the dependencies for easier readability - it was already readable since the list of dependencies is pretty small... Anyway.
They are still ordered alphabetically, but firstly, we sort the "dependencies" alphabetically and then we sort the "optional dependencies" alphabetically.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,17 +9,17 @@ repository = "https://github.com/RustAudio/rodio"
 documentation = "http://docs.rs/rodio"
 
 [dependencies]
-claxon = { version = "0.4.2", optional = true }
 cpal = "0.10"
-hound = { version = "3.3.1", optional = true }
-lazy_static = "1.0.0"
+lazy_static = "1.4.0"
+claxon = { version = "0.4.2", optional = true }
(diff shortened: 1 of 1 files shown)


est31:
The versions mentioned in Cargo.toml only give a range. Cargo is allowed to use any newer semver-compatible version and in fact will try to use the newest version available. What this PR does is a reduction of the window of versions rodio is compatible with. Which isn't an improvement. Closing.

tirz:
@est31 not really, only lewton in pinned to "0.9", the others dependencies have a minor version and hound is even pinned to "3.3.1" while "3.4" is out.
Is your prefer I can send a PR with all the mimors versions removed.

tirz:
I did not know about https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements nether that it apply by default.
Tank you :)

Issue #192, opened 2018-10-15
Is a scheduled-playback-feature and retrieving playback position planned?


kerskuchen:
It would be very useful (i.e. for rhythm games and effects like [mickey-mousing](https://en.wikipedia.org/wiki/Mickey_Mousing)) to play sounds starting at a certain sample-index and also to retrieve the index of the last sample that was committed to cpal.

An example of an API that does this would be Unity3D's [AudioSource.PlaySheduled](https://docs.unity3d.com/ScriptReference/AudioSource.PlayScheduled.html) function and its [AudioSettings.dspTime](https://docs.unity3d.com/ScriptReference/AudioSettings-dspTime.html)

Combining those two APIs make the gameplay in the following 1 minute [video](https://www.youtube.com/watch?v=DTgn_bKCWyg) possible.

Building this from scratch over cpal would be possible but tedious and would duplicate the existing functionality of rodio in every case except for the above API additions. So I thought I just ask if are any plans to do this already :)

yong-jie:
@PaddyCo Any plans to do a PR for this? I think this would be a great addition that could benefit those who are requesting for the feature!

ngmiller:
@PaddyCo +1 for a PR for the elapsed feature. I used your branch in a private project and it worked great. Would love to see this make it into master.

PaddyCo:
Oh sorry I totally forgot about this, I'm gonna try and find some time this weekend and do some clean-up if necessary and then I'll open a PR 😅

TonalidadeHidrica:
Hi, I really want want the "playback position" feature for my project and waiting for the pull request to be merged.  By the way, are there any info about scheduled playback yet?

temportalflux:
Has there been any movement on this, or has this request been discarded?

PR #273, opened 2020-03-01
Add Elapsed filter

PaddyCo:
as discussed in #192, a way to get the current elapsed time is almost required for things that are supposed to be synced to the audio (rhythm games for example)

I tried solving this by making a filter, and it appears to work pretty well!

I'm not sure this is the most idiomatic approach to the problem (I'll be honest, not super familiar with rodio, or audio in general for that matter) so if you have any suggestions or better ways to solve this, shoot :smile:


TonalidadeHidrica:
I think this implementation has several problems:
* The duration updates only when next iteration of a chunk occurs.  This is a problem when we want more precise timestamp than the chunk size.
* Audio timer is not necessarily synchronized with system timer, and nor is the timing when the `next` called.   Fortunately, now that the new feature of tracking precise playing/recording position is introduced in RustAudio/cpal#397, we should make use of this instead.

JackRedstonia:
Hello, any progress on this? Currently implementing a rhythm game using rodio and I'm hitting on this roadblock. Should I go ahead and open a PR utilising [RustAudio/cpal#397](https://github.com/RustAudio/cpal/pull/397) instead?

TonalidadeHidrica:
@JackRedstonia Although there's no direct way to get the elapsed time info, current API already enables us getting it.  I'm developing a rhythm game too and already succeeded in implemented precise a/v sync feature.  You can see my code in my repository: TonalidadeHidrica/taiko-untitled .   By the way if you want to implement this feature you can go ahead, I think.  (I'm not an owner of this repo though)

JackRedstonia:
@TonalidadeHidrica I looked through the code and it seems that it only works for one music track at a time, which isn't exactly what I need, I think. I also tried to add the functionality into the rodio crate for the PR but ended up nuking it all as it was too complex.

I believe I should leave this up to someone with better understanding of rodio and cpal.

TonalidadeHidrica:
@JackRedstonia I agree that current rodio interface is, although quite easy when performing "fuzzy" playback like music player, not really cut out for precise, framewise playback, when you want to build rhythm game or DTM software for example.  That's why I only used rodio for decoding several audio data, and write to audio stream with cpal directly by myself.

Just for curiosity, what is your specific usecase you thought that my code lacks?   (Should I refrain from discussing here?)

JackRedstonia:
@TonalidadeHidrica I'm making a game framework and I'm looking to have the ability to have any (reasonable) number of tracks playing and to be able to get their playing positions, so I'll have to change quite a lot of things to get that working. Nevertheless, the information about [RustAudio/cpal#397](https://github.com/RustAudio/cpal/pull/397) was quite useful, so much thanks for that.
Hopefully I'll come up with a decent way to get the above idea into decent code, but for now I'll experiment with another library called BASS. (not that rodio and cpal is bad of course, I just want to see what the available options are, and I'll likely not use BASS anyways due to its horrendous licensing)

dvdsk:
This has been implemented slightly differently in rodio 0.20 as [TrackPosition](https://docs.rs/rodio/latest/rodio/source/trait.Source.html#method.track_position)

Issue #283, opened 2020-04-04
Stack overflow on audio change


mathstuf:
Not sure what's up here (backtrace below). I don't see repeating pointer values in the first few pages, so not obviously a refcycle. But, you can see that we're *way* deep in a call stack here. The `Arc::drop_slow` is the set of frames seen over and over.

```
#53678 0x00005642f4b1d23b in alloc::sync::Arc<T>::drop_slow (self=0x5642f52a9870) at /builddir/build/BUILD/rustc-1.42.0-src/src/liballoc/sync.rs:739
#53679 0x00005642f4b1dfd6 in <alloc::sync::Arc<T> as core::ops::drop::Drop>::drop (self=0x5642f52a9870) at /builddir/build/BUILD/rustc-1.42.0-src/src/liballoc/sync.rs:1249
#53680 0x00005642f4b27c0e in core::ptr::drop_in_place () at /builddir/build/BUILD/rustc-1.42.0-src/src/libcore/ptr/mod.rs:174
#53681 0x00005642f4b27ebe in core::ptr::drop_in_place () at /builddir/build/BUILD/rustc-1.42.0-src/src/libcore/ptr/mod.rs:174
#53682 0x00005642f4b29a79 in core::ptr::drop_in_place () at /builddir/build/BUILD/rustc-1.42.0-src/src/libcore/ptr/mod.rs:174
#53683 0x00005642f4b28bce in core::ptr::drop_in_place () at /builddir/build/BUILD/rustc-1.42.0-src/src/libcore/ptr/mod.rs:174
#53684 0x00005642f4b28212 in core::ptr::drop_in_place () at /builddir/build/BUILD/rustc-1.42.0-src/src/libcore/ptr/mod.rs:174
#53685 0x00005642f4b1d23b in alloc::sync::Arc<T>::drop_slow (self=0x5642f51bcee0) at /builddir/build/BUILD/rustc-1.42.0-src/src/liballoc/sync.rs:739
#53686 0x00005642f4b1dfd6 in <alloc::sync::Arc<T> as core::ops::drop::Drop>::drop (self=0x5642f51bcee0) at /builddir/build/BUILD/rustc-1.42.0-src/src/liballoc/sync.rs:1249
#53687 0x00005642f4b27c0e in core::ptr::drop_in_place () at /builddir/build/BUILD/rustc-1.42.0-src/src/libcore/ptr/mod.rs:174
[...]

PR #285, opened 2020-04-22
Fix stack overflow on audio change

gridbugs:
The types `Frame` and `FrameData` are mutually recursive, and the
incidental linked lists that can be formed as a result can be long (at
least in the order of thousands of elements). As a result, when a frame
is deallocated, rust appears to recursively call `drop_in_place` down
the list, causing stack overflows for long lists.

Addresses https://github.com/RustAudio/rodio/issues/283

This problem is hard to reproduce. I've tried constructing toy programs to no avail. The only way I've managed to trigger it is running https://github.com/stevebob/slime99/ `graphical` or `ansi-terminal` crates with the `--audio` flag (audio is disabled by default because of exactly this problem). Start a new game and leave it on the first floor for ~5 minutes, then walk down the stairs. The track will switch and _sometimes_ there will be a stack overflow on the "rodio audio processing" thread.

I was able to reliably reproduce the problem last night, but I'm trying to again now (after changing nothing) and it's not happening. Something something phase of the moon. In other words, I'm not satisfied with my testing of this patch thus far. I'm opening this PR now to start a discussion. Is there a way to construct a long chain of `Frame`->`FrameData`->`Frame`->...? I think doing so would reliably trigger this problem.

### DIFF
--- a/src/source/buffered.rs
+++ b/src/source/buffered.rs
@@ -1,4 +1,5 @@
 use std::cmp;
+use std::mem;
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
@@ -64,6 +65,40 @@ where
     channels: u16,
     rate: u32,
     next: Mutex<Arc<Frame<I>>>,
+}
+
+impl<I> Drop for FrameData<I>
+where
+    I: Source,
+    I::Item: Sample,
+{
+    fn drop(&mut self) {
+        // This is necessary to prevent stack overflows deallocating long chains of the mutually
+        // recursive `Frame` and `FrameData` types. This iteratively traverses as much of the
+        // chain as needs to be deallocated, and repeatedly "pops" the head off the list. This
+        // solves the problem, as when the time comes to actually deallocate the `FrameData`,
+        // the `next` field will contain a `Frame::End`, or an `Arc` with additional references,
+        // so the depth of recursive drops will be bounded.
+        loop {
+            if let Ok(arc_next) = self.next.get_mut() {
+                if let Some(next_ref) = Arc::get_mut(arc_next) {
+                    // This allows us to own the next Frame.
+                    let next = mem::replace(next_ref, Frame::End);
+                    if let Frame::Data(next_data) = next {
+                        // Swap the current FrameData with the next one, allowing the current one
+                        // to go out of scope.
+                        *self = next_data;
+                    } else {
+                        break;
+                    }
+                } else {
+                    break;
+                }
+            } else {
+                break;
+            }
+        }
+    }
 }
 
 /// Builds a frame from the input iterator.


stevebob:
Ok I managed to reproduce it as per the above. My goal is to determine a correlation between the length of `Frame`->`FrameData` chain, and stack overflows. Then demonstrate that after my change, chains of problematic lengths no longer cause stack overflows.

I added the following right at the start of `FrameData`'s drop:
```rust
    fn drop(&mut self) {
        let mut count = 0;
        let mut cursor = self.next.lock().unwrap().clone();
        loop {
            if let Frame::Data(data) = &*cursor.clone() {
                let next = data.next.lock().unwrap();
                cursor = next.clone();
            } else {
                break;
            }
            count += 1;
        }
        println!("count: {}", count);
        return;
       ...
```
...to measure the length of the chain. As an added bonus, the above code runs every time one of the nested `Frame`s is dealloc'd, so I can see how far it got at the point where it crashed.

No crash:
- 5645
- 5530

Crash:
- 12136 (crashed with 4476 to go (7760 complete))
- 9853 (crashed with 2193 to go (_also_ 7760 complete))
- 12342 (crashed with 4682 to go (7760 complete))

Then I got rid of the `return` statement at the end of the block above, so the rest of `drop` can run but it still prints out the count. I saw a case where a chain of 11681 was dealoc'd, and it didn't stack overflow. Ran it again with 9126 chain length, and still no crash.

src/source/buffered.rs
```diff
@@ -66,6 +67,40 @@ where
...
+        // so the depth of recursive drops will be bounded.
+        loop {
+            if let Ok(arc_next) = self.next.get_mut() {
+                if let Some(next_ref) = Arc::get_mut(arc_next) {
+                    // This allows us to own the next Frame.
+                    let next = mem::replace(next_ref, Frame::End);
+                    if let Frame::Data(next_data) = next {
+                        // Swap the current FrameData with the next one, allowing the current one
+                        // to go out of scope.
+                        mem::replace(self, next_data);
```
nicokoch, line 90: This will trigger a warning on nightly rust, see rust-lang/rust#71256
You can just assign instead.
`*self = next_data`

stevebob:
Squashed fix up commits and rebased onto the tip of master

stevebob:
@nicokoch any thoughts on whether this change is ok to merge? Would you like me to rebase it onto master?

mathstuf:
Bump?

est31 (changes requested):
Ok except for detail. Sorry for the delay.

est31 (approved):
This has sat unreviewed for more than a year. I'm merging despite the small issue. Subsequent PRs for improvement are welcome!

gridbugs:
Thanks for the merge @est31! I'll make the change you suggested in a subsequent PR.

PR #287, opened 2020-05-14
Clear and skip

ChiefMilesEdgeworth:
This PR still has some room for work, but this gives basic functionality for `clear()`ing and `skip_one()`ing `Sink`s. This seems like very useful functionality, and helps make the library more accessible.

There's also some simple formatting stuff done here, since `cargo-fmt` didn't work when I was using it due to an error in the `.rustfmt.toml` file. That's a simple enough to change to something else if need be :) 

Closes? #171

### DIFF
--- a/src/queue.rs
+++ b/src/queue.rs
@@ -94,16 +94,15 @@ where
             .store(keep_alive_if_empty, Ordering::Release);
     }
 
-    pub fn clear(&self) {
+    /// Removes all the sounds from the queue. Returns the number of sounds cleared.
+    pub fn clear(&self) -> usize {
         let mut sounds = self.next_sounds.lock().unwrap();
+        let len = sounds.len();
         sounds.clear();
+        len
     }
 
-    pub fn len(&self) -> usize {
-        self.next_sounds.lock().unwrap().len()
-    }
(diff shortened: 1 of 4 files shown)


ChiefMilesEdgeworth:
Just saw a different PR where there was a comment to not fix `cargo-fmt` in a PR that does functionality. I'm reverting the commit that ran `cargo-fmt`.

tralph3:
Please for the love of god merge this! I'm really needing this *basic* functionality!

tramhao:
I tested this clear function, but after clear, the player will always generate an alsalib buffer underrun warning. Without clear it's no problem. my code is in https://github.com/tramhao/termusic/tree/dev/src .

ivakam:
Is there any status on this? If there's a blocker somewhere, I'm willing to lend a hand in any way I can. This functionality seems essential to a library like this.

ChiefMilesEdgeworth:
I think it's been too long since I first put this together, so if someone else wants to take this branch and run with it I would welcome them to it. I'm going to close this PR though.

ivakam:
This PR is continued by me in #442.

PR #288, opened 2020-05-16
edition = 2018; dependencies updated; associated changes

paulirotta:
I had some audible artifacts. Not sure if this affects it, but a refresh seemed in order


est31:
#247 has done most of this

Issue #229, opened 2019-05-16
Panics when finish playing MP3 on Windows in debug mode


azriel91:
When playing an mp3 on Windows in debug mode, sometimes `self.next_frame()` is not empty in AAA, leading to this panic:

```
thread 'rodio audio processing' panicked at 'assertion failed: self.next_frame.is_empty()', D:\work\github\tomaka\rodio\src\conversions\sample_rate.rs:193:13
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

Subsequent sounds are no longer played because the `ENGINE` is a static variable and doesn't re-initialize itself.

The line that panics is in `SampleRateConverter`:

https://github.com/tomaka/rodio/blob/92a018bf537a3c571baea1a1084d750416d2249e/src/conversions/sample_rate.rs#L186

`self.next_frame()` is empty `[]` while the sound is playing, but when the sound stops, it had two values, example:

```
// needed to add `Debug` bound in many places
[-0.046173096, -0.046173096]
```

**Stack backtrace**

Note: line numbers may not be accurate because I was munging code to get the information.

```
stack backtrace:
   0: rodio::conversions::sample_rate::{{impl}}::next<rodio::source::uniform::Take<rodio::source::samples_converter::SamplesConverter<rodio::decoder::Decoder<std::io::cursor::Cursor<amethyst_audio::source::Source>>, f32>>> (0x7ff60ca4b213)
             at D:\work\github\tomaka\rodio\src\conversions\sample_rate.rs:188
   1: rodio::conversions::channels::{{impl}}::next<rodio::conversions::sample_rate::SampleRateConverter<rodio::source::uniform::Take<rodio::source::samples_converter::SamplesConverter<rodio::decoder::Decoder<std::io::cursor::Cursor<amethyst_audio::source::Sourc (0x7ff60ca48bec)
             at D:\work\github\tomaka\rodio\src\conversions\channels.rs:62
   2: rodio::conversions::sample::{{impl}}::next<rodio::conversions::channels::ChannelCountConverter<rodio::conversions::sample_rate::SampleRateConverter<rodio::source::uniform::Take<rodio::source::samples_converter::SamplesConverter<rodio::decoder::Decoder<std (0x7ff60ca49203)
             at D:\work\github\tomaka\rodio\src\conversions\sample.rs:38
[...]

PR #289, opened 2020-05-20
Remove debug_assert which was triggering unnecessary panic

gavynriebau:
When playing an mp3 on Windows in debug mode, sometimes
self.next_frame() is not empty in SampleRateConverter, leading to a
panic

See issue: #229


est31 (approved):
Thanks!

Issue #293, opened 2020-06-17
How to detect when the Decoder has reached the end of the audio file?


Boscop:
I'm using this function that should return `true` when the decoder has reached the end of the audio file:
```rust
fn sum_channels_return_reached_end(
	channel_count: usize,
	decoder: &mut Decoder<BufReader<File>>,
	output: &mut [f32],
) -> bool {
	let mut reached_end = true;
	for (chunk, dst) in decoder.chunks(channel_count).into_iter().zip(output.iter_mut()) {
		let (sum, count) = chunk.map(sample_conv).fold((0., 0), |(s, c), v| (s + v, c + 1));
		*dst = sum / count as f32;
		reached_end = false;
	}
	reached_end
}
```
It works with wav files but always returns `false` for mp3 files: 
The decoder iterator still yields values after the end of the mp3 file. I tested with several mp3 files, all behave the same.
Is there a reliable way to detect when the decoder has reached the end of the audio file?

est31:
`next_frame` seems to always return Some. So this is a minimp3 bug, not ours.

est31:
Maybe our bug after all. #297 has been merged. Could you try out latest master of rodio and check whether the issue still persists?

roderickvd:
Today there are several ways of knowing for sure:
- when `current_span_len` returns `Some(0)`
- when `size_hint` returns `(0, Some(0))`
- by calling `peek()`

PR #297, opened 2020-07-07
Fix mp3 decoder bug: only reset offset after new frame

skomski:
I stumbled upon it using the decoder directly. It could explain #229 and #293.


est31:
Hmm yeah this seems like an issue.

PR #302, opened 2020-07-13
Update sink.rs

derekdreery:
I struggled for quite a while to work out why my sounds weren't playing. I think this is the cause. The new docs will save time for future users of the crate.

### DIFF
--- a/src/sink.rs
+++ b/src/sink.rs
@@ -15,6 +15,9 @@ use crate::Source;
 ///
 /// Dropping the `Sink` stops all sounds. You can use `detach` if you want the sounds to continue
 /// playing.
+///
+/// Once the sink has finished playing all sounds in its queue, the handle stops looking for more sounds, and so the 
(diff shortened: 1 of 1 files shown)


src/sink.rs
```diff
@@ -15,6 +15,9 @@ use crate::Source;
 ///
 /// Dropping the `Sink` stops all sounds. You can use `detach` if you want the sounds to continue
 /// playing.
+///
+/// Once the sink has finished playing all sounds in its queue, the handle stops looking for more sounds, and so the 
+/// `Sink` will stop working. Create a new `Sink` to continue playing sounds.
```
CleanCut, line 20:
```suggestion
/// Once the sink has finished playing all sounds in its queue, it will stop. Call the `append` method
/// to provide it with additional sounds to play.
```

derekdreery:
I've lost track of this. I'll reopen if necessary.

PR #309, opened 2020-08-08
Add SkipDuration source.

ely-uf:
`SkipDuration` is a source that skips specified duration of the given
source from it's current position.

The already existing `TakeDuration` source allows to truncate the given source
from the end. `SkipDuration` complements it, introducing the option of
truncating the given source from it's current position.

### DIFF
--- a/src/source/skip.rs
+++ b/src/source/skip.rs
@@ -231,14 +231,14 @@ mod tests {
     #[test]
     fn skip_duration_equal_to_source_length() {
         skip_duration_test_block! {
-            channels: 1, sample rate: 44100, seconds: 1, seconds to skip: 5;
-            channels: 1, sample rate: 96000, seconds: 10, seconds to skip: 11;
+            channels: 1, sample rate: 44100, seconds: 1, seconds to skip: 1;
+            channels: 1, sample rate: 96000, seconds: 10, seconds to skip: 10;
 
-            channels: 2, sample rate: 44100, seconds: 1, seconds to skip: 5;
-            channels: 2, sample rate: 96000, seconds: 10, seconds to skip: 11;
+            channels: 2, sample rate: 44100, seconds: 1, seconds to skip: 1;
+            channels: 2, sample rate: 96000, seconds: 10, seconds to skip: 10;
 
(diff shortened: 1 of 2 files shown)


ely-uf:
Hello,

The initial use-case that became the motive of that PR is the need to loop a segment of an audio file that is _not_ located in the beginning.
It's possible to loop a segment in the beginning using `TakeDuration` and `Repeat` sources, but I was not able to find a way to skip to a certain point in the audio stream.

I would appreciate any comments about correctness and how this fits into the library.

Thanks.

src/source/skip.rs
```diff
@@ -0,0 +1,244 @@
...
+            $(
+                test_skip_duration_samples_left($ch, $sr, $sec, $sec_to_skip);
+            )+
+        }
+    }
+
+    #[test]
+    fn skip_duration_shorter_than_source() {
+        skip_duration_test_block! {
+            channels: 1, sample rate: 44100, seconds: 5, seconds to skip: 3;
```
ely-uf, line 192: The format of the tests is subject to discussion.

The main rationale for using a macro rule to generate these cases is to provide an obvious tag for every numeric parameter.

Nonetheless, this approach adds it's own mini-DSL that can be avoided by using plain rust function calls:
```rust
test_skip_duration_samples_left(1, 44100, 5, 3);
test_skip_duration_samples_left(1, 96000, 5, 3);
...
``` t

File: src/source/skip.rs [hunk 2]
```diff
@@ -0,0 +1,244 @@
...
+    fn total_duration(&self) -> Option<Duration> {
+        self.input.total_duration().map(|val| {
+            val.checked_sub(self.skipped_duration)
+                .unwrap_or(Duration::from_secs(0))
+        })
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
ely-uf, line 156: Current tests only verify `SkipDuration` behavior for audio sources with constant sample rate and number of channels.

I'm working on adding tests that cover the behavior with sources that may change sample rate or number of channels.
Until now, the behavior was verified using example `mp3` files and a stopwatch :)

est31 (approved):
Thanks!

PR #311, opened 2020-08-20
Make docs easier for beginners

### LINKED ISSUE #304: Docs example doesn't play sound on Windows
Hi y'all, I'm trying to run this example and it compiles but no sound plays.

```rust
use rodio::Sink;

fn main() {
    let device = rodio::default_output_device().unwrap();
    let sink = Sink::new(&device);

    let source = rodio::source::SineWave::new(440);
    sink.append(source);
}
```

I'm on this toolchain
```
nightly-x86_64-pc-windows-msvc (default)
rustc 1.46.0-nightly (feb3536eb 2020-06-09)
```

And this Rodio version
`rodio = "0.11.0"`

How do you troubleshoot this?

savovsa:
@vladinator1000 and I thought it would be easier for beginners if the docs worked out of the box.

Should be enough to close #304

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -20,11 +20,19 @@
 //! use std::io::BufReader;
 //! use rodio::Source;
 //!
-//! let (stream, stream_handle) = rodio::OutputStream::try_default().unwrap();
+//! fn main() {
+//!    let (stream, stream_handle) = rodio::OutputStream::try_default().unwrap();
 //!
-//! let file = File::open("sound.ogg").unwrap();
-//! let source = rodio::Decoder::new(BufReader::new(file)).unwrap();
-//! stream_handle.play_raw(source.convert_samples());
+//!    // Load a sound from a file, using a path relative to cargo.toml
+//!    let file = File::open("sound.ogg").unwrap();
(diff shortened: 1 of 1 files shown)


vladinator1000:
This was fun! We made sure to test that it runs using the example audio files, and it does. 😍

src/lib.rs
```diff
@@ -20,11 +20,19 @@
...
 //!
-//! let (stream, stream_handle) = rodio::OutputStream::try_default().unwrap();
-//!
-//! let file = File::open("sound.ogg").unwrap();
-//! let source = rodio::Decoder::new(BufReader::new(file)).unwrap();
-//! stream_handle.play_raw(source.convert_samples());
+//! fn main() {
+//!    let (stream, stream_handle) = rodio::OutputStream::try_default().unwrap();
+//!
+//!    // Load a sound from a file, using a path relative to cargo.toml
```
est31, line 26: `Cargo.toml`

vladinator1000, line 26:
```suggestion
//!    // Load a sound from a file, using a path relative to Cargo.toml
```

PR #314, opened 2020-08-27
Add `LoopedDecoder`

exrook:
`LoopedDecoder` works like the normal `Decoder`, but when the decoder
reaches the end of file it will restart playback from the beginning.
This has the advantage over `Decoder::new().repeat_infinite()` that it
will not buffer the entire uncompressed audio file into memory.

~~The implementation of `LoopedDecoder` for MP3 files is dependent on this upstream PR: https://github.com/germangb/minimp3-rs/pull/23~~


est31 (approved):
Looks good, thanks!

PR #317, opened 2020-10-01
Add new_X format specific methods to Decoder

vickles:
In my project I want to decode data only if it matches an expected file format. I can't see anything in the current API to either retrieve the file format after decoding has started, or to only attempt to decode a specific format.

This is a small addition to the API, to add per file-format decoding methods so that I can ensure that the data matches a specified format.


est31 (approved):
Thanks!

PR #320, opened 2020-10-02
Fix crackling audio when using Sink

### LINKED ISSUE #225: Crackly audio on OSX
The Amethyst game engine has been receiving reports of crackly distorted audio on OSX. Has this been extensively tested on OSX? One contributor believed this may be due to Mutex use in rodio and cpal, possibly blocking the threads used to send samples through.

### LINKED ISSUE #266: Sound crackles when using Sink
When I play a sound using `Sink`, there is a slight crackling to be heard in the output, like very short interruptions in the sound. This crackling is not present if I play the same source directly via `play_raw`. In [src/sink.rs:90](https://github.com/RustAudio/rodio/blob/master/src/sink.rs#L90), if instead of submitting the source to `queue_tx`, I call `play_raw` directly, the crackling goes away (but of course that makes appending no longer work). So it seems to be related to the use of queues in `Sink`, but I don't know enough to investigate further.

I'm running Linux Mint 19.3.

### LINKED ISSUE #318: `Done` source does not implement size_hint

### LINKED ISSUE #319: SourcesQueueOutput::current_frame_length relies on upper bound of iterator size instead of lower bound.
SourcesQueueOutput::current_frame_length should return the shortest guaranteed frame, however it is conceivable that a source has an arbitrarily high upper bound but it is in fact much shorter.

It should instead rely on the lower bound of the iterator size.

HalfVoxel:
Fixes #318.
Fixes #319.
Fixes #266.

Combined, these two fixes also fix crackling audio on my computer. The reason for the crackling audio was that Done did not implement size_hint so it would always fall back to a frame size of 512. Somewhere down the pipeline this causing crackling every 512 samples I suspect (which might be worth fixing too).

The current_frame_length fix is necessary because in my case I was playing from a sound file and it returns a size hint on the format `(very_large_number, Some(very_large_number))` and the current_frame_length method would just ignore that and fall back to a frame length of 512 samples anyway.

This might also fix #225 (I am running on linux, so I cannot test that).

### DIFF
--- a/src/queue.rs
+++ b/src/queue.rs
@@ -132,10 +132,11 @@ where
         }
 
         // Try the size hint.
-        if let Some(val) = self.current.size_hint().1 {
-            if val < THRESHOLD && val != 0 {
-                return Some(val);
-            }
+        let (lower_bound, _) = self.current.size_hint();
+        // The iterator default implementation just returns 0.
+        // That's a problematic value, so skip it.
+        if lower_bound > 0 {
+            return Some(lower_bound);
         }
 
         // Otherwise we use the constant value.
--- a/src/source/done.rs
+++ b/src/source/done.rs
@@ -57,6 +57,10 @@ where
         }
         next
     }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        self.input.size_hint()
+    }
 }
 
 impl<I> Source for Done<I>


est31 (changes requested):
There should still be an upper bound to the length. It influences buffer sizes for example. If THRESHOLD is too small and  that causes issues, one can define another larger constant by which to clamp the returned value, e.g. 4096.

HalfVoxel:
Is this method really the right place to clamp it though? If a method further down the pipeline needs to break things up into chunks for more efficient buffering it should be free to do that anyway, right?

The frame_len is only supposed to be an upper bound on the allowed buffer if I understand it correctly.

HalfVoxel:
In a test I changed it to always return Some(4096) and that does get rid of the crackling as well.

est31:
Hmm yeah I think I misread `src/source/buffered.rs`. It's fine.

est31 (approved):
Thanks!

HalfVoxel:
Hmmm, actually UniformSourceIterator::bootstrap does use the frame size without any limits. Might want to have some limit there.

est31:
@HalfVoxel PRs welcome!

HalfVoxel:
I would gladly open one, but I'm not sure what the best limit value is? I haven't got enough insight into the library to have a feel for the possible consequences.

est31:
buffered.rs uses 32768. I guess that value is okay.

Xaeroxe:
giphy

Issue #316, opened 2020-09-17
SampleRateConverter converts frames at wrong rate


TonalidadeHidrica:
When I played a 48000Hz mono-channel ogg file converted by `UniformSourceIterator`, in which the sample rate is converted in `SampleRateConverter`, the conversion rate seems to be slightly wrong.  

First I generated a 48000Hz mono-channel ogg file which includes a metronome sound at 120 BPM for 400 beats with Audacity.  Then I played it by this library, output the sound from the speaker, where the device output sample rate is set to 44100Hz, and record the sound from the speaker with the microphone via Audacity again.  Then, the duration between the 0th and 200th beat was exactly 100.050 seconds.  When I synthesize the wave derived from the ogg file with another tone constantly made 22050 samples, the 0th and 200th beat has the exact elapsed time of 100.000 seconds.  The gap between the two source was becoming proportionally larger.

When I did the same experiment with 44100- and 22050-Hz ogg file, this gap didn't happen.   We can infer that there is something wrong with conversion of sample rate.  I have not made an in-depth inspection on the code though.

Since I found several other issues in `SampleRateConverter` other than this, I didn't fix the converter.  That's why I'm just submitting an issue.  FWIW, [this](https://github.com/TonalidadeHidrica/taiko-untitled/blob/feature/cpal-and-rodio/src/rodio/sample_converter.rs) is an alternative implementation I wrote.  In this converter, I used floating-point operations to make a linear interpolation, so it may not be as concise as the converter in this library.  Nevertheless my code does not have the issue of conversion error.

Rua:
This might have the same cause as #303. What happens if you use `rustup override` to force it to use Rust 1.44.1?

TonalidadeHidrica:
In short, the issue exists as well in 1.44.1.

I created a minimal example to reproduce the issue.  You can see the code in [my fork](https://github.com/TonalidadeHidrica/rodio/blob/sandbox/issue-316/examples/issue_316.rs).  I wanted to attach the ogg files I used, but OGG files are not allowed for GitHub Issue attachments (how can I?).


The following is the results:

```shell
[N] ⋊> ~/d/g/c/g/T/rodio on sandbox/issue-316  rustup run 1.46.0 cargo run --example issue_316 ~/Documents/Metronome2/M2_44100.ogg                                                                                                  17:19:03
    Finished dev [unoptimized + debuginfo] target(s) in 0.05s
[...]

Issue #303, opened 2020-07-19
Bad sound quality when upgrading to Rust 1.45


Rua:
I updated my toolchain to use Rust 1.45, this has broken the audio playback using Rodio 0.11.0. The sound comes out as somewhat muffled, like it was low-pass filtered. I can restore the correct audio by downgrading Rust back to 1.44.1, so clearly there is something going on with changes made in the compiler and how they interact with Rodio. Perhaps the changes in float-to-integer casts are to blame.

I've attached a sound recording for comparison. First, two sound effects are heard from Rust 1.45, then the same two effects in Rust 1.44.1. (Yes, these are old Doom sound effects, if anyone is wondering :) )

[brokensound.ogg.zip](https://github.com/RustAudio/rodio/files/4944570/brokensound.ogg.zip)

To play this audio, I use a custom `Source` that reads its samples from an `Arc<[i16]>`, so that I can keep the buffer around after playing it. This is then wrapped in the `ChannelVolume`, `Done` and `SamplesConverter` sources, sent to another thread (to avoid the COM initialisation bug on Windows), where it is then played via `rodio::play_raw`.

est31:
Hi! interesting report. You could try to narrow it down by bisecting the rustc: https://github.com/rust-lang/cargo-bisect-rustc/blob/master/TUTORIAL.md

Then we'd know what to look for.

Rua:
I tried it out, but unfortunately it only builds my project, it doesn't actually run it. It needs to run in order for me to hear whether the sound is good or bad.

Rua:
Here's the result I got from it:

searched nightlies: from nightly-2020-05-01 to nightly-2020-07-20
regressed nightly: nightly-2020-05-20
searched commits: from https://github.com/rust-lang/rust/commit/d8878868c8d7ef3779e7243953fc050cbb0e0565 to https://github.com/rust-lang/rust/commit/3a7dfda40a3e798bf086bd58cc7e5e09deb808b5
regressed commit: https://github.com/rust-lang/rust/commit/672b272077561ca7b5027a8aff9ea2957c7d4c21

Rua:
I just checked the output of the `Source` that's being fed to `rodio::play_raw` in both versions. It gives the exact same stream of `f32`s. So it's definitely not a problem in my side of the code, something is going wrong after Rodio gets the `Source`.

est31:
It's probably due to some UB caused by the tiny vecs change. Interesting...

Rua:
[...]

PR #324, opened 2020-10-04
SampleRateConverter: Fix reliance on Vec internals (see #303)

nbraud:
## Bugfixes

- [x] Removed use of `Vec::capacity`, instead carry the number of channels as an explicit member of the `struct SampleRateConverter`.
- [x] Fixed an underflow (leading to `panic!`) in `<u16 as Sample>::lerp`, discovered when switching to property-based tests.

## Testsuite improvements

- [x] Added a reproducing testcase for #316 (marked `#[ignore]`, pending a fix)
- [x] Switched to property-based tests (using `quickcheck`)
- [x] Generalised the tests over their input data, number of channels, and sample rates.
- [x] Generalised the sample rate multiplication and division tests over the (integer) conversion ratio.
- [x] Removed redundant tests.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,3 +23,6 @@ flac = ["claxon"]
 vorbis = ["lewton"]
 wav = ["hound"]
 mp3 = ["minimp3"]
+
+[dev-dependencies]
+quickcheck = "0.9.2"
--- a/src/conversions/sample.rs
+++ b/src/conversions/sample.rs
@@ -84,7 +84,11 @@ pub trait Sample: CpalSample {
 impl Sample for u16 {
     #[inline]
     fn lerp(first: u16, second: u16, numerator: u32, denominator: u32) -> u16 {
-        (first as u32 + (second as u32 - first as u32) * numerator / denominator) as u16
+        let a = first as i32;
+        let b = second as i32;
+        let n = numerator as i32;
+        let d = denominator as i32;
+        (a + (b - a) * n / d) as u16
     }
 
     #[inline]
--- a/src/conversions/sample_rate.rs
+++ b/src/conversions/sample_rate.rs
@@ -274,27 +274,15 @@ mod test {
             assert_eq!(output, []);
         }
 
-        fn identity(from: u32, n: u16) -> () {
+        fn identity(from: u32, n: u16, input: Vec<u16>) -> () {
             let from = if from == 0 { return; } else { SampleRate(from) };
             if n == 0 { return; }
 
-            let input = vec![2u16, 16, 4, 18, 6, 20, 8, 22];
             let output =
-                SampleRateConverter::new(input.into_iter(), from, from, n);
+                SampleRateConverter::new(input.clone().into_iter(), from, from, n);
 
             let output = output.collect::<Vec<_>>();
-            assert_eq!(output, [2u16, 16, 4, 18, 6, 20, 8, 22]);
-        }
-
-        fn identity_2channels_misalign(from: u32) -> () {
-            let from = if from == 0 { return; } else { SampleRate(from) };
-
-            let input = vec![2u16, 16, 4, 18, 6];
-            let output =
-                SampleRateConverter::new(input.into_iter(), from, from, 2);
-
-            let output = output.collect::<Vec<_>>();
-            assert_eq!(output, [2u16, 16, 4, 18, 6]);
+            assert_eq!(input, output);
         }
 
         fn half_sample_rate(to: u32) -> () {
--- a/src/conversions/sample_rate.rs
+++ b/src/conversions/sample_rate.rs
@@ -285,28 +285,46 @@ mod test {
             assert_eq!(input, output);
         }
 
-        fn half_sample_rate(to: u32) -> () {
+        fn half_sample_rate(to: u32, input: Vec<u16>) -> () {
             let to = if to == 0 { return; } else { SampleRate(to) };
             let from = multiply_rate(to, 2);
+            let n = 2u16;
 
-            let input = vec![1u16, 16, 2, 17, 3, 18, 4, 19, 5, 20, 6, 21];
+            // Truncate the input, so it contains an integer number of frames.
+            let input = {
+                let ns = n as usize;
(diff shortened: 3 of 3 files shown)


TonalidadeHidrica:
> @TonalidadeHidrica Yeah, I'm pretty sure it ought to be simplified a lot; what it's actually doing is just finding the 2 input samples in-between which the (current) output sample falls, and interpolating linearly between the two.
> What I was planning to do was first refactor and simplify the current implementation, fix this issue, and then look at better resampling algorithms to implement (windowed `sinc` interpolation might be a big win already, but TBH it might make more sense to go straight for band-limited interpolation)

_Originally posted by @nbraud in https://github.com/RustAudio/rodio/issues/316#issuecomment-703259661_

I thought discussion about new implementation of Sample Rate Converter should be done in here, so I quoted here.

@nbraud I think that linear of sinc interpolation seems good.  By the way, why don't we reuse an existing algorithm instead of reinventing wheels?  With a little googling I found many open-source libraries available, some of which listed [here](https://ccrma.stanford.edu/~jos/resample/Free_Resampling_Software.html).

est31:
I won't accept PRs that replace existing Rust code with a dependency on a C library. If the specific component was rewritten in Rust I might do it, but licenses need to be compatible.

est31 (approved):
Thanks!

nbraud:
> I think that linear of sinc interpolation seems good. By the way, why don't we reuse an existing algorithm instead of reinventing wheels? With a little googling I found many open-source libraries available, some of which listed [here](https://ccrma.stanford.edu/~jos/resample/Free_Resampling_Software.html).

@TonalidadeHidrica I didn't want to suggest a replacement with a C library, both for portability and ergonomics reasons (`cargo` can build or cross-build any pure-Rust dependency easily, but dealing with C dependencies can be a pain) and because it seems counter-productive to throw out all guarantees of memory safety, and it looks like @est31 agrees:

> I won't accept PRs that replace existing Rust code with a dependency on a C library. If the specific component was rewritten in Rust I might do it, but licenses need to be compatible.

... however, I looked earlier at published crates that implement audio interpolation, and found [`rubato`], which seems to be the most reasonable implementation, is MIT-licensed, and pure Rust, though it has some potential `unsafe`ty (not directly, but through its dependencies, mostly `rustfft`) : 

```
$ cargo geiger
[...]
Functions  Expressions  Impls  Traits  Methods  Dependency

0/0        0/0          0/0    0/0     0/0      ?  rubato 0.4.3
1/1        16/28        0/0    0/0     0/2      !  ├── log 0.4.11
0/0        0/0          0/0    0/0     0/0      ?  │   └── cfg-if 0.1.10
0/0        0/0          0/0    0/0     0/0      ?  ├── num 0.2.1
0/0        2/2          0/0    0/0     0/0      !  │   ├── num-bigint 0.2.6
0/0        0/0          0/0    0/0     0/0      ?  │   │   ├── num-integer 0.1.43
0/0        4/10         0/0    0/0     0/0      !  │   │   │   └── num-traits 0.2.12
0/0        4/10         0/0    0/0     0/0      !  │   │   └── num-traits 0.2.12
0/0        0/0          0/0    0/0     0/0      ?  │   ├── num-complex 0.2.4
0/0        4/10         0/0    0/0     0/0      !  │   │   └── num-traits 0.2.12
0/0        0/0          0/0    0/0     0/0      ?  │   ├── num-integer 0.1.43
0/0        0/0          0/0    0/0     0/0      ?  │   ├── num-iter 0.1.41
0/0        0/0          0/0    0/0     0/0      ?  │   │   ├── num-integer 0.1.43
0/0        4/10         0/0    0/0     0/0      !  │   │   └── num-traits 0.2.12
0/0        0/0          0/0    0/0     0/0      ?  │   ├── num-rational 0.2.4
0/0        2/2          0/0    0/0     0/0      !  │   │   ├── num-bigint 0.2.6
0/0        0/0          0/0    0/0     0/0      ?  │   │   ├── num-integer 0.1.43
0/0        4/10         0/0    0/0     0/0      !  │   │   └── num-traits 0.2.12
0/0        4/10         0/0    0/0     0/0      !  │   └── num-traits 0.2.12
0/0        0/0          0/0    0/0     0/0      ?  └── realfft 0.2.1
3/3        1362/1362    0/0    0/0     22/22    !      └── rustfft 3.0.1
0/0        0/0          0/0    0/0     0/0      ?          ├── num-complex 0.2.4
0/0        0/0          0/0    0/0     0/0      ?          ├── num-integer 0.1.43
0/0        4/10         0/0    0/0     0/0      !          ├── num-traits 0.2.12
0/0        0/0          0/0    0/0     0/0      ?          ├── strength_reduce 0.2.3
1/1        27/27        0/0    0/0     0/0      !          └── transpose 0.1.0

5/5        1411/1429    0/0    0/0     22/24  
```

@est31 Would a PR that replaces the `SampleRateConverter` with [`rubato`] be acceptable?
I might also look into submitting patches to `rustfft` to reduce the reliance on `unsafe` code.

[`rubato`]: https://crates.io/crates/rubato

est31:
Yes, rubato sounds good, same goes for the pure-rust variant of speexdsp.

I'm not sure if rubato really needs to use all the num dependencies, and it would be great if it updated to 0.3.0. Will try to file a PR to them, but the issues are no blockers to adoption.

est31:
The speexdsp crate to look for would probably be [speexdsp-resampler](https://docs.rs/speexdsp-resampler/). Zero dependencies, I like that. What about using that one?

nbraud:
> The speexdsp crate to look for would probably be [speexdsp-resampler](https://docs.rs/speexdsp-resampler/). Zero dependencies, I like that. What about using that one?

If that's the one you'd insist on, sure, but I had looked at it and disregarded it because:
- there's no documentation and not much of a testsuite;
- the API seems very C-ish and error-prone: for instance, all quantities (latency, quality, sample rate, etc.) are expressed as `usize` (and we are left to guess the actual unit, since there's not documentation), instead of using `Duration` and some wrapper types;
- there was a single release ever, in February, and development seems to have stalled in June.

PR #329, opened 2020-10-08
Implement Display and Error traits for PlayError

Paoda:
Noticed that `rodio::StreamError` had implemented `Error`, however, `rodio::PlayError` had not.

The implementations of `Display` and `Error`for `PlayError` are identical to the ones for `StreamError`.


est31 (approved):
Thanks!

PR #335, opened 2020-10-26
Fix for overflow on 32bit

eloc3147:
Just encountered a multiplication overflow while trying to play a wav file on a Raspberry Pi 4B. This fixed it for me.


est31 (approved):
Thanks!

PR #336, opened 2020-11-03
Update cpal with Android support

enfipy:
Bumps `cpal` version to 0.13.0 and `rodio` 0.13.0 as Android support was added but `rodio` interface hasn't changed.

@est31

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# Version 0.12.1 (2020-11-03)
+# Version 0.13.0 (2020-11-03)
 
 - Update `cpal` to [0.13](https://github.com/RustAudio/cpal/blob/master/CHANGELOG.md#version-0130-2020-10-28).
 - Add Android support.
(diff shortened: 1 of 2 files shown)


est31 (changes requested):
Some of the cpal error types are reused, so a semver incompatible update of cpal is a breaking change. Can you change the rodio release number to 0.13?

est31 (changes requested):
Changelog needs changes too

enfipy:
Oh, so sorry. Didn't notice that 😳

est31 (approved):
Thanks!

enfipy:
@est31 Sorry for bothering you but It looks like rodio 0.13 disappeared from crates.io 😳
I'm sure I saw it an hour ago on crates but now - there's only 0.12!

est31:
I doubt that would be possible as the publishing is broken entirely.

I need to update the crates.io token, one moment.

enfipy:
Thank you a lot! Maybe it's just my eyes were lagging xD

est31:
The build is here, currently queued: https://travis-ci.org/github/RustAudio/rodio/builds/741138623

Once it finishes, the crate should be published.

Issue #313, opened 2020-08-24
Support for webassembly builds?


robert-w-gries:
Is there interest in supporting webassembly in `rodio`? I found the [`amethyst` fork](https://github.com/amethyst/rodio) that hasn't been maintained in several months and was interested in getting the `wasm` branch back into a buildable state.

One issue I found is that the `minimp3` project does not currently support wasm due to the [`slice_deque` dependency](https://github.com/gnzlbg/slice_deque/issues/86).

est31:
Sure why not? cpal has wasm support too.

glalonde:
This already works if you remove the mp3 feature:
`rodio = { git = "https://github.com/RustAudio/rodio" , default-features = false, features = ["vorbis", "flac", "wav"] }`

It also works if we use puremp3 as the mp3 decoder. I've tried this, and it seems okay but there are audible artifacts

robert-w-gries:
I had to modify the rodio `Cargo.toml` and add the following feature before getting webassembly to work: 

```
[features]
...
wasm-bindgen = ["cpal/wasm-bindgen"]
```

I tried the following in my project (using raw audio instead of decoders) and was able to get slightly mangled sound:

```rust
    let v = raw_data.to_vec();
    let (_stream, stream_handle) = rodio::OutputStream::try_default().unwrap();
    let buffer = SamplesBuffer::new(1,44100, v);
    let sink = Sink::try_new(&stream_handle).unwrap();
    sink.append(buffer);
    sink.sleep_until_end();
```

However, I see the following error: `panicked at 'can't block with web assembly', src/libstd/sys/wasm/condvar.rs:21:9`

According to [this comment](https://github.com/rustwasm/wasm-bindgen/issues/614#issuecomment-411781711) it's possible to use a `Promise` instead of sleeping. I'll try that and update with my results

robert-w-gries:
I'm running into the same problem as #310. I am running rodio webassembly on the main thread, so I can't call `sleep_until_end()`. I can't use a web worker to run the webassembly code because web workers don't have access to the necessary audio apis.

What's strange to me is that `cpal` works fine without blocking. Is there a way to use `rodio` without the blocking behavior, or do I need to give up and just use `cpal` for playback?

glalonde:
It works if you use a OutputStream: https://github.com/glalonde/wasm_audio/blob/f01b022e5e9a0525f37f556367154aedf19c7dbc/src/lib.rs#L55

[...]

PR #342, opened 2020-12-09
Add feature flag to enable wasm-bindgen

SnoozeTime:
Rodio can work on wasm32 target if the cpal/wasm-bindgen feature is enabled.

This is described in more details here #313.

Only mp3 needs to be disabled for it to work.


est31 (approved):
Thanks!

PR #350, opened 2021-02-07
Better error handling for stream creation

jiminycrick:
I am a contributor to Veloren (an open source game) and we have had players with different devices reaching the expects in stream.rs. This MR removes all the expects and replaces it with some error handling. Instead of panicking, this will allow the program to decide what to do with the error.
You can ignore the changes in src/decoder/mod.rs, that was just a result of cargo fmt.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -97,8 +97,8 @@
 //!
 #![cfg_attr(test, deny(missing_docs))]
 pub use cpal::{
-    traits::DeviceTrait, Device, Devices, DevicesError, InputDevices, OutputDevices,
-    SupportedStreamConfig, self,
+    self, traits::DeviceTrait, Device, Devices, DevicesError, InputDevices, OutputDevices,
+    SupportedStreamConfig,
 };
 
 mod conversions;
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -105,7 +105,28 @@ impl error::Error for PlayError {
 #[derive(Debug)]
 pub enum StreamError {
     PlayStreamError(cpal::PlayStreamError),
+    DefaultStreamConfigError(cpal::DefaultStreamConfigError),
+    BuildStreamError(cpal::BuildStreamError),
+    SupportedStreamConfigsError(cpal::SupportedStreamConfigsError),
     NoDevice,
+}
+
+impl From<cpal::DefaultStreamConfigError> for StreamError {
+    fn from(err: cpal::DefaultStreamConfigError) -> Self {
+        Self::DefaultStreamConfigError(err)
+    }
+}
+
+impl From<cpal::SupportedStreamConfigsError> for StreamError {
+    fn from(err: cpal::SupportedStreamConfigsError) -> Self {
+        Self::SupportedStreamConfigsError(err)
+    }
+}
+
+impl From<cpal::BuildStreamError> for StreamError {
+    fn from(err: cpal::BuildStreamError) -> Self {
+        Self::BuildStreamError(err)
+    }
 }
 
 impl From<cpal::PlayStreamError> for StreamError {
@@ -118,6 +139,9 @@ impl fmt::Display for StreamError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
             Self::PlayStreamError(e) => e.fmt(f),
+            Self::BuildStreamError(e) => e.fmt(f),
+            Self::DefaultStreamConfigError(e) => e.fmt(f),
+            Self::SupportedStreamConfigsError(e) => e.fmt(f),
             Self::NoDevice => write!(f, "NoDevice"),
         }
     }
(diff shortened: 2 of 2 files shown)


LiquidityC:
This PR would also help me a lot. Have an application that offers audio support but it's not needed. Some users run on systems that don't have audio at all (VPS, WSL) which is fine, I have code to deal with the "no audio" situation. But theses users now get an error from inside rodio which I can't capture and deal with and then everything crashes.

est31 (changes requested):
This is good for the most part, but could you separate the rustfmt changes from the actual behaviour changes?

Also I wonder if there are other places in the code base where expect is used where it should not be. This PR constitutes a breaking change and would require a new semver breaking release. I don't want to be in the same situation if you discover another place in a few weeks :).

est31 (approved):
Thanks! This PR looks great now. I wonder about the other places that unwrap/expect is used.  Before this PR gets merged, it'd be great to have a survey for them to check whether any situation would benefit from passing on the error instead of unwrapping.

jiminycrick:
There are no more `expect()`s in the codebase. Two unwraps that I think could be replaced with `?` in `src/source/blt.rs` and `src/source/uniform.rs`. I found at least one other place where an unwrap could be replaced with an unwrap_or but the documentation comment says that it is supposed to panic there.

jiminycrick:
I should also note that in the case of the `stream.rs` code all the functions that were panicking return Results so changing the internal error handling shouldn't have any effect on the API.

est31:
> all the functions that were panicking return Results so changing the internal error handling shouldn't have any effect on the API.

Huh thought the CpalDeviceExt trait were public but seems no. The `StreamError` enum is public though and not marked as non exhaustive so extending it is a breaking change.

Imberflur:
What is the status on this?

est31:
I think this PR can go in, if rebased. Then we can release 0.14.0.

jiminycrick:
> 
> 
> I think this PR can go in, if rebased. Then we can release 0.14.0.

Rebased and ready to go :D

alexheretic:
This looks good to me :+1: , I was in the area and pretty much implemented the same changes. One thing I did differently was add _thiserror_ dependency and flatten the cpal errors into a renamed `OutputStreamError`.

I'll leave it here as an idea, feel free to discard. Also, in any case, we may as well use _thiserror_ to compact the code as cpal already pulls it in.

```rust
/// Errors that can occur working with output streams.
///
/// Caused by [`cpal::PlayStreamError`], [`cpal::BuildStreamError`],
/// [`cpal::DefaultStreamConfigError`].
#[derive(Debug, thiserror::Error)]
pub enum OutputStreamError {
    /// The device associated with the stream is no longer available.
    #[error("the device associated with the stream is no longer available")]
    DeviceNotAvailable,
    /// The specified stream configuration is not supported.
    #[error("The requested stream configuration is not supported by the device.")]
    StreamConfigNotSupported,
    /// Returned if e.g. the default input format was requested on an output-only audio device.
    #[error("The requested stream type is not supported by the device.")]
    StreamTypeNotSupported,
    /// We called something the C-Layer did not understand
    ///
    /// On ALSA device functions called with a feature they do not support will yield this. E.g.
    /// Trying to use capture capabilities on an output only format yields this.
    #[error("The requested device does not support this capability (invalid argument)")]
    InvalidArgument,
    /// Occurs if adding a new Stream ID would cause an integer overflow.
    #[error("Adding a new stream ID would cause an overflow")]
    StreamIdOverflow,
    /// See the `BackendSpecificError` docs for more information about this error variant.
    #[error(transparent)]
    BackendSpecific(#[from] cpal::BackendSpecificError),
}

impl From<cpal::PlayStreamError> for OutputStreamError {
    fn from(err: cpal::PlayStreamError) -> Self {
        use cpal::PlayStreamError::*;
        match err {
            DeviceNotAvailable => Self::DeviceNotAvailable,
            BackendSpecific { err } => err.into(),
        }
    }
}

impl From<cpal::BuildStreamError> for OutputStreamError {
    fn from(err: cpal::BuildStreamError) -> Self {
        use cpal::BuildStreamError::*;
        match err {
            DeviceNotAvailable => Self::DeviceNotAvailable,
            StreamConfigNotSupported => Self::StreamConfigNotSupported,
            InvalidArgument => Self::InvalidArgument,
            StreamIdOverflow => Self::StreamIdOverflow,
            BackendSpecific { err } => err.into(),
        }
    }
}

impl From<cpal::DefaultStreamConfigError> for OutputStreamError {
    fn from(err: cpal::DefaultStreamConfigError) -> Self {
        use cpal::DefaultStreamConfigError::*;
        match err {
            DeviceNotAvailable => Self::DeviceNotAvailable,
            StreamTypeNotSupported => Self::StreamTypeNotSupported,
            BackendSpecific { err } => err.into(),
        }
    }
}
```

est31:
Merged. Sorry for the delay!

Regarding your suggestion @alexheretic, personally I'd prefer to avoid adding dependencies like `thiserror`.

PR #352, opened 2021-02-13
Adds 'cpal' to re-exports

LiquidityC:
If a user wants some finer access to devices and hosts this will allow
that without having to add a separate dependency to cpal and risk having
a cersion conflict between the cpal dependency and the version of cpal
within rodio.


CleanCut (approved):

est31 (approved):
Alright there were good points made by @Ten0 in #363. Let's get this in!

PR #356, opened 2021-03-06
Remove unnecessary 'static bound and cleanup

CryZe:
The decoder had an unnecessary 'static bound. Additionally I noticed that there's a lot of (clippy) warnings, so I cleaned up the code a bit as well. There's a few warnings left, but a few of those require breaking changes.

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -10,9 +10,10 @@
 //! ```
 //!
 
-use std::{time::Duration, vec::IntoIter as VecIntoIter};
+use std::time::Duration;
+use std::vec::IntoIter as VecIntoIter;
 
-use crate::{source::Source, Sample};
+use crate::{Sample, Source};
 
 /// A buffer of samples treated as a source.
 pub struct SamplesBuffer<S> {
@@ -104,7 +105,8 @@ where
 
 #[cfg(test)]
 mod tests {
-    use crate::{buffer::SamplesBuffer, source::Source};
+    use crate::buffer::SamplesBuffer;
+    use crate::source::Source;
 
     #[test]
     fn basic() {
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -10,11 +10,9 @@
 //! ```
 //!
 
-use std::time::Duration;
-use std::vec::IntoIter as VecIntoIter;
+use std::{time::Duration, vec::IntoIter as VecIntoIter};
 
-use crate::source::Source;
-use crate::Sample;
+use crate::{source::Source, Sample};
 
 /// A buffer of samples treated as a source.
 pub struct SamplesBuffer<S> {
@@ -32,9 +30,9 @@ where
     ///
     /// # Panic
     ///
-    /// - Panicks if the number of channels is zero.
-    /// - Panicks if the samples rate is zero.
-    /// - Panicks if the length of the buffer is larger than approximatively 16 billion elements.
+    /// - Panics if the number of channels is zero.
+    /// - Panics if the samples rate is zero.
+    /// - Panics if the length of the buffer is larger than approximately 16 billion elements.
     ///   This is because the calculation of the duration would overflow.
     ///
     pub fn new<D>(channels: u16, sample_rate: u32, data: D) -> SamplesBuffer<S>
@@ -55,9 +53,9 @@ where
 
         SamplesBuffer {
             data: data.into_iter(),
-            channels: channels,
-            sample_rate: sample_rate,
-            duration: duration,
+            channels,
+            sample_rate,
+            duration,
         }
     }
 }
@@ -106,8 +104,7 @@ where
 
 #[cfg(test)]
 mod tests {
-    use crate::buffer::SamplesBuffer;
-    use crate::source::Source;
+    use crate::{buffer::SamplesBuffer, source::Source};
 
     #[test]
     fn basic() {
--- a/src/conversions/sample_rate.rs
+++ b/src/conversions/sample_rate.rs
@@ -1,5 +1,4 @@
 use crate::conversions::Sample;
-use cpal;
 
 use std::mem;
 
@@ -39,7 +38,7 @@ where
     ///
     /// # Panic
     ///
-    /// Panicks if `from` or `to` are equal to 0.
+    /// Panics if `from` or `to` are equal to 0.
     ///
     #[inline]
     pub fn new(
@@ -85,7 +84,7 @@ where
         };
 
         SampleRateConverter {
-            input: input,
+            input,
             from: from / gcd,
             to: to / gcd,
             channels: num_channels,
@@ -133,7 +132,7 @@ where
         }
 
         // Short circuit if there are some samples waiting.
-        if self.output_buffer.len() >= 1 {
+        if !self.output_buffer.is_empty() {
             return Some(self.output_buffer.remove(0));
         }
 
@@ -174,7 +173,7 @@ where
             .zip(self.next_frame.iter())
             .enumerate()
         {
-            let sample = Sample::lerp(cur.clone(), next.clone(), numerator, self.to);
+            let sample = Sample::lerp(*cur, *next, numerator, self.to);
 
             if off == 0 {
                 result = Some(sample);
@@ -192,7 +191,7 @@ where
             debug_assert!(self.next_frame.is_empty());
 
             // draining `self.current_frame`
-            if self.current_frame.len() >= 1 {
+            if !self.current_frame.is_empty() {
                 let r = Some(self.current_frame.remove(0));
                 mem::swap(&mut self.output_buffer, &mut self.current_frame);
                 self.current_frame.clear();
--- a/src/decoder/flac.rs
+++ b/src/decoder/flac.rs
@@ -1,6 +1,9 @@
-use std::io::{Read, Seek, SeekFrom};
-use std::mem;
-use std::time::Duration;
+use std::{
+    cmp::Ordering,
+    io::{Read, Seek, SeekFrom},
+    mem,
+    time::Duration,
+};
 
 use crate::Source;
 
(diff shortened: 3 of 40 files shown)


src/buffer.rs
```diff
@@ -106,8 +104,7 @@ where
 
 #[cfg(test)]
 mod tests {
-    use crate::buffer::SamplesBuffer;
-    use crate::source::Source;
+    use crate::{buffer::SamplesBuffer, source::Source};
```
est31, line 107: I don't like this change, can you keep it?

src/conversions/sample_rate.rs
```diff
@@ -133,7 +132,7 @@ where
         }
 
         // Short circuit if there are some samples waiting.
-        if self.output_buffer.len() >= 1 {
+        if !self.output_buffer.is_empty() {
```
est31, line 135: I don't like this change, can you keep it?

src/conversions/sample_rate.rs
```diff
@@ -192,7 +191,7 @@ where
             debug_assert!(self.next_frame.is_empty());
 
             // draining `self.current_frame`
-            if self.current_frame.len() >= 1 {
+            if !self.current_frame.is_empty() {
```
est31, line 194: I don't like this change, can you keep it?

src/decoder/flac.rs
```diff
@@ -1,6 +1,9 @@
-use std::io::{Read, Seek, SeekFrom};
-use std::mem;
-use std::time::Duration;
+use std::{
```
est31, line 1: I don't like this change, can you keep it?

src/decoder/mod.rs
```diff
@@ -1,12 +1,14 @@
 //! Decodes samples from an audio file.
 
-use std::error::Error;
-use std::fmt;
-use std::io::{Read, Seek};
 #[allow(unused_imports)]
 use std::io::SeekFrom;
-use std::mem;
-use std::time::Duration;
+use std::{
```
est31, line 5: I don't like this change, can you keep it?

src/decoder/mp3.rs
```diff
@@ -1,5 +1,7 @@
-use std::io::{Read, Seek};
-use std::time::Duration;
+use std::{
```
est31, line 1: I don't like this change, can you keep it?

src/decoder/vorbis.rs
```diff
@@ -1,6 +1,8 @@
-use std::io::{Read, Seek, SeekFrom};
-use std::time::Duration;
-use std::vec;
+use std::{
```
est31, line 1: I don't like this change, can you keep it?

src/decoder/wav.rs
```diff
@@ -1,5 +1,7 @@
-use std::io::{Read, Seek, SeekFrom};
-use std::time::Duration;
+use std::{
```
est31, line 1: I don't like this change, can you keep it?

src/lib.rs
```diff
@@ -106,9 +104,11 @@ pub mod queue;
 pub mod source;
 pub mod static_buffer;
 
-pub use crate::conversions::Sample;
-pub use crate::decoder::Decoder;
-pub use crate::sink::Sink;
-pub use crate::source::Source;
-pub use crate::spatial_sink::SpatialSink;
-pub use crate::stream::{OutputStream, OutputStreamHandle, PlayError, StreamError};
+pub use crate::{
```
est31, line 107: I don't like this change, can you keep it?

src/sink.rs
```diff
@@ -1,15 +1,14 @@
...
-use std::sync::mpsc::Receiver;
-use std::sync::Arc;
-use std::sync::Mutex;
-use std::time::Duration;
-
-use crate::queue;
-use crate::source::Done;
-use crate::Sample;
-use crate::Source;
+use std::{
```
est31, line 2: I don't like this change, can you keep it? Only the `std::sync::atomic::Ordering` should be put into the group of `{AtomicBool, AtomicUsize, Ordering}`.

src/sink.rs
```diff
@@ -178,9 +177,7 @@ impl Drop for Sink {
 
 #[cfg(test)]
 mod tests {
-    use crate::buffer::SamplesBuffer;
-    use crate::sink::Sink;
-    use crate::source::Source;
+    use crate::{buffer::SamplesBuffer, sink::Sink, source::Source};
```
est31, line 180: I don't like this change, can you keep it?

src/source/amplify.rs
```diff
@@ -1,18 +1,14 @@
 use std::time::Duration;
 
-use crate::Sample;
-use crate::Source;
+use crate::{Sample, Source};
```
est31, line 3: I don't like this change, can you keep it?

src/source/blt.rs
```diff
@@ -1,5 +1,4 @@
-use std::f32::consts::PI;
-use std::time::Duration;
+use std::{f32::consts::PI, time::Duration};
```
est31, line 1: I don't like this change, can you keep it?

src/source/buffered.rs
```diff
@@ -1,9 +1,10 @@
-use std::cmp;
-use std::sync::{Arc, Mutex};
-use std::time::Duration;
+use std::{
```
est31, line 1: I don't like this change, can you keep it?

src/source/crossfade.rs
```diff
@@ -1,8 +1,9 @@
 use std::time::Duration;
 
-use crate::source::{FadeIn, Mix, TakeDuration};
-use crate::Sample;
-use crate::Source;
+use crate::{
```
est31, line 3: I don't like this change, can you keep it?

src/source/empty.rs
```diff
@@ -1,12 +1,17 @@
-use crate::Sample;
-use crate::Source;
-use std::marker::PhantomData;
-use std::time::Duration;
+use crate::{Sample, Source};
+use std::{marker::PhantomData, time::Duration};
```
est31, line 2: I don't like this change, can you keep it? The sample, Source change is fine, only really about the second line here.

src/source/from_iter.rs
```diff
@@ -140,9 +139,10 @@ where
 
 #[cfg(test)]
 mod tests {
-    use crate::buffer::SamplesBuffer;
-    use crate::source::from_iter;
-    use crate::source::Source;
+    use crate::{
+        buffer::SamplesBuffer,
+        source::{from_iter, Source},
```
est31, line 144: I don't like this change, can you keep it?

src/source/periodic.rs
```diff
@@ -122,10 +121,8 @@ where
 
 #[cfg(test)]
 mod tests {
-    use crate::buffer::SamplesBuffer;
-    use crate::source::Source;
-    use std::cell::RefCell;
-    use std::time::Duration;
+    use crate::{buffer::SamplesBuffer, source::Source};
```
est31, line 124: I don't like this change, can you keep it?

src/source/sine.rs
```diff
@@ -1,5 +1,5 @@
 use crate::Source;
-use std::time::Duration;
+use std::{f32::consts::TAU, time::Duration};
```
est31, line 2: Can you put it onto two separate lines? Also I'd prefer usage of PI not TAU.

src/source/skip.rs
```diff
@@ -147,15 +147,14 @@ where
...
+                .unwrap_or_else(|| Duration::from_secs(0))
         })
     }
 }
 
 #[cfg(test)]
 mod tests {
-    use crate::buffer::SamplesBuffer;
-    use crate::source::Source;
+    use crate::{buffer::SamplesBuffer, source::Source};
```
est31, line 157: I don't like this change, can you keep it?

src/source/uniform.rs
```diff
@@ -1,14 +1,8 @@
-use std::cmp;
-use std::time::Duration;
+use std::{cmp, time::Duration};
```
est31, line 1: I don't like this change, can you keep it? The two changes below are fine, only issues with this.

src/source/zero.rs
```diff
@@ -1,7 +1,5 @@
-use crate::Sample;
-use crate::Source;
-use std::marker::PhantomData;
-use std::time::Duration;
+use crate::{Sample, Source};
+use std::{marker::PhantomData, time::Duration};
```
est31, line 2: I don't like this change, can you keep it? The sample, source change is fine, just issues with the second line.

src/static_buffer.rs
```diff
@@ -10,11 +10,9 @@
 //! ```
 //!
 
-use std::slice::Iter as SliceIter;
-use std::time::Duration;
+use std::{slice::Iter as SliceIter, time::Duration};
```
est31, line 13: I don't like this change, can you keep it?

src/static_buffer.rs
```diff
@@ -106,8 +104,7 @@ where
 
 #[cfg(test)]
 mod tests {
-    use crate::source::Source;
-    use crate::static_buffer::StaticSamplesBuffer;
+    use crate::{source::Source, static_buffer::StaticSamplesBuffer};
```
est31, line 107: I don't like this change, can you keep it?

est31 (changes requested):
Thanks for the pr. most of it is fine but there are some changes I dont like. Especially:

* `is_empty` is longer than `.len() == 0`. I disagree that the first form is any better.
* The overeager import regrouping hurts readability a lot. I don't like it. It's ok to unify `foo::bar::Baz` and `foo::bar::Bar` into `foo::bar::{Bar, Baz}` but not ok if Bar or Baz require additional path components, nor is it ok if there are nesting levels involved. *reordering* imports is fine so that everything with std follows each other etc, but not omitting the path components.

Most other changes are obvious improvements so greatly welcome. I'll merge if the comments are addressed.

src/conversions/sample_rate.rs
CryZe: There are quite a few other .is_empty() such as literally at the top of this inline diff at 191. Do I replace them with .len() >= 1 as well? Also what about .len() > 0 maybe?

est31: Alright let's go with your code then.

est31 (approved):
Thanks!

PR #357, opened 2021-03-08
Improve documentation about getting default output stream to a device

### LINKED ISSUE #355: Documentation is unclear since some functions removal
Main documentation is unclear and still have references to old functions that doesn't exist anymore 

See https://docs.rs/rodio/0.13.0/rodio/ -> Choose an output with the devices or default_output_device functions.

CleanCut:
Resolves #355 

/cc @grzi I ran into this as well!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -6,8 +6,8 @@
 //! - Create an object that represents the streaming sound. It can be a sine wave, a buffer, a
 //!   [decoder](decoder/index.html), etc. or even your own type that implements
 //!   [the `Source` trait](source/trait.Source.html).
-//! - Choose an output with the [`devices`](fn.devices.html) or
-//!   [`default_output_device`](fn.default_output_device.html) functions.
+//! - Get an output stream to a physical device. For example, get a stream to system's default sound
+//!   device with [`OutputStream::try_default()`](struct.OutputStream.html#method.try_default)
 //! - Call [`play_raw(output, source)`](fn.play_raw.html).
 //!
 //! The `play_raw` function expects the source to produce `f32`s, which may not be the case. If you
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,68 +1,77 @@
 //! Audio playback library.
 //!
-//! The main concept of this library is [the `Source` trait](source/trait.Source.html), which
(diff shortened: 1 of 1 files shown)


est31 (changes requested):
Thanks! Could you use intra-doc-links instead of linking to the html? https://doc.rust-lang.org/rustdoc/linking-to-items-by-name.html

est31:
Also not a native speaker but I think it should be "to *the* system's default sound device"

CleanCut:
> Also not a native speaker but I think it should be "to _the_ system's default sound device"

Good catch, fixed!

> Thanks! Could you use intra-doc-links instead of linking to the html?

Excellent, I've been wanting an excuse to try out the new doc linking method.

✅ It is done. Take another look!

I may have ended up going a bit overboard. 😊 I started by updating just my links, and then I went and fixed all the other links, and then I noticed the examples were out of date and broken so I fixed those, and I saw some wording that could be fixed... Anyway, I ended up refreshing quite a bit of the documentation.

- [x] All the links work for me when generated locally
- [x] The doctests all work for me locally.
  - Should we restore the `no_run` on the doctest that plays sound?  I prefer always running doctests, but I can understand why one would not want to if sound-less systems like CI crashed.

src/lib.rs
```diff
@@ -1,68 +1,77 @@
...
-//! The `play_raw` function expects the source to produce `f32`s, which may not be the case. If you
-//! get a compilation error, try calling `.convert_samples()` on the source to fix it.
+//! The [`play_raw`](OutputStreamHandle::play_raw) function expects the source to produce [`f32`]s,
+//! which may not be the case. If you get a compilation error, try calling
+//! [`.convert_samples()`](Source::convert_samples) on the source to fix it.
 //!
 //! For example, here is how you would play an audio file:
 //!
-//! ```no_run
+//! ```
```
est31, line 18: Travis CI failed... maybe add back the no_run here?

CleanCut: ✅ Sure thing

CleanCut:
Travis is now happy once again. Though I think server farms should include sound cards and pipe all the sound to the overhead public-address system as entertainment for those in the building. 😆

PR #362, opened 2021-03-28
Fix a error when rodio just panics instead of propagate an error to the user

xMAC94x:
and Update version to 0.13.1

### DIFF
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -28,7 +28,7 @@ impl OutputStream {
     pub fn try_from_device(
         device: &cpal::Device,
     ) -> Result<(Self, OutputStreamHandle), StreamError> {
-        let (mixer, _stream) = device.new_output_stream();
+        let (mixer, _stream) = device.try_new_output_stream()?;
         _stream.play()?;
         let out = Self { mixer, _stream };
         let handle = OutputStreamHandle {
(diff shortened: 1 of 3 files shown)


src/stream.rs
```diff
@@ -185,6 +188,27 @@ impl CpalDeviceExt for cpal::Device {
...
+
+        Ok(self
+            .new_output_stream_with_format(default_format)
+            .unwrap_or_else(|err| {
+                // look through all supported formats to see if another works
+                supported_output_formats(self)
+                    .filter_map(|format| self.new_output_stream_with_format(format).ok())
+                    .next()
+                    .ok_or(err)
+                    .expect("build_output_stream failed with all supported formats")
```
est31, line 208: Is there a way to convert this into an Err as well?

est31 (changes requested):

src/stream.rs
xMAC94x: @est31  prob, but i am not sure if this should be a NoDevice or another or a new error in the enum.
As i have no big domain knowledge i just did the minimal change which i could test (by deleting pulsaudio-alsa package) to reproduce that this is catched.
Feel free to put a commit on top :) but as ve have many people complaining why rodio is panicing we wanted to get a fix quick for now :)

est31 (approved):
Thanks

PR #376, opened 2021-06-27
Add optional Symphonia backend

### LINKED ISSUE #256: Switching to pure-rust mp3 decoder
We should consider switching to a pure-rust mp3 decoder.

There seem to be two alternatives, puremp3 and Symphonia:

* https://github.com/pdeljanov/Symphonia
* https://github.com/Herschel/puremp3

We should study their quality as well as their speed to find out whether adoption is feasible.

### LINKED ISSUE #351: AAC + MPEG/AAC decoder
I made the crate [redlux](https://crates.io/crates/redlux) which decodes AAC, as well as AAC from MPEG-4 containers. It uses [fdk-aac bindings](https://crates.io/crates/fdk-aac) to decode AAC and [rust-mp4](https://crates.io/crates/mp4) to read the MPEG-4 containers.

If it's something that could potentially be used by rodio, feel free to. It already implements the `Source` trait.

It does have some potential [caveats](https://github.com/probablykasper/redlux#caveats) though, and to decode MPEG-4 files it needs to get the size of the file.

aschey:
Resolves #351, Resolves #256

This adds the option to use Symphonia as an alternative backend. 
- Symphonia decoders are all disabled by default and can be enabled using the `symphonia-{codec name}` features.
- Symphonia decoders and the default decoders are mutually exclusive so the user must set `default-features = false` to enable any of the Symphonia features. 
- Added `symphonia-all` feature so the user doesn't have to enable each codec separately. 
- Added a compile error that should clearly explain the issue when both implementations of the same codec are enabled.
- Updated the CI script to run against both the Symphonia codecs and the default codecs.
- Updated the docs to explain how to use these features.


Cargo.toml
```diff
@@ -1,29 +1,44 @@
...
 authors = ["Pierre Krieger <pierre.krieger1708@gmail.com>"]
-license = "MIT OR Apache-2.0"
 description = "Audio playback library"
-keywords = ["audio", "playback", "gamedev"]
-repository = "https://github.com/RustAudio/rodio"
 documentation = "http://docs.rs/rodio"
 edition = "2018"
+keywords = ["audio", "playback", "gamedev"]
+license = "MIT OR Apache-2.0"
+name = "rodio"
```
est31, line 8: Why do you reorder this? The name is the most important thing. Then come the other things, but in that order please.

Cargo.toml
```diff
@@ -1,29 +1,44 @@
...
 
 [dependencies]
+claxon = {version = "0.4.2", optional = true}
 cpal = "0.13"
-claxon = { version = "0.4.2", optional = true }
-hound = { version = "3.3.1", optional = true }
-lewton = { version = "0.10", optional = true }
-minimp3 = { version = "0.5.0", optional = true }
+hound = {version = "3.3.1", optional = true}
+lewton = {version = "0.10", optional = true}
```
est31, line 16: Please don't reorder this

Cargo.toml
```diff
@@ -1,29 +1,44 @@
...
+minimp3 = {version = "0.5.0", optional = true}
+symphonia = {version = "0.3", optional = true, default-features = false}
+thiserror = "1.0.25"
 
 [features]
-default = ["flac", "vorbis", "wav", "mp3"]
+default = ["flac", "mp3", "vorbis", "wav"]
```
est31, line 22: Please don't reorder.

Cargo.toml
```diff
@@ -1,29 +1,44 @@
...
 mp3 = ["minimp3"]
+symphonia-aac = ["symphonia/aac"]
+symphonia-all = ["symphonia-aac", "symphonia-flac", "symphonia-isomp4", "symphonia-mp3", "symphonia-wav"]
+symphonia-flac = ["symphonia/flac"]
+symphonia-isomp4 = ["symphonia/isomp4"]
+symphonia-mp3 = ["symphonia/mp3"]
+symphonia-wav = ["symphonia/wav", "symphonia/pcm"]
+vorbis = ["lewton"]
 wasm-bindgen = ["cpal/wasm-bindgen"]
+wav = ["hound"]
```
est31, line 34: Please don't reorder.

src/lib.rs
```diff
@@ -84,6 +84,18 @@
 //! let source = source.take_duration(Duration::from_secs(5)).repeat_infinite();
 //! ```
 //!
+//! ## Alternative Decoder Backends
+//!
+//! [Symphonia](https://github.com/pdeljanov/Symphonia) is an alternative deocder library that can be used in place of many of the default backends.
+//! Currently, the main benefit is that Symphonia is the only decoder library that supports M4A and AAC,
+//! but it may be used to implement additional optional functionality in the future.
+//!
+//! To enable, set `default-features = false` and enable either the `symphonia-all` feature to enable all Symphonia codecs
```
est31, line 93: This makes it sound like `default-features` has to be set to false which is not the case. Maybe change the wording to make it more sound like a reccomendation?

src/decoder/mp3.rs
```diff
@@ -18,9 +18,12 @@ impl<R> Mp3Decoder<R>
...
     R: Read + Seek,
 {
-    pub fn new(data: R) -> Result<Self, ()> {
+    pub fn new(mut data: R) -> Result<Self, R> {
+        if !is_mp3(data.by_ref()) {
+            return Err(data);
+        }
         let mut decoder = Decoder::new(data);
-        let current_frame = decoder.next_frame().map_err(|_| ())?;
+        let current_frame = decoder.next_frame().unwrap();
```
est31, line 26: Why do you put an unwrap here, especially when before there used to be a map_err.

Cargo.toml
```diff
@@ -1,29 +1,44 @@
...
 cpal = "0.13"
-claxon = { version = "0.4.2", optional = true }
-hound = { version = "3.3.1", optional = true }
-lewton = { version = "0.10", optional = true }
-minimp3 = { version = "0.5.0", optional = true }
+hound = {version = "3.3.1", optional = true}
+lewton = {version = "0.10", optional = true}
+minimp3 = {version = "0.5.0", optional = true}
+symphonia = {version = "0.3", optional = true, default-features = false}
+thiserror = "1.0.25"
```
est31, line 19: I'm not in favour of adding thiserror, it pulls in a huge set of dependencies for very little improvement in comfort. Plus it has nothing to do with symphonia addition so it should be done in a separate PR if at all.

README.md
```diff
@@ -6,23 +6,27 @@
...
 Rust playback library.
 
- - Playback is handled by [cpal](https://github.com/RustAudio/cpal).
- - MP3 decoding is handled by [minimp3](https://github.com/lieff/minimp3).
- - WAV decoding is handled by [hound](https://github.com/ruud-v-a/hound).
- - Vorbis decoding is handled by [lewton](https://github.com/est31/lewton).
- - Flac decoding is handled by [claxon](https://github.com/ruuda/claxon).
+- Playback is handled by [CPAL](https://github.com/RustAudio/cpal).
+- MP3 decoding is handled by [Minimp3](https://github.com/lieff/minimp3).
+- WAV decoding is handled by [Hound](https://github.com/ruud-v-a/hound).
```
est31, line 11: Please keep the formatting as it used to be.

README.md
```diff
@@ -6,23 +6,27 @@
...
 [The documentation](http://docs.rs/rodio) contains an introduction to the library.
 
 ## License
-[License]: #license
+
+[license]: #license
 
 Licensed under either of
 
-* Apache License, Version 2.0, ([LICENSE-APACHE](LICENSE-APACHE) or http://www.apache.org/licenses/LICENSE-2.0), or
```
est31, line 24: Please keep the formatting as it used to be.

build.rs
```diff
@@ -0,0 +1,17 @@
+use cfg_aliases::cfg_aliases;
+
+fn main() {
+    // Add alias to see if any symphonia features are enabled
+    // This prevents having to copy/paste this large cfg check each time
+    cfg_aliases! {
```
est31, line 6: Is this build.rs file needed at all? I've thought enabling `symphonia-mp3` feature for example enables the `symphonia` feature anyways.

src/decoder/mod.rs
```diff
@@ -1,18 +1,40 @@
 //! Decodes samples from an audio file.
 
-use std::error::Error;
+#[cfg(all(feature = "symphonia-mp3", feature = "mp3"))]
+compile_error!("Both symphonia-mp3 and mp3 are enabled.
```
est31, line 4: Please don't add these compiler errors. Cargo features are designed to be additive and some distros enable *all* features of a crate. This causes headaches for them. This case can be handled gracefully in the code by always preferring one backend if both are enabled.

est31 (changes requested):

Cargo.toml
aschey: One of my VS Code extensions has a TOML formatter and I think it tries to order things alphabetically. I'll revert it.

src/decoder/mp3.rs
aschey: The `is_mp3` function already runs through this logic before it gets to this point so this should never return an error now. This mirrors the functionality used by the other codecs. I had to add this because the mp3 decoder used to be last but now the Symphonia one is last so this needs to return the input like the other codecs do.

Cargo.toml
aschey: I added it because I needed to map the errors that Symphonia returns to a DecoderError and it made it a little easier but I can do that manually instead.

build.rs
aschey: Oh, you're right. I didn't realize `#[cfg(feature = "symphonia")]` would work since there isn't an explicit feature just called symphonia. I'll get rid of this and use that instead.

src/decoder/mod.rs
aschey: Makes sense, I was thinking people would rather not have duplicate dependencies enabled to reduce binary size, but I can try to use the docs for that instead.

est31 (approved):
Thanks!

Shnatsel:
@pdeljanov FYI - you've got users!

Issue #377, opened 2021-06-29
decoder/mod.rs has infinite recursion in `impl Display`


compenguy:
When an error is thrown by decoder it causes infinite recursion:

```
#0  0x000055555599b674 in alloc::string::String::new ()                                                        [0/1306]
    at /rustc/e82b65026dcadf0998e3abf9dabc677f63160de7/library/alloc/src/string.rs:379
#1  0x00005555559a9b70 in alloc::string::{{impl}}::to_string<rodio::decoder::DecoderError> (self=0x5555568edce0)
    at /rustc/e82b65026dcadf0998e3abf9dabc677f63160de7/library/alloc/src/string.rs:2379
#2  0x00005555559ac3fd in rodio::decoder::{{impl}}::fmt (self=0x5555568edce0, f=0x7fffff7ff170)
    at /home/wpage/.cargo/git/checkouts/rodio-6458f6e95b054746/0988e8c/src/decoder/mod.rs:545
#3  0x00005555559a9b9f in alloc::string::{{impl}}::to_string<rodio::decoder::DecoderError> (self=0x5555568edce0)
    at /rustc/e82b65026dcadf0998e3abf9dabc677f63160de7/library/alloc/src/string.rs:2382
#4  0x00005555559ac3fd in rodio::decoder::{{impl}}::fmt (self=0x5555568edce0, f=0x7fffff7ff2a0)
    at /home/wpage/.cargo/git/checkouts/rodio-6458f6e95b054746/0988e8c/src/decoder/mod.rs:545
#5  0x00005555559a9b9f in alloc::string::{{impl}}::to_string<rodio::decoder::DecoderError> (self=0x5555568edce0)
    at /rustc/e82b65026dcadf0998e3abf9dabc677f63160de7/library/alloc/src/string.rs:2382
#6  0x00005555559ac3fd in rodio::decoder::{{impl}}::fmt (self=0x5555568edce0, f=0x7fffff7ff3d0)
    at /home/wpage/.cargo/git/checkouts/rodio-6458f6e95b054746/0988e8c/src/decoder/mod.rs:545
#7  0x00005555559a9b9f in alloc::string::{{impl}}::to_string<rodio::decoder::DecoderError> (self=0x5555568edce0)
    at /rustc/e82b65026dcadf0998e3abf9dabc677f63160de7/library/alloc/src/string.rs:2382
[...]

PR #378, opened 2021-06-29
Fix for #377 infinite recursion crash

compenguy:
Implementing the Display trait on a type, and calling self.to_string()
in its implementation results in infinite recursion.

The default implementation of ToString for sized types that implement
Display relies on the Display trait in its implementation.
Consequently, calling `self.to_string()` in the implementation of
Display leads to infinite recursion.

This failure has been directly observed on the DecoderError type, and
the fix here is to move the logic for rendering the error as text out of
`Error::description()` and into `impl Display`. As it stands, the
`Error` trait provides a default implementation for `description()`
based on the `Display` trait if provided, so this is a simple, clean
change.

The failure is presumed to exist for Mp4Type, as it follows a similar
pattern of invoking `self.to_string()` in the `Display` implementation.
The solution there is basically to reverse the logic for `from_str()`.

### DIFF
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -230,8 +230,16 @@ impl FromStr for Mp4Type {
 
 impl fmt::Display for Mp4Type {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        let res = self.to_string().to_lowercase();
-        write!(f, "{}", res)
+        let text = match self {
+            Mp4Type::Mp4 => "mp4",
+            Mp4Type::M4a => "m4a",
+            Mp4Type::M4p => "m4p",
+            Mp4Type::M4b => "m4b",
+            Mp4Type::M4r => "m4r",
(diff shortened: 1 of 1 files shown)


src/decoder/mod.rs
```diff
@@ -542,11 +550,25 @@ pub enum DecoderError {
...
+            DecoderError::ResetRequired => "Reset required",
+            #[cfg(feature = "symphonia")]
+            DecoderError::NoStreams => "No streams",
+        };
+        write!(f, "{}", text)
     }
 }
 
 impl Error for DecoderError {
+    /*
```
est31, line 571: Can you instead completely remove this?

est31 (changes requested):

src/decoder/mod.rs
compenguy: Yes.  That's my mistake - I forgot to review my changes after I tested them.

est31 (approved):
Thanks!

PR #380, opened 2021-06-30
Fix sample rate for symphonia decoder

### LINKED ISSUE #379: AAC tracks playing back at 2x speed
Using the new symphonia backends, playing back an AAC track in an MP4 container yields 2x speed playback through rodio.  This behavior is not observed when playing the same track through the symphonia-play utility, nor is it observed when manually setting the rodio decoder to use redlux.

This may be a mismatch in the sample rate being reported to rodio for these tracks, as they all use SBR, and some additionally use PS, to allow (but not require) upconversion from a lower base sample rate to a higher rate.

aschey:
Resolves #379 

Use the sample rate from the sample instead of the file metadata. Thanks @compenguy for finding the root cause.

### DIFF
--- a/src/decoder/symphonia.rs
+++ b/src/decoder/symphonia.rs
@@ -1,7 +1,7 @@
 use std::time::Duration;
 use symphonia::{
     core::{
-        audio::SampleBuffer,
+        audio::{AudioBufferRef, SampleBuffer, SignalSpec},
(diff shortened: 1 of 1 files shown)


compenguy:
I tested from the source branch, it works for me.

est31 (approved):
Thanks!

PR #383, opened 2021-07-06
Fix `WavDecoder::total_duration`

### LINKED ISSUE #382: `total_duration` for Wav is not consistent with other implementations
Currently `WavDecoder`'s `total_duration` returns the remaining duration of the sound ([code](https://github.com/RustAudio/rodio/blob/f6bb97e0d32b81ad7dbd9d20e6b7dc32e23a658f/src/decoder/wav.rs#L115-L120)). This is not consistent with the rest of the crate, where the `total_duration` is stored at the beginning and always returns the same value.

However, I thing that the logical thing to do is to return the *remaining* duration, as it is generally the interesting information. Additionally, it would be consistent with how `Iterator::size_hint` and `ExactSizeIterator::len` work.

So we should decide what we prefer, document it, and make it consistent within the crate.

a1phyr:
Fixes #382 

Also make `total_duration` microsecond-precise to be consistent with `FlacDecoder`.

### DIFF
--- a/src/decoder/wav.rs
+++ b/src/decoder/wav.rs
@@ -11,6 +11,7 @@ where
     R: Read + Seek,
 {
     reader: SamplesIterator<R>,
+    total_duration: Duration,
     sample_rate: u32,
     channels: u16,
 }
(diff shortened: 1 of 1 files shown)


est31 (approved):
Thanks!

PR #388, opened 2021-07-17
Check if the frame length is zero in SkipDuration.

### LINKED ISSUE #387: Skipping past the end of an OGG file causes infinite loop.
When I run the following code, "Skipping..." is printed but not "Successfully skipped 31 seconds of audio!". 

```rust
use rodio::{source::Source, Decoder};
use std::fs::File;
use std::io::BufReader;
use std::time::Duration;

fn main() {
    // metronome120.ogg is an ogg file with 15 measures of a 120 BPM click track
    // (hence the file is only 30.0 seconds long).
    let file = BufReader::new(File::open("src/metronome120.ogg").unwrap());
    let source = Decoder::new(file).unwrap();

    println!("Skipping...");
    let _ = source.skip_duration(Duration::from_secs_f32(31.0));
    // The line below will never print, instead the program just locks up on the line above.
    println!("Successfully skipped 31 seconds of audio!");
}
```

Note that this seems to happen only with ogg files. When I try this with a WAV file instead, the program prints the second line as expected. It also only happens when going "out of bounds" of the ogg file--if I only skip 5 seconds, then the program works also.

Here is a zip containing a minimal example: [rodio-bug.zip](https://github.com/RustAudio/rodio/files/6835140/rodio-bug.zip)

I am using version `0.14.0` of Rodio on Windows 10.

a2aaron:
If the source we are skipping runs out of data, it will return Some(0) as the frame length. Before, this would cause an infinite loop because we would skip zero samples forever. This PR fixes it by checking explicitly if the frame length is zero and bailing out if it is.

Fixes #387

### DIFF
--- a/src/source/skip.rs
+++ b/src/source/skip.rs
@@ -33,6 +33,12 @@ where
         // .unwrap() safety: if `current_frame_len()` is None, the body of the `if` statement
         // above returns before we get here.
         let frame_len: usize = input.current_frame_len().unwrap();
+        // If frame_len is zero, then there is no more data to skip. Instead
+        // just bail out.
+        if frame_len == 0 {
+            return;
+        }
+
         let ns_per_sample: u128 =
             NS_PER_SECOND / input.sample_rate() as u128 / input.channels() as u128;
 


est31 (approved):
Thanks!

Issue #353, opened 2021-02-18
take_duration inaccurate time durations


fculross:
When trying to set start and end points for a sound loaded from an ogg file into a sink, skip_duration correctly starts playback from the required time but take_duration appears to somehow be dividing the desired time by a factor of just under 5. So a take_duration set to say, 40 seconds actually stops playback after just over 8.3 seconds! (OS is Windows 10 - code used shown below.) 

UPDATE: have also now tried same code on Mac OS X El Capitan with same result.]

Has anyone else encountered this problem? 

    let (_stream, stream_handle) = rodio::OutputStream::try_default().unwrap();
    let sink = rodio::Sink::try_new(&stream_handle).unwrap();
    let file = File::open(&self.audio_file_name).unwrap();
    let source = rodio::Decoder::new(BufReader::new(file)).unwrap();
    let source = source.skip_duration(Duration::from_secs(40)).take_duration(Duration::from_secs(40));
    sink.append(source);

fculross:
Although I'm new to Rust, I did a closer comparison of the source for "take" with that of "skip" and found what I believe to be the cause of the problem. In line 35 of skip, nanoseconds per sample is calculated as:

    let ns_per_sample: u128 =
            NS_PER_SECOND / input.sample_rate() as u128 / input.channels() as u128; 

while at line 70 of take, the equivalent is given as:

    let ns = NANOS_PER_SEC / input.sample_rate() as u64 * input.channels() as u64;

I created a local version of the take.rs file and changed line 70 to match the logic suggested in skip:

    let ns = NANOS_PER_SEC / input.sample_rate() as u64 / input.channels() as u64;

By using the modified local file instead, I can now correctly play from my desired start point to desired end point, so this simple logic error seems to be the issue - hopefully this would be an easy fix for a future release.

PR #389, opened 2021-07-22
Fix inaccurate calculations in take_duration

theoahfeldt:
See #353 .


src/source/take.rs
```diff
@@ -66,7 +66,7 @@ where
     /// Returns the duration elapsed for each sample extracted.
     #[inline]
     fn get_duration_per_sample(input: &I) -> Duration {
-        let ns = NANOS_PER_SEC / input.sample_rate() as u64 * input.channels() as u64;
```
est31, line 69: Can you add () instead of doing two divisions?

est31 (changes requested):

Issue #386, opened 2021-07-16
The first sample from a stereo source is sometimes played in the wrong channel


tuzz:
Hello, thanks for maintaining this crate.

I'm trying to write a Source that balances the volumes of the left and right stereo channels separately.
I noticed that the following minimal program sometimes doesn't do what I expect.
It plays the sound in the left channel ~95% of the time and the right channel ~5% of the time.

Shouldn't the first call to `Source#next()` be the sample played in the left channel?
I've noticed that it does the right thing if the OutputStream is recreated in each iteration of the loop.
If you run the program with `cargo run --release` it should reproduce the problem. I'm running macOS 11.4.

Thanks

```rust
use rodio::*;
use std::{io::Cursor, time::Duration, thread::sleep};

fn main() {
    // Create a source from the bytes of an ogg file.
[...]

Issue #286, opened 2020-05-02
Left and right channels seem to be swapped


Keithcat1:
I'm on windows 64-bit.
Any audio files seem to play with their left and right channels swapped.

hugovdm:
Potentially related: I've played with CPAL's "feedback" example (in version 0.12) on my Linux laptop (with ALSA). It seemed like *sometimes* the channels were swapped, and *sometimes* they weren't (talking input vs output).

My conclusion: CPAL isn't managing to guarantee (on my Linux laptop with ALSA) that default output and input settings alwayus produces the two channels in the same order. *Maybe* this CPAL behaviour could be the source of this bug in rodio...

(If swapped, one hears the "alternating left-right" effect quite well, and I didn't always hear it. And dropping either the L or the R channel made the feedback die out quickly, whereas not having them swapped meant consistent slow decay, as expected, for the channel not attenuated - I was generally setting my mixer settings to give between 0.9 and 1.0 end-to-end gain.)

keigen-shu:
I'm experiencing this issue on Linux too. For me, the output is always swapped.

Here's a simplified code to test device output. To be fair, this makes the conventional assumption that the first sample in a stereo frame maps to the left output, and the second sample maps to the right output.

```rust
extern crate rodio;

use rodio::source::Source;

fn main() {
    let device = rodio::default_input_device().unwrap();
    let sink = rodio::Sink::new(&device);

    let lt_ch: Vec<f32> = vec![0.0, 0.1, 0.0, -0.1]; // triangle signal on left
    let rt_ch: Vec<f32> = vec![0.0, 0.0, 0.0, 0.0]; // silent on right
    let data: Vec<f32> = lt_ch.iter().zip(rt_ch.iter()).flat_map(|(&a, &b)| vec![a, b]).collect();

    let buffer = rodio::buffer::SamplesBuffer::new(2, 400, data); // run at 400 Hz

    sink.append(buffer.repeat_infinite());

    std::thread::sleep_ms(5000);
}
```

paulirotta:
This is still happening. For me on Linux but not in Windows. It is almost always reversed, but I did once hear the correct left-right channel mapping.

tuzz:
I've opened a pull request ^ that might fix this issue. If someone could kindly check, that would be great. I don't really have access to a linux/windows machine. Thanks

keigen-shu:
Hello,

[...]

Issue #252, opened 2019-10-18
Mono ogg files on a different sample rate play through wrong channels with ChannelVolume


SpiralP:
When playing a mono ogg file that is a different sample rate than your output device's through `ChannelVolume` the audio switches between the left and right channels.  (I think it has to do with compression because it usually happens on the beat of a song)

I've narrowed the issue down to `SampleRateConverter`'s [Iterator.next() function](https://github.com/RustAudio/rodio/blob/e5474a2ef15f2d0a3bae2538de159b6d3e5bdf79/src/conversions/sample_rate.rs#L123), but I don't know much about converting sample rate so I'm lost as to where the problem actually is.

This issue could be related to the lower level decoding library `lewton`, but since the ogg plays perfectly fine when the sample rates match, I think it's related to `SampleRateConverter` in rodio.

I found this issue when using Windows 10 with my output sample rate set to 48000 in the sound settings.  I was trying to play a 44100 Hz sound through the `SpatialSink`, which in turn uses `ChannelVolume`.

Here's a minimal repro-repo: https://github.com/SpiralP/rust-rodio-samplerate-bug

Thanks!

Cocalus:
I fixed some issues like this in #249 which might help but I suspect this is a different issue with the same root cause of generated samples not respecting the number of channels. I suspect Rodio could benefit from switching to using the "sample" crate's Frame trait to avoid this issue at the type level.

will3942:
@dvdsk Flagging that this issue is affecting us too when resampling from 48kHz to a 44.1kHz OutputDevice for an 8 channel source (or a 6 channel) when using ChannelVolume but probably also on multichannel sources without ChannelVolume.

This may be related to https://github.com/RustAudio/rodio/pull/684

PR #390, opened 2021-07-22
Fix sounds playing on the wrong channels

tuzz:
Previously, the DynamicMixer implementation had a problem where it would start playing pending sources immediately. This meant that sometimes the first sample of a source was sent to the wrong audio channel. For example, if a stereo source starts playing on an odd-numbered call to DynamicMixer::next() then its left/right channels would be reversed because the output is expecting the right-channel's sample at that time.

This fixes https://github.com/RustAudio/rodio/issues/386 and possibly https://github.com/RustAudio/rodio/issues/286 (I haven't checked) and maybe even https://github.com/RustAudio/rodio/issues/252 (I haven't checked).

This is my first PR to rodio. I'm not really sure if there's a style guide or anything so any suggestions would be more than welcome! Thanks

### DIFF
--- a/src/dynamic_mixer.rs
+++ b/src/dynamic_mixer.rs
@@ -30,6 +30,8 @@ where
     let output = DynamicMixer {
         current_sources: Vec::with_capacity(16),
         input: input.clone(),
+        sample_count: 0,
+        still_pending: vec![],
     };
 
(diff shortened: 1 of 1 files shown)


est31 (approved):
Thanks!

Issue #391, opened 2021-08-09
Change speed of a Sink?


tree786:
Hallo

i would like to change playback speed in realtime for my application, however it seems there is no 'speed' field for Sink, but only for Decoder (which is not useful for my case). Any ways to change speed of a sink?

Thanks

Issue #177, opened 2018-08-20
Playback speed on sinks


nico-abram:
Would it be possible to add a playback speed to sinks?

PR #398, opened 2021-10-25
Change Sink playback speed

PieKing1215:
Adds the ability to change the playback speed of a Sink while playing.

Adds `Sink::set_speed`, `Sink::speed`, `SpatialSink::set_speed`, and `SpatialSink::speed` (also adds `Speed::set_factor` like Amplify has).
These fns get/set a new `speed` field in `Controls`, and this PR modifies `Sink::append` to use it.

Basically I just copied what was already there for volume and it seems to just work?
I haven't tested this very thoroughly because I don't really know what might break.
I have a [fork of ggez](https://github.com/PieKing1215/ggez/tree/changing_pitch) that I set up to make use of this feature and it seems to work in that case at least.

Related: #177 & #391

### DIFF
--- a/src/sink.rs
+++ b/src/sink.rs
@@ -24,6 +24,7 @@ struct Controls {
     pause: AtomicBool,
     volume: Mutex<f32>,
     stopped: AtomicBool,
+    speed: Mutex<f32>,
 }
 
 impl Sink {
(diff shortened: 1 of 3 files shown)


est31 (approved):
Thanks!

PR #399, opened 2021-11-16
Change SineWave::new() to take a float.

nwmahoney:
It was immediately casting it to a float anyway.

### DIFF
--- a/src/conversions/sample_rate.rs
+++ b/src/conversions/sample_rate.rs
@@ -341,7 +341,7 @@ mod test {
         /// Check that resampling does not change the audio duration,
         ///  except by a negligible amount (± 1ms).  Reproduces #316.
         /// Ignored, pending a bug fix.
-        fn preserve_durations(d: Duration, freq: u32, to: u32) -> () {
+        fn preserve_durations(d: Duration, freq: f32, to: u32) -> () {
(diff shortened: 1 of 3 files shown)


aromeronavia:
I also thought this would have been a better interface for this implementation, although I don't know if the maintainers would like to approve this breaking change (as the compiler will suddenly stop working for people using the i32 implementation).

est31:
Yeah I'm generally in favour of this PR, but it can't be merged until a new semver version of rodio due to the issue you pointed out. So it has to wait.

nwmahoney:
Waiting for a new semver version seems right to me. Do you have an idea of when a new version including this change might be released?

est31:
IDK there are tons of users of rodio and inconveniencing them to do a semver update requires an important reason. If there's some bigger change than just the type change it might make sense to do a semver breaking update.

Rdna123:
Since SineWave only has two variables, the inner valuables of it could be made public without doing a semver breaking update

est31:
I think exposing a different constructor that takes a float is a better idea. Then in a future semver incompatible upgrade, one can change new to use the new constructor.

est31:
That being said, there are good arguments in #408 for why a major semver bump should be done. So I'll merge this.

Issue #401, opened 2021-11-20
aac file doesn't play in rodio with error «invalid data»


Rrogntudju:
I first opened an issue in symphonia: https://github.com/pdeljanov/Symphonia/issues/67

Please take a look at the symphonia maintainer answer.

est31:
Interesting. I guess decode errors need to be ignored. PRs welcome!

pdeljanov:
Just wanted to clarify something since my answer to the linked issue was a bit simplistic.

A decode error produced by `Decoder::decode` is not considered fatal. This is because if there is an error (corruption, invalid data, etc.), it's usually limited to that individual packet. The correct action is to just get a new packet and try again. However, a decode error anywhere else is fatal.

I plan to make fatal errors a distinct error kind in the future, but I haven't gotten to it yet unfortunately.

est31:
Fixed by #403

PR #403, opened 2021-11-29
Try next packet after a decode error (Symphonia)

Rrogntudju:
PR for issue #401. In symphonia, a decode error is not fatal and one must try the next packet in case of a decode error.
However a limit is set to the number of successive packets with decode errors. Only tested with aac files.


est31 (approved):
Thanks!

PR #407, opened 2022-01-13
Replace loop with while

gridbugs:
Tiny PR that just applied the feedback from https://github.com/RustAudio/rodio/pull/285 (now merged), replacing an explicit `loop` with a `while let` loop.


est31 (approved):
Thanks!

PR #408, opened 2022-01-19
Release 0.15.0

gridbugs:
Prepares the changelog for 0.15 (I've just included externally-visible changes) and bumps version number. In a separate commit I've updated the symphonia dep to 0.4 (please lmk if you'd rather that go in its own PR). Am I correct in understanding that as soon as this is merged, the release will take place automatically?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+# Version 0.15.0 (2022-01-23)
+
+- Remove requirement that the argument `Decoder::new` and `LoopedDecoder::new` implement `Send`.
+- Add optional symphonia backend.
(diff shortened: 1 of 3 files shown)


est31:
Why a minor semver update and not a patch update?

gridbugs:
Because of

> `WavDecoder`'s `total_duration` now returns the total duration of the sound rather than the remaining duration.

Think anyone is likely depending on the old behaviour and would be surprised if it changed in a patch update?

est31:
I see, makes sense.

est31:
Can you incorporate #399 too?

gridbugs:
Rebased to include https://github.com/RustAudio/rodio/pull/399 and https://github.com/RustAudio/rodio/pull/394 as well and updated changelog.

PR #410, opened 2022-01-24
Allow creating a SymphoniaDecoder directly from a FormatReader

Alch-Emi:
Hello!

I'm working on using `rodio` for audio playback in a personal application, with Symphonia as my decoder.  The problem is, I've already parsed the data I'm working with (in my case, to extract the metadata).  This leaves me in the interesting situation of having a partially decoded file that needs to be thrown away and read from scratch in order to be played.

As it stands, `rodio` already has all of the internals for playing the `FormatReader` that any user who has already started deconstructing a song in Symphonia would have.  I believe that creating a user-friendly interface to instantiate a decoder from a `FormatReader` (with the disclaimer that this is not the most common usecase) is an easy way to expand the functionality of `rodio` without adding too much complexity or unnecessary code.

This PR does exactly this, by creating the new `pub fn new_from_format_reader(..)` method for the `Decoder`, which is enabled when any of the `symphonia-*` features are enabled.  To facilitate this, the PR also makes a slight internal change to the `SymphoniaDecoder` that just adds another entrypoint to the constructor.

This should be an entirely non-breaking change, and the aforementioned new `Decoder` constructor is the only public facing API change.

I'd love to hear your thoughts on this!

### DIFF
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -13,7 +13,10 @@ use crate::Source;
 #[cfg(feature = "symphonia")]
 use self::read_seek_source::ReadSeekSource;
 #[cfg(feature = "symphonia")]
-use ::symphonia::core::io::{MediaSource, MediaSourceStream};
+use ::symphonia::core::{
+	io::{MediaSource, MediaSourceStream},
+	formats::FormatReader,
+};
 
 #[cfg(all(feature = "flac", not(feature = "symphonia-flac")))]
 mod flac;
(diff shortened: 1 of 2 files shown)


est31:
I think a proper metadata API would be better than exposing functionality that only works with one decoder backend.

Alch-Emi:
That's fair!  I got the general vibe that a metadata api was a long way down the roadmap, and a pretty big task for a newbie to the project to take on by themself, which is why I went for this solution.

If there is a plan for a metadata api in the near future, I don't really mind using a private repo for development for now, but if it's more of a down the road thing, it might be good to have a stopgap?  Especially since the code is all already there, it's just a difference between creating a symphonia decoder from an arbitrary data source vs creating a symphonia decoder from a FormatReader.

That said, if you don't think that this is a good call, I totally respect your decision on that.  If that's the case though, (and if there's no planned metadata api) could I instead use this thread to ask for your advice for what I should do to work with metadata in the meantime?

est31:
The issue with this function is that it exposes symphonia internals. This means that rodio suddenly has to follow symphonia when it does breaking semver updates.

Generally there should be low overhead in extracting the metadata only and then throwing away the datastructure. The metadata is not used for decoding anyway.

Alch-Emi:
Alright, that makes sense.  Thank you for your help

PR #415, opened 2022-02-17
remove the need for a source to be debug on a spatial sink

mockersf:
[`SpatialSink::append`](https://docs.rs/rodio/latest/rodio/struct.SpatialSink.html#method.append) requires the source to be `Debug`, while [`Sink::append`](https://docs.rs/rodio/latest/rodio/struct.Sink.html#method.append) doesn't.

I removed the bound, and all the other that were needed to get it to compile again

### DIFF
--- a/src/source/spatial.rs
+++ b/src/source/spatial.rs
@@ -1,4 +1,3 @@
-use std::fmt::Debug;
 use std::time::Duration;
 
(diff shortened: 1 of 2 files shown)


est31 (approved):
Thanks!

PR #416, opened 2022-02-17
spatial: ignore NaNs when setting the volume of each ear

### LINKED ISSUE #284: Spatial / SpatialSink outputs no sound when left_ear == right_ear
I suspect that it is caused by the division by zero in Spatial::set_positions
```rust
        let max_diff = dist_sq(left_ear, right_ear).sqrt();
        let left_dist = left_dist_sq.sqrt();
        let right_dist = right_dist_sq.sqrt();
        let left_diff_modifier = ((left_dist - right_dist) / max_diff + 1.0) / 4.0 + 0.5;
```

mockersf:
fixes #284 

When the two ears have the same position, their distance would be 0.0 which led to NaN for the modifiers. I ignored the value when it's a NaN

### DIFF
--- a/src/source/spatial.rs
+++ b/src/source/spatial.rs
@@ -60,10 +60,18 @@ where
         let right_diff_modifier = ((right_dist - left_dist) / max_diff + 1.0) / 4.0 + 0.5;
         let left_dist_modifier = (1.0 / left_dist_sq).min(1.0);
         let right_dist_modifier = (1.0 / right_dist_sq).min(1.0);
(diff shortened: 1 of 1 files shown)


est31:
I think a panic would make more sense? The code wasn't built with one ear in mind, was it?

mockersf:
I would like to avoid a panic here as it could be hard to catch in anything integrating it.
I would like to use it from Bevy, and would prefer to have a failure case of "the sound is just playing".

PR #419, opened 2022-02-28
Use pinboard crate to avoid blocking Sinks during playback

sapir:
Calls to `Sink` and `SpatialSink`'s set methods risk causing delays to the playback thread. This pull requests attempts to fix that by replacing the mutex with a lock-free `Pinboard` from the [pinboard](https://lib.rs/pinboard) crate.


dvdsk:
Sink is set to be deprecated then removed. In its place we will get a new Source called Player. I will definitely use some of ideas from this PR for that. Thank you for your work on this!

Since Sink will be removed we are no longer working on it, because of that ill have to close this PR, sorry.

PR #421, opened 2022-03-26
Move sound to `assets/`; Exclude it from package

### LINKED ISSUE #420: Exclude examples from crates.io package
Right now, the [crates.io package](https://crates.io/crates/rodio) weighs an astounding 7.41 MB. Most of it is composed of the example files found in the `examples/` directory. I think that [excluding these files](https://doc.rust-lang.org/cargo/reference/manifest.html#the-exclude-and-include-fields) from the package would greatly help in reducing the size of the crate.

aleokdev:
Closes #420 (nice).
Moves sound files to `assets/` and excludes them from the crates.io package, saving 6.1MiB (!!) of data.


est31:
`cargo test` if you run it from the cargo package is not broken, right? Because crater does that, and I would prefer not breaking it.

aleokdev:
Sadly it does panic because it can't find the files, but I'm not sure how to handle it otherwise. I guess it could just skip the tests, does that sound good?

est31:
@aleokdev maybe exclude the examples directory too?

aleokdev:
> @aleokdev maybe exclude the examples directory too?

I've seen many crates (e.g. bevy) avoid doing this for some reason, and the [cargo book](https://doc.rust-lang.org/cargo/reference/manifest.html#the-exclude-and-include-fields), as an example, explicitly includes the examples directory, so I'm not sure about that.

How about (only) including `examples`, `src` as well as files in the root (e.g. license)?

est31:
I think excluding `examples` and `tests` without changing anything else is the cleanest approach.

aleokdev:
Sadly the examples can't go because they are declared in `Cargo.toml`, and so if they are removed they will point to invalid paths.

aleokdev:
I excluded `tests`. The current way is the cleanest way to do it without breaking anything, I think.

est31 (approved):
Thanks!

I've checked locally, `cargo test` does not seem to *run* the examples, only compile them. If they compile but don't run, that's enough for us.

PR #430, opened 2022-07-05
add `Callback` source

goolord:
this lets you execute code when the sink is empty, when you've reached a certain position in the sink, etc. it's useful if you can't/don't want to poll constantly to see if the sink has emptied


ivakam:
What is the status on this and could you provide an example of how to use it? I needed I can rebase it against master (looking at it it seems to be a trivial rebase).

goolord:
using it here 
https://github.com/goolord/tundra/blob/c3e4515ad60b86739638952d93d9a50031162527/src/source/callback.rs
https://github.com/goolord/tundra/blob/aa85b23e6fccde69fa6cba4dd1b6225c3079b1b9/src/types/player.rs#L256

feels ready to merge

goolord:
bump, wasn't sure how useful this was but it appears that someone else has wanted it at least https://github.com/goolord/tundra/issues/3

src/source/callback.rs
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    #[inline]
+    fn channels(&self) -> u16 {
+        1
+    }
+
+    #[inline]
+    fn sample_rate(&self) -> u32 {
+        48000
```
est31, line 48: Why is this not adjustable?

src/source/callback.rs
```diff
@@ -0,0 +1,55 @@
...
+    S: Sample,
+{
+    #[inline]
+    fn current_frame_len(&self) -> Option<usize> {
+        None
+    }
+
+    #[inline]
+    fn channels(&self) -> u16 {
+        1
```
est31, line 43: Why is this not adjustable?

est31 (changes requested):

src/source/callback.rs
goolord: cargo culted from https://github.com/RustAudio/rodio/blob/master/src/source/empty.rs

i imagine it doesn't actually matter since this source never contains audio

est31: Fair, I rest my case.

est31:
@goolord it can easily be confused with a source that invokes a callback to generate samples and it was my first idea when I read this PR. I think it would be better to rename it to `EmptyCallback`.

src/source/empty_callback.rs
```diff
@@ -4,22 +4,22 @@ use std::time::Duration;
 use crate::{Sample, Source};
 
 /// A source which execute a callback function
```
est31, line 6:
```suggestion
/// An empty source which executes a callback function 
```

est31 (approved):
Except for the docs issue, this is good to go

est31 (approved):

PR #441, opened 2022-09-12
Release version 0.16.0

enfipy:
Bumps the version to 0.16.0 to trigger CI publishing.

Releases `cpal` and `symphonia` updates.

P.S: Feel free to mention other changes that should be put inside `CHANGELOG.md`.


est31:
I think it needs to be `0.16` because it uses cpal's errors as part of the public API.

enfipy:
@est31 Updated to 0.16.0 ✌️

est31 (approved):
Thanks!

PR #447, opened 2022-10-14
Update sine.rs

DanikVitek:


est31:
Thanks but I like PI more.

PR #449, opened 2022-10-20
Fix: Audio playback not working after stop on sink

pixelafk:
Hi,

I'm working on a project of my own in which I'm using rodio for audio playback, however I came across the issue that I am unable to play audio after I've called the `.stop` method on `rodio::sink::Sink`, in my project I've fixed this by changing the `.play` method  on Sink to this:
```rust
    #[inline]
    pub fn play(&self) {
        self.controls.stopped.store(false, Ordering::SeqCst);
        self.controls.pause.store(false, Ordering::SeqCst);
    }
```

At first I thought the documentation for `.stop` was wrong, but it turns out it does actually clear the queue but it completely prevents your Sink from doing any more audio playback (at least I wasn't able to figure out how to make it play anything after the fact).

This fix works for me, and I'd like some feedback on whether this is a feasible solution, warning I'm pretty new to Rust.

P.S. the tests ran fine

### DIFF
--- a/src/sink.rs
+++ b/src/sink.rs
@@ -132,6 +132,7 @@ impl Sink {
     /// No effect if not paused.
     #[inline]
     pub fn play(&self) {
+        self.controls.stopped.store(false, Ordering::SeqCst);
         self.controls.pause.store(false, Ordering::SeqCst);
     }
 


pixelafk:
I just discovered a bug in my code, if you immediately after calling .stop on the sink try to append something to the queue using .append, something like this:
```rust
...
sink.stop();
sink.append(source);
sink.play();
```
It doesn't actually clear the queue but instead just continues playback and appends the source to the end of the queue like .stop() was never called.

BUT: if you wait a little bit (longer than instantly) after calling .stop and then call `.append(source)` it does work, it clears the queue and appends to the queue just fine.

pixelafk:
A cheeky workaround I found for this is doing this:
```rust
...
sink.stop();
while !audio.sink.empty() {audio.sink.sleep_until_end()}
sink.append(source);
sink.play();
...
```

I think this indicates that if `.append` is called before the queue is emptied the queue doesn't get emptied at all and instead the source just gets appended to the end of the queue like normal, completely ignoring the `.stop` call.

myyrakle:
This is a feature I need too. I hope it can be applied in some form.

pixelafk:
Closing in favor of #464

Issue #423, opened 2022-03-28
how to pass a reference to sink append


alexzanderr:
i have this code
![image](https://user-images.githubusercontent.com/54772464/160490932-899d015b-9e00-4c77-a642-d814ad87e2f2.png)

i saw that i cannot append more than once the same sink because source it doesnt have copy trait.

but how do i pass a reference to `source` or use multiple times ?

because i want to play the same sound multiple times. this way the program will be more efficient, instead of loading the file from disk everytime i want to the the audio file.

load `once`, play `multiple times`.

how can i achieve this ?

thanks in advance.

VladasZ:
I have the same question. How can I reuse file data if all methods take it by value and move it?

alexzanderr:
i guess `clone` is a temporary solution

Keithcat1:
Create a new source based on rodio::static_buffer::StaticSamplesBuffer which, instead of accepting a &'static [S], accepts an Arc<Vec<S>> which allows you to load a file from disk once, decode it into a Vec<S> by taking advantage that all sources (including Decoders) are iterators, then wrap that Vec in an Arc which allows it to be read efficiently but not written from as many different sources at a time as you want.

Ownezx:
So I proposed a workaround with buffered sources and clones. But the downside is that the types become painfully long as per the example : `Buffered<TakeDuration<SkipDuration<Buffered<Decoder<BufReader<File>>>>>>`

RottenFishbone:
I believe the solution in #141 applies to this.
It worked well in my case.

PR #451, opened 2022-10-28
Added a storing file example.

Ownezx:
This is an example of loading files and playing them several times after reading issue #423. It would not make a lot of sense for larger audio as it is not streaming.

One downside of that is that, when adding operation to the saved file, the type becomes painfully long in rust as per the example : `Buffered<TakeDuration<SkipDuration<Buffered<Decoder<BufReader<File>>>>>>`.

I'm not really sure how to address that for now.

### DIFF
--- a/examples/storing_file.rs
+++ b/examples/storing_file.rs
@@ -1,0 +1,31 @@
+use rodio::source::{Buffered, SkipDuration, TakeDuration};
+use rodio::{source::Source, Decoder, Sink};
+use std::fs::File;
+use std::io::BufReader;
+use std::time::Duration;
+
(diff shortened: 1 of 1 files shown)


dvdsk:
> I'm not really sure how to address that for now.

I think you could solve that by putting the source in a Box. Then the type could be coerced to Box<dyn Source + Clone> using `as Box<dyn Source + clone>`. This comes at a slight performance cost. An alternative is not naming the entire type. Just: `Buffered<_>` might work.

dvdsk:
we now have the `into_file.rs` example using the new `output_to_wav` function.

PR #454, opened 2022-11-18
Enable gapless playback

kuviman:
Only when using symphonia

I have no idea why it is not the default

minimp3 still has the issue, it seems like there is a way fix minimp3 as well using minimp3-ex or something: https://github.com/lieff/minimp3/issues/63


kuviman:
This resolves the issue of a small silence in the beginning of mp3's

This can be heard if looping or sequencing

Here's an example repo: https://github.com/kuviman/mp3-test

est31 (approved):

Issue #369, opened 2021-05-21
Spatial audio is glitchy


nobbele:
```rust
fn main() {
    let (_stream, handle) = rodio::OutputStream::try_default().unwrap();
    let sink =
        rodio::SpatialSink::try_new(&handle, [-2.0, 0.0, 0.0], [1.0, 0.0, 0.0], [-1.0, 0.0, 0.0])
            .unwrap();

    let file = std::fs::File::open("beep.wav").unwrap();

    let source = rodio::Decoder::new(file).unwrap();
    sink.append(source);

    sink.sleep_until_end();
}
```
using this code (use any beep.wav file), nearing the end of the audio, it will start to play the sound in the other ear before switching back.

DaforLynx:
Testing this code with a very short sound (less than half a second) over and over, this glitchiness is even more apparent. And when run in quick succession, the sound will occasionally swap sides for its duration.

denizsincar29:
It is a very serious bug that rodio developers doesn't seam to fix. I also wrote an issue on this.

denizsincar29:
upd. This issue is almost 1 year old! Dear developers. What are you doing for 1 year? Rodio is the most popular crate but has ridiculous bugs.

est31:
@denizsincar29 bumping old threads every now and then is fine-ish, but the way you expressed your frustration is absolutely not. Software often has bugs for way longer than one year. This is a project ran by volunteers, and we are not here to read such comments.

denizsincar29:
yep sorry. I didn't find a crate that would play 3d sound as easy as rodio, and I'm trying to make an audio game for blind people- (i'm also blind). audio games are based on 3d sound and I want to be concentrated on the game logic, not on the millions of lines of code with e.g. hrtf library. Rodio's spatial sink is perfect for that one, because it's easy to use. but buggy.
I know that i was a bit rood maybe...

denizsincar29:
Hello! I updated rodio to the latest version but this bug doesn't seam to be fixed. Is there a way to workarround?

PR #455, opened 2022-11-22
Force `queue` to play sources on frame boundaries

### LINKED ISSUE #444: Audio channels are sometimes reversed depending on debug and release builds
To reproduce:

Run `cargo run --example spatial`. Notice the sound starting from the right.

Run `cargo run --example spatial --release`. Notice the sound starting from the left.

Related to this comment:
- https://github.com/bevyengine/bevy/pull/6028#issuecomment-1253328622

dis-da-moe:
# Purpose
- Fixes #444 
- Fixes https://github.com/bevyengine/bevy/issues/6122

# Problem
As I understand it, this is the issue:

- Previously, when a queue has `keep_alive_if_empty` set to true, and it becomes empty, then it will push a silence lasting 10ms onto the queue.
- This is an issue because `current_frame_len` would have returned the worst case, `512`, and the silence lasts less than that.
- This means that unless the source is added immediately to the queue, and so a silence is never played, then the first actual source could start playing at a frame that is not aligned to its channels, or play at the wrong sample rate.
- This is only determined by when the source is added to the queue after its initialization. This explains why the issue was inconsistent, as it relied on the speed of execution of code which is basically random.

# Solution
- Change the functionality of `Zero` to add a method to create a silence with a certain number of frames.
- Replace the 10ms silence with a silence the length of `THRESHOLD`
- Change queue's `current_frame_len` to return `THRESHOLD` if a silence will be played next.

# Other Changes
At first, I could only reproduce this issue in the bevy game engine where the time a sink is created to when a source is appended can be long. The `spatial` example was the only example where stereo sound mattered, and so the issue was actually audible. However it had a bug where the right ear was at `-1` and the left was at `1`, so that combined with the inconsistency of the bug made it hard to reproduce.

To hopefully prevent stereo bugs in the future, I fixed the `spatial` example and made it clearer what was happening, and added a `stereo`example where the file is played in the right ear and then in the left.


harudagondi:
`stereo` and `spatial` examples worked for me in debug and release builds.

Would it be fine to include @rparrett's sound file here?

est31 (changes requested):
a .wav file of 4.61 MB is way too big of an addition. Check the other files in the directory, they are all less than 1 MB. Is that example needed for something? I don't think we should add it. Otherwise the PR is fine.

dis-da-moe:
Forget to compress. It's needed and only 200kb now.

est31:
What's the license of the file? Can you maybe add it to README.md in the assets dir and explain the license?

dis-da-moe:
I'd need @rparrett to confirm but I assume it would be the same as any other contribution.

est31 (approved):

est31:
So it's MIT OR Apache then. Fine.

DaforLynx:
Possibly fixes https://github.com/RustAudio/rodio/issues/369 too?

dis-da-moe:
Possibly, could you test it and see? Also @est31 do you know when this will be included in a release?

est31:
@dis-da-moe I don't know, i think I want to do a cpal release first, then make a rodio release.

DaforLynx:
@dis-da-moe Testing using what I believe is master, the issue doesn't seem to be fixed. The beginning and end of the beep (pure sine wave) are weird.

https://user-images.githubusercontent.com/13303368/204057398-f0e4f67d-8924-469d-80bf-112cc54b4ac5.mp4

dis-da-moe:
That seems like a separate issue, I'm not sure what causes it.

keigen-shu:
@dis-da-moe I had encountered the issue that this PR proposes to fix before, in #286.

Have you considered what happens if the output channel count is not a factor of 512, like 11?

Prior to this PR, the silence is hard-coded as a single channel 44100Hz sample at 10ms. That is a 441 samples long buffer, which does not make a stereo frame boundary.

I had worked around the issue by simply hard-coding queue.rs, sink.rs and empty.rs to 2, since I only needed it to work with stereo output. Changing the sampling rate, or frame count, would've also worked; as long as it generates a buffer that is a multiple of the output stream channel count.

dis-da-moe:
I haven't considered it since I've never encountered a case where the channel count is not a factor of 512. I assumed that's why 512 was chosen as the length of `THRESHOLD` before hand. If you are able to test the issue and find a solution to it then feel free to.

Issue #462, opened 2023-01-01
[Question] Is there any good way to restart the sink?


myyrakle:
![image](https://user-images.githubusercontent.com/16988115/210163990-bc7f8ff6-038e-43c6-94a0-d7473b83190a.png)

As above, I want to remove the existing sink playback and play the new audio.

But it seems to me that only the Sink::stop method stops existing playback. However, if it is stopped, it cannot be restarted.

Do you have any recommendations for how to do this?

Issue #315, opened 2020-09-07
When Sink is stopped it can no longer append sources, is this intended?


Stormsabre:
As the title says, when i create a sink, if i stop it and then append a new source it will not play. Would have to create another sink altogether to play music again. Is this by design? Pausing and resuming works fine though. Tried on 0.11 and master.

PolyMeilex:
According to: [implementation](https://github.com/RustAudio/rodio/blob/73857e3d51e38d9f92dfe9fe909b7c1b6693d8ae/src/sink.rs#L34) it is impossible to unstop a playback, no idea if this is intentional but it's quite annoying

LiquidityC:
The sources seem to append if the sink is stopped but they are appended as 'stopped'. I can however find no way of unstopping the stink or it's queue. In my mind using `stop()` should clear the queue and thereby abort output. Calling `play()` should unstop the the sink and allow further appends to the queue to play.

But perhaps that's just my brain being used to how old CD players used to work. Either way, not being able to stop and start playing new sources is a big hurdle for my project.

PR #464, opened 2023-01-04
Fix Sink playback after stop

RottenFishbone:
Addressing issues #462 and #315 and #171

Given how the sink's queue is drained on stop I made some modifications in `sink.append()`. I added a check to ensure the sink was flushed completely (not sure this is necessary? I think it is, to avoid race conditions) and to restart the sink on append (when stopped.)

I added a test demonstrating correctness, as well as performing some trials on my own.

However, in #171 there is mention that the stopped sink being dead is intended behavior. It would appear the general sentiment is that this should not be the case, and as a new user of rodio I can say this took me by surprise as well. 
In the case that it should kill a sink, I think it prudent to amend the documentation to reflect this, instead.

I think this provides a more correct solution than #449 and does not require any workaround for use.

### DIFF
--- a/src/sink.rs
+++ b/src/sink.rs
@@ -206,9 +206,9 @@ impl Drop for Sink {
 
 #[cfg(test)]
 mod tests {
-    use std::sync::atomic::Ordering;
     use crate::buffer::SamplesBuffer;
     use crate::{Sink, Source};
+    use std::sync::atomic::Ordering;
 
     #[test]
     fn test_pause_and_stop() {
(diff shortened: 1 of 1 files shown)


pixelafk:
I agree completely that this approach is probably better and that this PR should be more favorable.
If it is indeed meant to kill a sink completely I would have loved to know that from documentation as well.

est31 (approved):

est31:
There is popular demand for such a feature so I think the descision voiced in #171 should be revisited. This might have some implications that are not great, but until those are found we can support playback after stopping.

I think the original motivation was to make sure that when `next()` returns a `None`, it will never ever resume. Given that the iterator docs say [this](https://doc.rust-lang.org/1.65.0/core/iter/trait.Iterator.html#tymethod.next) however, we are fully allowed to resume playback after the end:

> Individual iterator implementations may choose to resume iteration, and so calling next() again may or may not eventually start returning Some(Item) again at some point.

PR #467, opened 2023-01-11
Add dynamic mixer example

nbennett320:
This example plays a C Major 6th chord by adding four sources, sine waves frequencies of C, E, G, and A, to a `DynamicMixerController`, and appends its respective `DynamicMixer` to a sink.

### DIFF
--- a/examples/mix_multiple_sources.rs
+++ b/examples/mix_multiple_sources.rs
@@ -1,0 +1,28 @@
+use std::time::Duration;
+use rodio::{OutputStream, Sink, dynamic_mixer};
+use rodio::source::{SineWave, Source};
+
+fn main() {
+  // Construct a dynamic controller and mixer, stream_handle, and sink.
+  let (controller, mixer) = dynamic_mixer::mixer::<f32>(2, 44_100);
+  let (_stream, stream_handle) = OutputStream::try_default().unwrap();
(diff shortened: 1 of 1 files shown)


examples/mix_multiple_sources.rs
```diff
@@ -0,0 +1,36 @@
...
+use rodio::{dynamic_mixer, OutputStream, Sink};
+use std::time::Duration;
+
+fn main() {
+    // Construct a dynamic controller and mixer, stream_handle, and sink.
+    let (controller, mixer) = dynamic_mixer::mixer::<f32>(2, 44_100);
+    let (_stream, stream_handle) = OutputStream::try_default().unwrap();
+    let sink = Sink::try_new(&stream_handle).unwrap();
+
+    // Create four unique sources.
```
est31, line 11: Can you maybe add a comment here explaining that these are the frequencies corresponding to the the octave corresponding to middle C? As it is an example, we should be extra clear here.

est31 (changes requested):

examples/mix_multiple_sources.rs
nbennett320: Sure, will do.

est31 (approved):
Thanks!

PR #470, opened 2023-01-30
Update to cpal 0.15

Jasper-Bekkers:
This just bumps the version number and doesn't yet address any breaking changes that may have occurred.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ exclude = ["assets/**", "tests/**"]
 edition = "2018"
 
 [dependencies]
-cpal = "0.14"
+cpal = "0.15"
 claxon = { version = "0.4.2", optional = true }
 hound = { version = "3.3.1", optional = true }
 lewton = { version = "0.10", optional = true }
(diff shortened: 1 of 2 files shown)


est31:
@Jasper-Bekkers do you want to work on this PR to perform the adjustments needed due to the breakage? If so, that would be great.

Jasper-Bekkers:
> @Jasper-Bekkers do you want to work on this PR to perform the adjustments needed due to the breakage? If so, that would be great.

I gave it an attempt, however after playing around with this a bit it seems that `cpal` has switched over it's `Sample` trait into `dasp_sample::Sample` which looks and behaves a bit different from the original `Sample`. Leading me to fight quite a few template errors throughout the project, which is a little bit more involved then what I as a casual contributor might be able to do.

est31:
Yeah it's a non trivial change. For migration, see [this](https://github.com/RustAudio/cpal/issues/747#issuecomment-1407828620) advice of looking into what was done to the examples directory. It's a non trivial effort. But it pays out, more sample formats are supported now than were previously.

Jasper-Bekkers:
Unfortunately this isn't a change I can commit time to at the moment.

est31:
@Jasper-Bekkers fair! Thanks for your answers. I will close this PR as it can't be merged in the current form. I will work on an upgrade for rodio in the coming weeks, but if someone cuts me to it, I'd be fine with that.

PR #478, opened 2023-02-17
update cpal to 0.15

mockersf:
Bump the version, and use the new `FromSample` trait for all conversions.

All the examples continue to work 🎉


est31 (approved):
Thanks!

PR #481, opened 2023-02-19
Calculate duration for mp3s

kuviman:
Currently, the duration is not calculated for mp3s when using symphonia

### DIFF
--- a/src/decoder/read_seek_source.rs
+++ b/src/decoder/read_seek_source.rs
@@ -1,11 +1,12 @@
 use std::io::{Read, Result, Seek, SeekFrom};
 use std::marker::Sync;
+use std::sync::{Mutex, RwLock};
 
 use symphonia::core::io::MediaSource;
 
 pub struct ReadSeekSource<T: Read + Seek + Send + Sync> {
-    inner: T,
-    byte_len: Option<u64>,
+    inner: Mutex<T>,
+    byte_len: RwLock<Option<Option<u64>>>, // One option is for lazy calculation
 }
 
 // Copied from std Seek::stream_len since its unstable
@@ -25,9 +26,11 @@ fn stream_len(stream: &mut impl Seek) -> std::io::Result<u64> {
 impl<T: Read + Seek + Send + Sync> ReadSeekSource<T> {
     /// Instantiates a new `ReadSeekSource<T>` by taking ownership and wrapping the provided
     /// `Read + Seek`er.
-    pub fn new(mut inner: T) -> Self {
-        let byte_len = stream_len(&mut inner).ok();
-        ReadSeekSource { inner, byte_len }
+    pub fn new(inner: T) -> Self {
+        ReadSeekSource {
+            inner: Mutex::new(inner),
+            byte_len: RwLock::new(None),
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/decoder/read_seek_source.rs
```diff
@@ -5,13 +5,29 @@ use symphonia::core::io::MediaSource;
...
     inner: T,
+    byte_len: Option<u64>,
+}
+
+// Copied from std Seek::stream_len since its unstable
+fn stream_len(stream: &mut impl Seek) -> std::io::Result<u64> {
+    let old_pos = stream.stream_position()?;
+    let len = stream.seek(SeekFrom::End(0))?;
+
+    // Avoid seeking a third time when we were already at the end of the
```
est31, line 16: Why a third time?

kuviman: This is the source: https://github.com/rust-lang/rust/blob/master/library/std/src/io/mod.rs#L1841

Third time is needed to go back to initial position

kuviman: Well, i guess its only 2 seeks, so its second, not third, bug in a comment in std I guess :)

kuviman: Actually, stream_position() is also calling a seek, so it is indeed a third

kuviman: Are you asking about `MediaSource::byte_len` method? Yes, it is provided by symphonia, and this PR changes the implementation to return `Some` calculated length now instead of returning `None` as before. This is then used to estimate the duration of the track

The method says that "This may be an expensive operation"

Although in my implementation here, I cache it when creating the source instead of calculating when the method is called.
I did it because of the method requiring shared reference to self instead of a `&mut` reference.

This could instead be done with a `RefCell` so this "expensive" operation is performed only when actually needed

Would that be better?

Other ways I see is changing the API bringing a trait like `MediaSource` into rodio itself. so that implementation of this is defined by the user, but I don't think thats a good idea

est31: I like the RefCell idea tbh.

NGnius: This would also make total_duration() work for all symphonia-decoded audio, which would be really helpful for my projects. @kuviman if you want me to try make the refcell changes, I'd be happy to.

kuviman: I have updated the PR to use lazy calculation. This is done not using `RefCell`, but with a `Mutex`/`RwLock` since we need the source to by `Sync`.

This should not have any performance issues tho since `get_mut` is enough and works without locking for normal usage, and locking will only happen once when we need to calculate the stream length

ryankopf:
Will someone please merge this.

dvdsk:
First off all, sorry this took a while we are short on maintainers. Thanks for making a PR though! Part of what you did in this PR overlaps with #513 (which has been merged).

This:
```rust
use std::io::BufReader;

use rodio::Source;

fn main() {
    let (_stream, handle) = rodio::OutputStream::try_default().unwrap();
    let sink = rodio::Sink::try_new(&handle).unwrap();

    let file = std::fs::File::open("assets/music.mp3").unwrap();
    let source = rodio::Decoder::new(BufReader::new(file)).unwrap();
    dbg!(source.total_duration());
    sink.append(source);

    sink.sleep_until_end();
}
```
prints:
```
[examples/music_mp3.rs:11:5] source.total_duration() = Some(
    10.000000006s,
)
```

Which is correct for the music.mp3 test file. 

And that's good news! That means we get a duration for mp3 files without introducing the overhead of implementing bytes_len().

@kuviman if I am wrong and total_duration is not working for you on the latest master branch commit please reopen this or let me known in some other way!

Issue #486, opened 2023-03-09
When playback ends, panic!


myyrakle:
Version 0.17.1
  
In the following form, I wrote a program that stops the currently playing sound and runs the next sound.
```
sink.stop();
while !sink.empty() {}
sink.append(new_source);
sink.play();
...
// music quit
```

However, when the replaced sound is completed, the following error occurs.
```
thread 'cpal_wasapi_out' panicked at 'index out of bounds: the len is 0 but the index is 0', C:\Users\sssan\Code\Rust\rodio\src\decoder\symphonia.rs:175:22
stack backtrace:
   0: std::panicking::begin_panic_handler
             at /rustc/658fad6c5506f41c35b64fb1a22ceb0992697ff3/library\std\src\panicking.rs:575
   1: core::panicking::panic_fmt
             at /rustc/658fad6c5506f41c35b64fb1a22ceb0992697ff3/library\core\src\panicking.rs:64
   2: core::panicking::panic_bounds_check
             at /rustc/658fad6c5506f41c35b64fb1a22ceb0992697ff3/library\core\src\panicking.rs:159
   3: rodio::decoder::symphonia::impl$2::next
             at C:\Users\sssan\Code\Rust\rodio\src\decoder\symphonia.rs:175
   4: rodio::decoder::impl$4::next<std::io::buffered::bufreader::BufReader<std::fs::File> >
             at C:\Users\sssan\Code\Rust\rodio\src\decoder\mod.rs:274
   5: rodio::source::speed::impl$1::next<rodio::decoder::Decoder<std::io::buffered::bufreader::BufReader<std::fs::File> > >
             at C:\Users\sssan\Code\Rust\rodio\src\source\speed.rs:56
   6: rodio::source::pausable::impl$1::next<rodio::source::speed::Speed<rodio::decoder::Decoder<std::io::buffered::bufreader::BufReader<std::fs::File> > > >
             at C:\Users\sssan\Code\Rust\rodio\src\source\pausable.rs:85
   7: rodio::source::amplify::impl$1::next<rodio::source::pausable::Pausable<rodio::source::speed::Speed<rodio::decoder::Decoder<std::io::buffered::bufreader::BufReader<std::fs::File> > > > >
             at C:\Users\sssan\Code\Rust\rodio\src\source\amplify.rs:56
[...]

PR #487, opened 2023-03-09
Resolving panic case 

myyrakle:
Exception handling for panic that can occur when using sink.

Related issues:
https://github.com/RustAudio/rodio/issues/486

It seems that the issue will be resolved after applying the modification.

### DIFF
--- a/src/decoder/symphonia.rs
+++ b/src/decoder/symphonia.rs
@@ -172,7 +172,7 @@ impl Iterator for SymphoniaDecoder {
             self.current_frame_offset = 0;
         }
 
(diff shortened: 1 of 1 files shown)


est31 (approved):
^

est31:
You need to rebase the PR. Easiest in this instance would probably to just do a hard reset to the upstream fork and then re-do the patch.

timdubbins:
@myyrakle I had an issue similar to #486 although only when playback completed on a particular file. This fix worked for me, thanks for providing it

est31 (approved):

PR #502, opened 2023-06-01
Sink: sleep on pause and add `.toggle()`

hinto-janai:
This makes `pause` a `Mutex<bool>` + `Condvar` combo instead of an `AtomicBool` so that the thread can sleep when paused instead of churning every 5 milliseconds (...!)
https://github.com/RustAudio/rodio/blob/9674ea56cfbdef90703833b3c6af7790d32be76e/src/sink.rs#L92

This reduces the idle CPU usage by half, as described in https://github.com/RustAudio/cpal/issues/785. The rest is in `cpal` it seems.

`.toggle()` was also added.

### DIFF
--- a/src/sink.rs
+++ b/src/sink.rs
@@ -1,5 +1,5 @@
 use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
-use std::sync::{Arc, Mutex};
+use std::sync::{Arc, Condvar, Mutex};
 use std::time::Duration;
 
 #[cfg(feature = "crossbeam-channel")]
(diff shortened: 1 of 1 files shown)


AnthonyMichaelTDM:
this causes tests to hang (seemingly) indefinitely

dvdsk:
@hinto-janaiI assume you no longer use rodio and thus (understandably) lost interest in this PR. Still thank you for putting this together. I am really interested in fixing the high idle CPU usage you point out and Ill build upon this PR in the future so your effort has not been in vane!

(right now I'm trying to work through the issue backlog and easily mergable PR's).

Issue #456, opened 2022-11-25
minimp3 uses an unsafe dependency


ImUrX:
It uses slice-deque which has [RUSTSEC-2021-0047](https://rustsec.org/advisories/RUSTSEC-2021-0047) 
I saw that germangb/minimp3-rs#38 exists, it could be used instead of the current minimp3-rs.

est31:
after #453 we now default to symphonia. I don't want to remove minimp3 completely as it still has its use cases, and I don't see alternatives that fulfill those use cases. Furthermore, it seems to me that the CVE is a hypothetical, that is, it requires a function to panic that is supplied by users. In fact, it seems that the `drain_filter` function isn't even used by minimp3 at all, so this is a false positive by tooling. Last, this is not a rodio issue. I'm closing in favour of https://github.com/germangb/minimp3-rs/issues/29 and most importantly https://github.com/gnzlbg/slice_deque/issues/90 .

PR #504, opened 2023-07-25
Replaces minimp3 with minimp3_fixed witch fixes a vital security flaw

BOB450:
https://rustsec.org/packages/slice-deque.html
This new minimp3 fork replaces this vulnerable package.


est31 (approved):
cc also #456 . Thanks for the reminder!

PR #506, opened 2023-08-01
Add a `SourcesIdQueueInput` that allows manipulating items in the queue

nicklan:
This PR abstracts the functionality the a `SourcesQueueOutput` needs from its upstream input queue, and then adds a new type of queue that can have ids associated with the items.

The idea is that you can use a `SourcesIdQueueInput` more like a playlist, which is something consumers of this library might well want.

This is a "first draft", it works, but I'm open to suggestions for api changes etc, and I know some of the doc comments are missing and/or need to be improved. Just wanted to get feedback on whether this is something you would consider before going further.

### DIFF
--- a/src/queue.rs
+++ b/src/queue.rs
@@ -50,10 +50,12 @@ where
     (input, output)
 }
 
-pub fn id_queue<S, I>(keep_alive_if_empty: bool) -> (Arc<SourcesIdQueueInput<S, I>>, SourcesQueueOutput<S>)
+pub fn id_queue<S, I>(
+    keep_alive_if_empty: bool,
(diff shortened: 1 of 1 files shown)


UnknownSuperficialNight:
I was going to write an issue about this idea to see if I should implement it, but to my surprise, someone is already on it. Legend

yara-blue:
When this PR was opened Rodio was struggling for maintainers. That's fixed now but we could not pick this up in time. 

I'm going to close this assuming the author has moved on since. Further more it is really hard to add such a feature correctly in the current Rodio. We are about to overhaul the audio engine (Source trait). It will be far easier in a few months. At that point we should definitely add something like this!

PR #507, opened 2023-08-15
Add resample example

wiccy46:
The documentation is not obvious regarding how to resample an input.


dvdsk:
The `sink` will resample input for you when you append see: https://github.com/RustAudio/rodio/blob/818f557466098942e7b928ca6c7d696bc786207c/src/sink.rs#L150

This example is therefore resampling twice (once using `UniformSourceIterator` explicit and once implicit in sink). That does not make sense right?

A PR clarifying the docs for Sink::append might be a better way to address this. Cause I do agree it is very confusing and not obvious how input is resampled! Please, if you are willing, help improve the docs, we could really use the help.

dvdsk:
API has changed, PR not mergable

Issue #443, opened 2022-09-18
skip_duration is unusably slow


ivakam:
Calling `skip_duration` on a source seems to block the thread for a completely unacceptable amount of time (several seconds, often double-digit). My guess is that this is due to how skipping is implemented where the source is manually stepped forward frame-by-frame until the offset is reached. Is is possible that the source could instead be indexed to reach the correct offset directly? I feel this should be possible, at least with sources that are constant in sample rate (or rather, sources that already use `do_skip_duration_unchecked`).

dev-callum-hopkins:
I suspect that because the skip_duration effect sits on top of what is essentially an iterator of audio samples it cannot directly jump forwards to where it needs to be...

est31:
It should be possible to implement fast duration skipping at least with lewton: decoding is a choice for the users.

dag-h:
My solution to this was making my own `Source` impl that used `seek` on the Symphonia reader to skip forward, which was pretty much instantaneous.

Issue #474, opened 2023-02-11
Get current sink track playback time


prajnastra:
Is there any way to get current sink track playback time? I have searched in the docs but haven't found anything yet.

mario-hess:
I have the same issue.

pixmusix:
Hi! When trying to solve this for myself I found this test in the buffer.rs file.

```rust
    #[test]
    fn duration_basic() {
        let buf = SamplesBuffer::new(2, 2, vec![0i16, 0, 0, 0, 0, 0]);
        let dur = buf.total_duration().unwrap();
        assert_eq!(dur.as_secs(), 1);
        assert_eq!(dur.subsec_nanos(), 500_000_000);
    }
```

I can imagine some function in a sink that calls buf.total_duration() of the currently playing song.
Reading through the sink.rs file I can see that there is a datatype queue_tx that looks promising.
If the repository authors are comfortable with it I'd be happy to try implementing it.

```rust
    #[inline]
    pub fn get_duration(&self) -> u64 {
        // Somehow get first element of queue_tx as buffer
        // Return duration of that buffer
    }
```

BastianHentschel:
There a bit of a problem, because the currently played track is not in the queue_tx, as it only has the "to-be-played" sources. The currently played is hold by queue_rx, which is immediately after construction moved into the OutputStreamHandle, so you don't have very much leverage on it. 

Also Source::total_duration() is total track length, not remaining, not sure if relevant here.

The easiest way way to get the total_duration is on track load, create your own trackname->duration map and save the information from the Decoder you create by calling its total_duration() before appending it to the sink.

BastianHentschel:
If you're still interested you could look over #510 and see if that solves part of the problem

PR #510, opened 2023-09-07
Elapsed Duration functionality for the Decoder/LoopedDecoder

BastianHentschel:
The commits add a member to the Decoders, which contains the required information for the elapsed duration.

The functionality is implemented in the `impl<R> Iterator for Decoder<R>/LoopedDecoder<R>` and ticks an AtomicUsize up each read sample. With the channel count and sample rate the effective elapsed duration of a Decoder can be calculated.

### DIFF
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -70,7 +70,7 @@ where
 pub struct DecoderInfo {
     channels: u16,
     sample_rate: u32,
-    pub(crate) samples_elapsed: AtomicUsize,
+    samples_elapsed: AtomicUsize,
 }
 
 impl DecoderInfo {
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -7,9 +7,11 @@ use std::io::{Read, Seek, SeekFrom};
 use std::marker::Sync;
 use std::mem;
 use std::str::FromStr;
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::Arc;
 use std::time::Duration;
 
-use crate::Source;
+use crate::{Sample, Source};
 
 #[cfg(feature = "symphonia")]
 use self::read_seek_source::ReadSeekSource;
@@ -32,13 +34,21 @@ mod wav;
 /// Source of audio samples from decoding a file.
 ///
 /// Supports MP3, WAV, Vorbis and Flac.
-pub struct Decoder<R>(DecoderImpl<R>)
+pub struct Decoder<R>
 where
-    R: Read + Seek;
+    R: Read + Seek,
+{
+    decoder: DecoderImpl<R>,
+    info: Arc<DecoderInfo>,
+}
 
-pub struct LoopedDecoder<R>(DecoderImpl<R>)
+pub struct LoopedDecoder<R>
 where
-    R: Read + Seek;
+    R: Read + Seek,
+{
+    decoder: DecoderImpl<R>,
+    info: Arc<DecoderInfo>,
+}
 
 enum DecoderImpl<R>
 where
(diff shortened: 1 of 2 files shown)


BastianHentschel:
This could maybe be a solution to #474 

The total duration can be gotten beforehand from the decoder and be cached for later use.

src/decoder/mod.rs
```diff
@@ -54,7 +64,50 @@ where
...
+    }
+
+    #[inline]
+    pub fn elapsed_samples(&self) -> usize {
+        self.samples_elapsed.load(Ordering::Relaxed)
+    }
+
+    #[inline]
+    pub fn elapsed_duration(&self) -> Option<Duration> {
+        Duration::try_from_secs_f64(
```
dvdsk, line 106: forgive me but why do we return an `Option` here? is there a combination of elapsed_samples, sample rate and channels that makes [try_from_secs_f64](https://doc.rust-lang.org/std/time/struct.Duration.html#method.try_from_secs_f64) fail? I would love a comment explaining that.

Or is it there because some combination of sources could not support `elapsed_duration`?

dvdsk (changes requested):
Love that you are adding this! I have wanted it for a while, my current workaround is guessing the time between pausing and playing however that has a ton of issues.

I am also adding seeking support to rodio. It would be very nice if I or someone else could use that and this to fix:  #443. For that we would need the *elapsed_duration* inside: https://github.com/RustAudio/rodio/blob/eda5934a209056c39cea3a4734394d5680a2a8a3/src/source/skip.rs#L21 

Is it possible to change the API/approach to this such that the elapsed_duration could be used there? 

It could also be nice if we could access the current elapsed duration for the source currently playing in a `Sink`.

Again thanks for contributing!

src/decoder/mod.rs
```diff
@@ -54,7 +64,50 @@ where
...
+        self.sample_rate
+    }
+
+    #[inline]
+    pub fn elapsed_samples(&self) -> usize {
+        self.samples_elapsed.load(Ordering::Relaxed)
+    }
+
+    #[inline]
+    pub fn elapsed_duration(&self) -> Option<Duration> {
```
dvdsk, line 105: This might not be the best name, elapsed sounds like its the time since the song started playing, however it is the position of the song right? So maybe name it .... position, playback_pos or something else

Jedsek:
sorry, but ping?

dvdsk:
This might fit better as a new method on `Source` with a default implementation returning `None` (just like `try_seek`). Will be a little more work since now all the adapters need to be adjusted. Though one can easily follow what `try_seek` does for most cases.

The bigger question is what to do about `Sink`. Since that is what most media player like uses for _rodio_ will use.

geeseofbeverlyroad:
Really nice! Am using rodio for a small music app, and would be very helpful to have this feature.

dvdsk:
Note for myself in the future:

Implement this through adding a new source wrapping whatever is beneath. 
 - The new source keeps track of the position (number of samples * sample rate for each frame). 
 - Will work universally on any source. 
 - The new source tracking position will also get a try_seek implementation that updates the position on success. 
 - Using a relaxed atomic might enable easy access. Something like `get_pos() -> &AtomicU32` (u32 fits 49.71027 days of milliseconds, is that enough and precise enough?).

Anyone should feel free to try something like this, I do not have the time atm. I do need this feature for a hobby project so I will implement it eventually, tho that could be in a few months.

geeseofbeverlyroad:
@dvdsk Ah, should have read more carefully, I assumed the current state of the PR is mergeable, since the most recent comment was just a few weeks old.

I could try to make a new PR based on the design you just outlined, but am afraid the architectural design decisions are a bit out of my comfort zone as I'm quite unfamiliar with the library. 

With a slightly more detailed set of instrucitons can definitely attempt to do this.

BastianHentschel:
I haven't looked at this in a while, but I can also later this week take a look at the requirements and see how the features could be implemented. 
I also only did this because I needed the feature myself and I put it here, because it seemed like somebody else could also be interested.

dvdsk:
> am afraid the architectural design decisions are a bit out of my comfort zone as I'm quite unfamiliar with the library.

I recommend taking a look at the text here: https://docs.rs/rodio/latest/rodio/source/trait.Source.html, browse a few of the structs implementing `Source` and take a look at the sink here: https://github.com/RustAudio/rodio/blob/ee73f6d97d9ffca450f497f5ac0eff266dfb9fc7/src/sink.rs#L120

That should give you a feel for the relevant part of the code base. Then you or @BastianHentschel can give it a try. Let me know if its still unclear after reading up, then ill write up some pseudo code.

Gusted:
Hi, I've gone ahead and implemented what was described in https://github.com/RustAudio/rodio/pull/510#issuecomment-2135872850 over at #585

dvdsk:
> Hi, I've gone ahead and implemented what was described in [#510 (comment)](https://github.com/RustAudio/rodio/pull/510#issuecomment-2135872850) over at #585

Ill have a look tomorrow :+1: thanks for the work!

Issue #514, opened 2023-10-05
Allow filter Q value to be changed


willxwmusic:
The functions `low_pass()` and `high_pass()`, although buggy, do not allow for the q value to be changed. I think this should be another parameter for the function which can be passed in order to change this property of the filter.

est31:
yeah that's a good idea :+1:

PR #515, opened 2023-10-06
added q value parameter to be passed to filter functions

willxwmusic:
I have added a 2nd parameter to the filter functions low_pass, high_pass, to_low_pass, and to_high pass; allowing a q value to be passed along with the frequency.

### DIFF
--- a/src/source/blt.rs
+++ b/src/source/blt.rs
@@ -36,7 +36,7 @@ where
     }
 }
 
-/// Same as low_pass but allows the q value (bandwidth) to be changed
+/// Same as high_pass but allows the q value (bandwidth) to be changed
 pub fn high_pass_with_q<I>(input: I, freq: u32, q: f32) -> BltFilter<I>
 where
     I: Source<Item = f32>,
--- a/src/source/blt.rs
+++ b/src/source/blt.rs
@@ -6,7 +6,37 @@ use crate::Source;
 // Implemented following http://www.musicdsp.org/files/Audio-EQ-Cookbook.txt
 
 /// Internal function that builds a `BltFilter` object.
-pub fn low_pass<I>(input: I, freq: u32, q: f32) -> BltFilter<I>
+pub fn low_pass<I>(input: I, freq: u32) -> BltFilter<I>
+where
+    I: Source<Item = f32>,
+{
+    BltFilter {
+        input,
+        formula: BltFormula::LowPass { freq, q: 0.5 },
+        applier: None,
+        x_n1: 0.0,
+        x_n2: 0.0,
+        y_n1: 0.0,
+        y_n2: 0.0,
+    }
+}
+
+pub fn high_pass<I>(input: I, freq: u32) -> BltFilter<I>
+where
+    I: Source<Item = f32>,
+{
+    BltFilter {
+        input,
+        formula: BltFormula::HighPass { freq, q: 0.5 },
+        applier: None,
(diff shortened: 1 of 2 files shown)


willxwmusic:
this is intended as a solution to #514

willxwmusic:
I agree, this is a better solution.
I will update the proposed API to involve an option defaulting to 0.5, the original default q value within new methods such as `high_pass_with_q()`.
Thank you for your help.

willxwmusic:
I have added secondary methods which are the same as `low_pass`, `high_pass`, `to_low_pass` and `to_high_pass` titled `low_pass_with_q`, etc.
They allow the q value to be passed to these functions too, changing the bandwidth of said filters.

dvdsk:
> Creating a new method in order to perform exactly the same function but with an extra parameter is not an efficient solution to the issue.

I think I have got a good solution for that. Lets make the implementation of the non q versions call the `with_q` versions with as q value  `0.5`. The compiler will inline the call so this won't make the code slower.

willxwmusic:
I have implemented this correction. `low_pass()`, etc. now just call `low_pass_with_q()`,etc. and pass the cutoff frequency along with the original default Q value of 0.5. This means that previously written code will not be broken and downstream users will not have to update their code, however, the option to change the q value is now there if required.

willxwmusic:
I believe this could be better implemented using a default instead of a secondary method. I will implement this change

willxwmusic:
Upon revision. that would not be possible. I believe this should be pulled to master as the `with_q` functions work as desired.

src/source/blt.rs
```diff
@@ -21,13 +36,14 @@ where
     }
 }
 
-pub fn high_pass<I>(input: I, freq: u32) -> BltFilter<I>
+/// Same as low_pass but allows the q value (bandwidth) to be changed
```
dvdsk, line 39:
```suggestion
/// Same as high_pass but allows the q value (bandwidth) to be changed
```

src/source/mod.rs
```diff
@@ -351,6 +351,23 @@ where
     {
         blt::high_pass(self, freq)
     }
+
```
dvdsk, line 354: Think you forgot the #[inline] here

```suggestion

#[inline]
```

src/source/mod.rs
```diff
@@ -351,6 +351,23 @@ where
...
+    fn low_pass_with_q(self, freq: u32, q: f32) -> BltFilter<Self>
+    where
+        Self: Sized,
+        Self: Source<Item = f32>,
+    {
+        blt::low_pass_with_q(self, freq, q)
+    }
+
```
dvdsk, line 362: Lets add a comment explaining what q is like you have in blt.rs

dvdsk (changes requested):
Looks great, I found some small stuff but that should be easy to fix. 

After that you should ask est31 for a review. They are a maintainer and can merge this. I am only trying to help them out a bit.

src/source/mod.rs
willxwmusic: Added comments similar to that of blt.rs

willxwmusic:
My apologies for the rookie mistakes. I am fairly new to rust as a programming language. I will submit the pull request for review.

willxwmusic:
@est31 could this please be merged? Thanks.

est31 (approved):

PR #519, opened 2023-10-20
Fix build with minimp3

a1phyr:
The last publish version of rodio does not compile


a1phyr:
Could you do a new release with this please ? (And maybe yank 0.17.2)

PR #532, opened 2023-12-24
Implement Source for mut references

dexterdy:
I added an implementation of Source for mutable references, to make it easier to work with and to add compatibility with types such as Arc and Rc.

I also extracted the repeated implementation for pointers into a macro, making the source code shorter and getting rid of a little boilerplate.


est31 (approved):
Thanks!

PR #547, opened 2024-03-05
Make `DynamicMixer` and `OutputStream` generic

eira-fransham:
This commit adds a new trait, `Mixer`, and refactors some of the code around generic mixers in order to support it. This is to make more of the code reusable without rewriting - specifically, to support a feature that I would like to implement in `bevy_audio` where audio sources can send audio to an intermediate mixer. The motivation for this was to add a global limiter to all game audio, so that loud sounds did not cause clipping on the master bus, but another possible usecase would be to split in-game audio and menu audio into separate "worlds" and apply, for example, a low-pass filter while in the pause screen.

### DIFF
--- a/src/dynamic_mixer.rs
+++ b/src/dynamic_mixer.rs
@@ -7,6 +7,8 @@ use std::time::Duration;
 use crate::source::{Source, UniformSourceIterator};
 use crate::Sample;
 
+type DynSource<S> = Box<dyn Source<Item = S> + Send + Sync + 'static>;
+
 /// Builds a new mixer.
 ///
 /// You can choose the characteristics of the output thanks to this constructor. All the sounds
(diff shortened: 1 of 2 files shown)


eira-fransham:
It looks like the test failures are due to unrelated code, and are because of the addition of redundant import checks to `unused_imports`.

eira-fransham:
A more-minimal version of this PR would just make `OutputStream` generic, which would still allow reusing `rodio`'s mixer without adding an actual interface around creating your own.

hamirmahal (commented):
> It looks like the test failures are due to unrelated code, and are because of the addition of redundant import checks to `unused_imports`.

The test failures should be gone on the main branch now because of https://github.com/RustAudio/rodio/pull/554.

dvdsk:
Could you rebase this on main? (or merge main into it). Then you'll get the fixed (thanks @hamirmahal) test suite.

hamirmahal:
You're welcome, @dvdsk!

dvdsk:
I'm going to close this given we just redesigned the mixer & stream and the code has diverged a lot now. During the design we discussed this use-case and decided to not make the mixer generic for now. 

The performance and API tradeoffs where unclear and the use-cases are already possible by adding a Source to the mixer that is itself a new special mixer. 

That reduced the use-case of this PR to optimization.

Once we have a better idea how the new `dynamic` source API will look & how the new `Player` will work designing this will be easier.

Feel free to re-open but the re-base is going to be hellish and its probably better to start again.

PR #557, opened 2024-03-21
refactor: remove unnecessary `Default::default()`

hamirmahal:


src/decoder/symphonia.rs
```diff
@@ -72,13 +72,8 @@ impl SymphoniaDecoder {
...
 
-        let mut decoder = symphonia::default::get_codecs().make(
-            &stream.codec_params,
-            &DecoderOptions {
-                verify: true,
-                ..Default::default()
-            },
-        )?;
+        let mut decoder = symphonia::default::get_codecs()
+            .make(&stream.codec_params, &DecoderOptions { verify: true })?;
```
hamirmahal, line 76: `pub struct DecoderOptions {` only has `pub verify: bool,`, which this code already specifies, so `..Default::default()` isn't doing anything here.

est31 (approved):

PR #559, opened 2024-03-25
Fix channel count upscaling

### LINKED ISSUE #558: Channel converter does not scale to more then 2 channels
Playing sounds on the left channel works fine, but if I try to play something on the right channel, it plays on both (but maybe with slightly less volume on the left? hard to tell).

When running `cargo run --example stereo`, what you're *supposed* to hear is a tone that alternates right/left/right/left channels, but what I hear is both/left/both/left.

If I play the actual file (`Assets/RL.ogg`) in a different program, it sounds fine. But when playing via rodio, I get this issue.

I'm assuming this doesn't happen to everyone. Any ideas on what the issue could be? A weird quirk of my system/setup, or a rodio bug?

ideka:
This changes channel upscaling so all input samples are spread out across the output channels, instead of repeating the last sample over and over. This approach should yield more accurate results. Fixes #558.

### DIFF
--- a/src/conversions/channels.rs
+++ b/src/conversions/channels.rs
@@ -87,11 +87,14 @@ where
     fn size_hint(&self) -> (usize, Option<usize>) {
         let (min, max) = self.input.size_hint();
 
-        let min =
-            (min / self.from as usize) * self.to as usize + self.next_output_sample_pos as usize;
-        let max = max.map(|max| {
-            (max / self.from as usize) * self.to as usize + self.next_output_sample_pos as usize
-        });
+        let consumed = std::cmp::min(self.from, self.next_output_sample_pos) as usize;
+        let calculate = |size| {
+            (size + consumed) / self.from as usize * self.to as usize
+                - self.next_output_sample_pos as usize
+        };
+
+        let min = calculate(min);
+        let max = max.map(calculate);
 
         (min, max)
     }
@@ -120,14 +123,11 @@ mod test {
     }
 
     #[test]
-    fn mono_to_stereo() {
+    fn add_channels() {
         let input = vec![1i16, 2, 3, 4];
         let output = ChannelCountConverter::new(input.into_iter(), 1, 2).collect::<Vec<_>>();
         assert_eq!(output, [1, 1, 2, 2, 3, 3, 4, 4]);
-    }
 
-    #[test]
-    fn add_channels() {
         let input = vec![1i16, 2];
         let output = ChannelCountConverter::new(input.into_iter(), 1, 4).collect::<Vec<_>>();
         assert_eq!(output, [1, 1, 0, 0, 2, 2, 0, 0]);
@@ -135,6 +135,28 @@ mod test {
         let input = vec![1i16, 2, 3, 4];
         let output = ChannelCountConverter::new(input.into_iter(), 2, 4).collect::<Vec<_>>();
         assert_eq!(output, [1, 2, 0, 0, 3, 4, 0, 0]);
+    }
+
+    #[test]
+    fn size_hint() {
+        fn test(input: &[i16], from: cpal::ChannelCount, to: cpal::ChannelCount) {
+            let input = input.to_vec();
+            let scaled_len = input.len() / from as usize * to as usize;
+            let mut converter = ChannelCountConverter::new(input.into_iter(), from, to);
+            for i in 0..scaled_len {
+                assert_eq!(
+                    converter.size_hint(),
+                    (scaled_len - i, Some(scaled_len - i))
+                );
+                converter.next();
+            }
+        }
+
+        test(&[1i16, 2, 3], 1, 2);
+        test(&[1i16, 2, 3, 4], 2, 4);
+        test(&[1i16, 2, 3, 4], 4, 2);
+        test(&[1i16, 2, 3, 4, 5, 6], 3, 8);
+        test(&[1i16, 2, 3, 4, 5, 6, 7, 8], 4, 1);
     }
 
     #[test]
--- a/src/conversions/channels.rs
+++ b/src/conversions/channels.rs
@@ -21,7 +21,7 @@ where
     ///
     /// # Panic
     ///
-    /// Panicks if `from` or `to` are equal to 0.
+    /// Panics if `from` or `to` are equal to 0.
     ///
     #[inline]
     pub fn new(
@@ -71,7 +71,7 @@ where
         self.next_output_sample_pos += 1;
 
         if self.next_output_sample_pos == self.to {
-            self.next_output_sample_pos -= self.to;
+            self.next_output_sample_pos = 0;
 
             if self.from > self.to {
                 for _ in self.to..self.from {
--- a/src/conversions/channels.rs
+++ b/src/conversions/channels.rs
@@ -7,7 +7,7 @@ where
     input: I,
     from: cpal::ChannelCount,
     to: cpal::ChannelCount,
-    sample_repeat: Option<I::Item>,
+    sample_repeat: Vec<Option<I::Item>>,
     next_output_sample_pos: cpal::ChannelCount,
 }
 
@@ -34,7 +34,11 @@ where
             input,
             from,
             to,
-            sample_repeat: None,
+            sample_repeat: {
+                let mut vec = Vec::with_capacity(from as usize);
+                vec.resize_with(from as usize, || None);
+                vec
+            },
             next_output_sample_pos: 0,
         }
     }
@@ -54,14 +58,12 @@ where
     type Item = I::Item;
 
     fn next(&mut self) -> Option<I::Item> {
-        let result = if self.next_output_sample_pos == self.from - 1 {
+        let result = if self.next_output_sample_pos < self.from {
             let value = self.input.next();
-            self.sample_repeat = value.clone();
+            self.sample_repeat[self.next_output_sample_pos as usize] = value.clone();
             value
-        } else if self.next_output_sample_pos < self.from {
-            self.input.next()
         } else {
-            self.sample_repeat.clone()
+            self.sample_repeat[(self.next_output_sample_pos % self.from) as usize].clone()
         };
 
         self.next_output_sample_pos += 1;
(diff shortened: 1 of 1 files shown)


dvdsk:
It seems like there is no standard order for 5.1 and 7.1 surround sound instead it differs across file formats and can even be configurable. For example see channel assignment for [flac](https://xiph.org/flac/format.html). That is gonna make correct channel conversion quite challenging. As it might need work on the decoders to unify how they provide channels. Symphonia does at least provide a consistent order it seems: [symphonia docs](https://docs.rs/symphonia-core/0.5.4/symphonia_core/audio/struct.Channels.html).

It makes perfect sense to turn mono to stereo by assigning both speakers the same sample. Scaling stereo to 5.1 and 7.1 should probably be done by recognizing which of the output channels is the left and which the right channel and sending the rest zero's/nothing.

Now **back to this PR** and stereo to 5.1/7.1 up-scaling :)
From this [cpal PR](https://github.com/RustAudio/cpal/pull/198/files) it seems cpal uses the WAVEFORMATEXTENSIBLE order (see [winapi docs](https://learn.microsoft.com/en-us/windows/win32/api/mmreg/ns-mmreg-waveformatextensible)) just like symphonia does. That means you should be able to figure out which is the front left and front right channel for any device. Then we can have rodio send stereo to only those channels.

Does that sound good to you?

ideka:
So it can be safely assumed that the first 18 channels will always be in the WAVEFORMATEXTENSIBLE order? Like in [this table here](https://learn.microsoft.com/en-us/windows/win32/api/mmreg/ns-mmreg-waveformatextensible)?

So the way it should be is...

* 2 to 4: [1, 2, 3, 4] -> [1, 2, 0, 0, 3, 4, 0, 0]
* 2 to 6: [1, 2, 3, 4] -> [1, 2, 0, 0, 1, 2, 3, 4, 0, 0, 3, 4]
* 2 to 8: [1, 2, 3, 4] -> [1, 2, 0, 0, 1, 2, 1, 2, 3, 4, 0, 0, 3, 4, 3, 4]
* 4 to 6: [1, 2, 3, 4, 5, 6, 7, 8] -> [1, 2, 3, 4, 1, 2, 5, 6, 7, 8, 5, 6]
* 4 to 8: [1, 2, 3, 4, 5, 6, 7, 8] -> [1, 2, 3, 4, 1, 2, 1, 2, 5, 6, 7, 8, 5, 6, 5, 6]
* 6 to 8: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12] -> [1, 2, 3, 4, 5, 6, 1, 2, 7, 8, 9, 10, 11, 12, 1, 2]

For channels from 9 to 18 it'd get more complicated because they don't alternate left/right anymore, but the idea would be the same. Match left with left, right with right, center with center.

This kinda begs the question though. As you said, when going from mono to stereo it makes sense to duplicate the sample, but should that be done as well when going higher? maybe for 2 to 8 for example, this makes more sense? [1, 2, 0, 0, 0, 0, 0, 0, 3, 4, 0, 0, 0, 0, 0, 0]

Let me know what you think.

dvdsk:
> So it can be safely assumed that the first 18 channels will always be in the WAVEFORMATEXTENSIBLE order? Like in [this table here](https://learn.microsoft.com/en-us/windows/win32/api/mmreg/ns-mmreg-waveformatextensible)?

As far as I can deduce from the cpal source thats correct. I would say anything beyond 8 channels it too nice for rodio right now (though that's not my call to make).

> Match left with left, right with right, center with center.

That would ensure audio to most speaker. Though that is not what (a quick google shows) is [recommended](https://hometheateracademy.com/listen-to-music-in-stereo-or-surround-sound). The tldr being:

> While you can attempt to stream music in surround sound, this will often just make the audio louder without improving clarity, and often results in sound distortion.

I think we can safely generalize that to: do not use more channels then the source is offers as that will decrease clarity. With an exception for going from mono to stereo (it is quite annoying to have sound only left or right).

> This kinda begs the question though. As you said, when going from mono to stereo it makes sense to duplicate the sample, but should that be done as well when going higher? maybe for 2 to 8 for example, this makes more sense? [1, 2, 0, 0, 0, 0, 0, 0, 3, 4, 0, 0, 0, 0, 0, 0]

So yes given the source above to me that makes more sense. Just for clarity that would correspond to this table:

| 7.1 speaker position | interleaved stereo sample numb | non interleaved stereo sample |
|------------------|----------------------------------------------| -----------------------------------------------|
| SPEAKER_FRONT_LEFT | 1 | sample 1 left channel |
| SPEAKER_FRONT_RIGHT | 2 | sample 1 right channel |
| SPEAKER_FRONT_CENTER |0 | silent |
| SPEAKER_LOW_FREQUENCY | 0| silent |
| SPEAKER_BACK_LEFT | 0 | silent |
| SPEAKER_BACK_RIGHT |0 | silent |
| SPEAKER_FRONT_LEFT_OF_CENTER |0 | silent |
| SPEAKER_FRONT_RIGHT_OF_CENTER | 0 | silent |
| SPEAKER_FRONT_LEFT | 3 | sample 2 left channel |
| SPEAKER_FRONT_RIGHT | 4 | sample 2 right channel |
| SPEAKER_FRONT_CENTER |0 | silent |
| SPEAKER_LOW_FREQUENCY | 0| silent |
| SPEAKER_BACK_LEFT | 0 | silent |
| SPEAKER_BACK_RIGHT |0 | silent |
| SPEAKER_FRONT_LEFT_OF_CENTER |0 | silent |
| SPEAKER_FRONT_RIGHT_OF_CENTER | 0 | silent |

src/conversions/channels.rs
dvdsk, line 1: I think `size_hint()` isnt correct anymore now right? The calculation should change. For the rest it looks great!

dvdsk:
oh and while your at would you mind doing a tiny refactor :)

At line 74 (right now) I think this:
```
if self.next_output_sample_pos == self.to {
    self.next_output_sample_pos -= self.to;
```
would be clearer like this
```
if self.next_output_sample_pos == self.to {
    self.next_output_sample_pos = 0;
```
and still work right? Would you agree?

src/conversions/channels.rs
```diff
@@ -106,36 +113,62 @@ mod test {
...
 
-        let input = vec![1u16, 2, 1, 2];
+        let input = vec![1i16, 2, 3, 4];
         let output = ChannelCountConverter::new(input.into_iter(), 2, 4).collect::<Vec<_>>();
-        assert_eq!(output, [1, 2, 2, 2, 1, 2, 2, 2]);
+        assert_eq!(output, [1, 2, 0, 0, 3, 4, 0, 0]);
+    }
+
+    #[test]
+    fn size_hint() {
```
dvdsk, line 141: this test seems quite complex and coupled to the implementation. How about something along these lines:

```
fn test(iter, from, to) {
    size_hint = iter.size_hint();
    actual size = iter.count() // from Iterator trait
    assert( is actual size contained in size_hint? )
}
```

If we wanted (might be overkill) we could define a custom Test iterator with a custom uncertain size_hint (I think the min and max size hints are the same for Vec since the len is known).

ideka: You have a point, but the thing is the tricky/most fallible part of the `size_hint` calculation is when some of the items have already been consumed, so that's the part that is most valuable to test I think.

What about something like this?

```rust
fn test(input: &[i16], from: cpal::ChannelCount, to: cpal::ChannelCount) {
    let count = ChannelCountConverter::new(input.iter().copied(), from, to).count();
    let mut converter = ChannelCountConverter::new(input.iter().copied(), from, to);
    for i in 0..count {
        assert_eq!(converter.size_hint(), (count - i, Some(count - i)));
        converter.next();
    }
}
```

But I am okay with your version if you think this is still overkill.

dvdsk: You can make the test function generic and clone the iterator: `fn test<T>(iter: impl Iterator<Item=T> + Clone, from, to)`. Then you can pass in iterators in various states. 

The overkill option would be passing in a custom iterator specifically made for this test. That would look something like this:

```rust
struct TestIter {
  bounds_to_report: (usize, Option<usize>),
  length: usize,
  consumed: usize,
}

impl Iterator for TestIter {
  Item = ();
  fn next(&mut self) -> Self::item {
      if self.consumed >= self.length {
         None
      } else 
         Some(())
      }
  }

  fn size_hint(&self) -> (usize, Option<usize) {
    self.bounds_to_report
  }
}
```

This would allow us to test the behavior when the upper and lower bound are not the same. But I have looked around and no one implementing Iterator does something like this :sweat_smile:, So lets not.

ideka: I don't see the appeal of the `+ Clone` version. It seems like it'd require a lot of calls to be very useful...

I think these two options are sound:

```rust
fn test_1(input: &[i16], from: cpal::ChannelCount, to: cpal::ChannelCount) {
    let converter = ChannelCountConverter::new(input.iter().copied(), from, to);
    let size_hint = converter.size_hint();
    let count = converter.count();
    assert_eq!(size_hint, (count, Some(count)));
}

fn test_2(input: &[i16], from: cpal::ChannelCount, to: cpal::ChannelCount) {
    let mut converter = ChannelCountConverter::new(input.iter().copied(), from, to);
    let count = converter.clone().count();
    for i in 0..count {
        assert_eq!(converter.size_hint(), (count - i, Some(count - i)));
        converter.next();
    }
}
```

Let me know which one you think is best and I'll commit and push the changes.

dvdsk: > I don't see the appeal of the `+ Clone` version. It seems like it'd require a lot of calls to be very useful...

oh seems like that was just me being silly there. I had the count before the size_hint (and thus consuming the iterator) which I fixed using clone. Swapping the order like you do is obv superior :+1: 

The idea of passing in an iterator instead of a slice is that you can pass in partially consumed iterators. You could pass in a `[1,2,3,4].into_iter().skip(3)`. Though the loop works too!

The whole `i in 0..count` and then `count - i`, do you agree with replacing that with `for left_in_iter in (0..count).rev()` and `(left_in_iter, Some(left_in_iter))`?

ideka: Oh yeah that is better. It'll have to be `(0..=count).rev()` though.
I'm also fixing a bug that caused the size_hint to go up again if calling `.next()` after the iterator was exhausted, and adding one final `assert_eq!` to test for it.

xMAC94x:
thanks for seeing a fix here :) Greetings from the veloren team

dvdsk:
Since this is a non trivial PR and touches an essential part of rodio I've asked @est32 to take a look. As soon as thats done we are ready to merge! :tada: 

Thanks a lot for adding proper multi channel support @ideka to rodio! As you've seen its already appreciated by the community!

est31 (approved):

PR #560, opened 2024-03-25
refactor: remove unnecessary `Box` from into_inner

hamirmahal:


src/decoder/symphonia.rs
```diff
@@ -48,7 +48,7 @@ impl SymphoniaDecoder {
         }
     }
 
-    pub fn into_inner(self: Box<Self>) -> MediaSourceStream {
```
hamirmahal, line 51: `self` doesn't need to be boxed here.

dvdsk:
`mod symphonia` is private to rodio, so this change is possible without changing rodio's API. 

Could you also remove the Box in the caller?

hamirmahal:
Sure! Is 371d963 what you're looking for?

dvdsk:
perfect!

Issue #562, opened 2024-03-26
Rename Sink::empty


dvdsk:
It should probably be called `is_empty`. Clippy even flags on it.

Changing it would break a lot of code though...
Could deprecate it first?

Let me hear you thoughts, for now I am letting clippy now to ignore it.

dvdsk:
same is true for SpatialSink::empty

daniellga:
Any improvements on this?

dvdsk:
not really, I am saving up some breaking changes for a big release. I rather not break ppls code multiple times. So it seems logical to save it up and provide a porting guide.

dvdsk:
though it might make sense to start collecting the changes on a new branch

PR #563, opened 2024-03-26
This fixes all clippy warnings

dvdsk:
Made an issue for missing `is_empty()`, see:
https://github.com/RustAudio/rodio/issues/562

The DecoderImpl enum triggerd large_enum_variant. There is not an easy fix for that. We could box the VorbisDecoder but at 572 bytes I really do not think thats worth it. Annotated it with an allow

### DIFF
--- a/src/decoder/flac.rs
+++ b/src/decoder/flac.rs
@@ -107,7 +107,7 @@ where
 
             // Load the next block.
             self.current_block_off = 0;
-            let buffer = mem::replace(&mut self.current_block, Vec::new());
+            let buffer = mem::take(&mut self.current_block);
             match self.reader.blocks().read_next_or_eof(buffer) {
                 Ok(Some(block)) => {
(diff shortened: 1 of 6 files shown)


dvdsk:
@est31 This is kinda a trivial PR but as I am the author, should this get a review by someone else before I merge it? 

And a followup which of the RustAudio members could ask for a review in general (not the monster that is #513 ill take care of that with a fellow engineer, but for example PR #559).

est31 (approved):

dvdsk:
> Tough question. I'm generally available as a reviewer, otherwise I don't really know. Maybe we should make a call for maintainers? rodio should not die because of maintainer bottleneck :).

I have been working through the recent issues and most reporters are quite willing to work on a PR so there is a clear need for Rodio. I am in favor of looking for more maintainers, however I have no idea on how to do so safely. Rodio is a popular dependency, any mistake (or malicious intent) could have a wide impact.

PR #565, opened 2024-03-28
The changelog is now in 'keep a changelog' format

dvdsk:
Mainly because it means we get an unreleased section.his will save maintainers time when releasing lowering the barrier to cut a release which in turn helps out users waiting for fixes/new features. I decided not to change the old entries to the keepachangelog format, simply because it is a lot of work and I do not see an advantage.

I would like to also automate checking if the changelog was edited. Since this would further lower the workload on maintainers. The responsibility of editing the changelog will then squarly fall on the Contributor of the PR. That should give maintainers more time to review PR's.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,23 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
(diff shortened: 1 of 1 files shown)


dvdsk:
I'm gonna go ahead and merge this, we can always revert it.

PR #568, opened 2024-04-13
fix: no variant found for `source::SeekError`

YXL76:


dvdsk:
Good catch thanks a lot!

Issue #525, opened 2023-12-05
Playing Mp3 only at cost of blocking UI thread 


EliasDerHai:
Help wanted on my first rust project that goes beyond cli todo / hangman.

I am trying to get a Audio Visualizer App going (FFT wave), but am a bit stuck on the start of my little learning project.

I cant make my mp3 play in the background without freezing the main thread.
Here is my code:

```
use iced::{
    widget::{Button, Column, Text},
    Application, Command,
};
use native_dialog::FileDialog;
use rodio::{Decoder, OutputStream, OutputStreamHandle, Sink};
use std::fs::File;
use std::path::PathBuf;

pub struct AudioVisualizer {
    file_path: Option<PathBuf>,
    audio_output: Option<OutputStreamHandle>,
    audio_sink: Option<Sink>,
}

#[derive(Debug, Clone)]
pub enum Message {
    OpenPressed,
    PlayPressed,
}

impl Application for AudioVisualizer {
    type Executor = iced::executor::Default;
    type Message = Message;
    type Flags = ();
    type Theme = iced::Theme;

    fn new(_flags: Self::Flags) -> (AudioVisualizer, Command<Self::Message>) {
        let (_stream, stream_handle) = OutputStream::try_default().unwrap();
        (
            AudioVisualizer {
                file_path: None,
                audio_output: Some(stream_handle),
                audio_sink: None,
            },
            Command::none(),
        )
    }

    fn title(&self) -> String {
        String::from("Audio Visualizer")
    }

    fn update(&mut self, _message: Self::Message) -> Command<Self::Message> {
        match _message {
            Message::OpenPressed => {
                match FileDialog::new()
                    .add_filter("Audio Files", &["mp3"])
                    .show_open_single_file()
                {
                    Ok(Some(path)) => {
                        println!("File selected: {:?}", path.file_name());
                        self.file_path = Some(path);
                    }
                    Ok(None) => {
                        self.file_path = None;
                    }
                    Err(err) => {
                        println!("File dialog error: {:?}", err);
                        self.file_path = None;
                    }
                }
            }
            Message::PlayPressed => {
                println!("Start pressed");
                if let Some(path) = &self.file_path {
                    match File::open(path) {
                        Ok(file) => match Decoder::new(std::io::BufReader::new(file)) {
                            Ok(source) => {
                                let (_stream, stream_handle) = OutputStream::try_default().unwrap();
                                match Sink::try_new(&stream_handle) {
                                    Ok(sink) => {
[...]

PR #575, opened 2024-04-22
Add comment about Sink and OutputStream

senlizishi:
I'm having the same issue described in #525  regarding Sink and OutputStream lifetimes. Since there is no relevant description in the documentation, I added the relevant description.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -50,7 +50,7 @@
 //! use std::time::Duration;
 //! use rodio::{Decoder, OutputStream, Sink};
 //! use rodio::source::{SineWave, Source};
-//! 
+//!
 //! // _stream must live as long as the sink
 //! let (_stream, stream_handle) = OutputStream::try_default().unwrap();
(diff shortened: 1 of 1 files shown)


dvdsk:
Thanks a lot @senlizishi! It seems rustfmt does not like it :sweat_smile: we can easily fix that though. If you run `cargo fmt` and then push again it should be fixed.

src/lib.rs
```diff
@@ -68,6 +69,8 @@
 //! sounds to play simultaneously, you should create multiple [`Sink`]s.
 //!
 //! The [`Sink`] type also provides utilities such as playing/pausing or controlling the volume.
+//! 
+//! **Please note that the [`Sink`] requires the [`OutputStream`], thus ensuring that it has to live as long as the sink.**
```
dvdsk, line 73: One small suggestion, maybe: "make sure that the outputstream is not dropped before the sink" is better then "thus ensuring that it has to live as long as the sink"?

Issue #581, opened 2024-06-01
Specific mp4a codec panics rodio


apolunar:
Audio streams downloaded from YouTube fail to play in rodio, likely as a result of the codec, although it plays fine in VLC.

Here's the panic:
```
thread 'main' panicked at src/main.rs:31:59:
called `Result::unwrap()` on an `Err` value: IoError("end of stream")
stack backtrace:
   0: rust_begin_unwind
             at /rustc/7cf61ebde7b22796c69757901dd346d0fe70bd97/library/std/src/panicking.rs:647:5
   1: core::panicking::panic_fmt
             at /rustc/7cf61ebde7b22796c69757901dd346d0fe70bd97/library/core/src/panicking.rs:72:14
   2: core::result::unwrap_failed
             at /rustc/7cf61ebde7b22796c69757901dd346d0fe70bd97/library/core/src/result.rs:1649:5
   3: core::result::Result<T,E>::unwrap
             at /rustc/7cf61ebde7b22796c69757901dd346d0fe70bd97/library/core/src/result.rs:1073:23
   4: rustube_demo::main::{{closure}}
             at ./src/main.rs:31:17
   5: tokio::runtime::park::CachedParkThread::block_on::{{closure}}
             at /home/user/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.38.0/src/runtime/park.rs:281:63
   6: tokio::runtime::coop::with_budget
             at /home/user/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.38.0/src/runtime/coop.rs:107:5
   7: tokio::runtime::coop::budget
             at /home/user/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.38.0/src/runtime/coop.rs:73:5
   8: tokio::runtime::park::CachedParkThread::block_on
             at /home/user/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.38.0/src/runtime/park.rs:281:31
   9: tokio::runtime::context::blocking::BlockingRegionGuard::block_on
             at /home/user/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.38.0/src/runtime/context/blocking.rs:66:9
  10: tokio::runtime::scheduler::multi_thread::MultiThread::block_on::{{closure}}
[...]

Issue #577, opened 2024-05-05
With symphonia m4a seek is inaccurate


dvdsk:
Seems to be an issue upstream in symphonia. Needs a minimal example to bring it up there.

PR #582, opened 2024-06-01
Fix m4a and mp4a playback

apolunar:
Addresses issue #581 where all tests (including the new one) pass.
1. It includes the necessary symphonia features
2. Handles "end of stream" decoding errors
3. Selects only tracks with supported codecs
4. Adds test for m4pa file from YouTube (CC License)
Let me know what revisions need to be made to get this merged.


apolunar:
Yikes, well it worked on my machine LOL

dvdsk:
> just not seeking on the RL.m4a, which is something I will fix later (for now maybe we can just disable the test again)

I strongly suspect that's a bug in symphonia, see: #577. I haven't been able to work on bringing that to the attention of the symphonia folk yet unfortunatly.

> Addresses issue https://github.com/RustAudio/rodio/issues/581 where all tests (including the new one) pass.

Thank you so much for adding a test :+1:  that makes reviewing sooo much easier. I'm gonna take a detailed look tomorrow! At a quick glance it seems great though!

src/decoder/symphonia.rs
```diff
@@ -76,8 +76,17 @@ impl SymphoniaDecoder {
             None => return Ok(None),
         };
 
+        // Select the first supported track
+        let track_id = probed.format.tracks()
+            .iter()
+            .find(|t| t.codec_params.codec != CODEC_TYPE_NULL).unwrap().id;
+
+        let decode_opts = Default::default();
```
dvdsk, line 84: might make sense to inline this down below

src/decoder/symphonia.rs
```diff
@@ -76,8 +76,17 @@ impl SymphoniaDecoder {
...
+
+        let track = probed.format.tracks().iter().find(|track| track.id == track_id).unwrap();
+
         let mut decoder = symphonia::default::get_codecs()
-            .make(&stream.codec_params, &DecoderOptions { verify: true })?;
+            .make(&track.codec_params, &decode_opts)?;
```
dvdsk, line 89: so this would be something like: `.make(...., &DecoderOptions::default())`?

dvdsk (changes requested):

dvdsk:
running `cargo fmt` should fix that last failing test

dvdsk:
Some nitpicks and a small mistake regarding the crate features otherwise looks great. With some tweaks we can get it merged!

Cargo.toml
```diff
@@ -15,7 +15,7 @@ claxon = { version = "0.4.2", optional = true }
 hound = { version = "3.3.1", optional = true }
 lewton = { version = "0.10", optional = true }
 minimp3_fixed = { version = "0.5.4", optional = true}
-symphonia = { version = "0.5.2", optional = true, default-features = false }
+symphonia = { version = "0.5.4", optional = true, default-features = false, features = ["all"] }
```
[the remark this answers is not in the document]
est31: yes we absolutely shouldn't enable all features by default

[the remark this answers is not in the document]
apolunar: Sorry, this was a quick fix I forgot to actually investigate :sweat: 

> This increases compile time, is it needed to fix m4a? Might be better to adjust the features below.

I'm not really sure where to put them below, but it seems like you need features = ["aac", "isomp4"] for support

apolunar:
> Some nitpicks and a small mistake regarding the crate features otherwise looks great. With some tweaks we can get it merged!

Yippee! I think I covered everything, thanks for pointing that stuff out.

src/decoder/symphonia.rs
```diff
@@ -77,16 +77,23 @@ impl SymphoniaDecoder {
...
-        let track_id = probed.format.tracks()
+        let track_id = probed
+            .format
+            .tracks()
             .iter()
-            .find(|t| t.codec_params.codec != CODEC_TYPE_NULL).unwrap().id;
-
-        let decode_opts = Default::default();
+            .find(|t| t.codec_params.codec != CODEC_TYPE_NULL)
+            .unwrap()
```
dvdsk, line 85: This panics if there is not track with a valid codec or if there are no tracks at all. It would be better to return a DecodeError I think. Since a music player (one of the usecases of rodio) might very well run into bad files.

src/decoder/symphonia.rs
```diff
@@ -77,16 +77,23 @@ impl SymphoniaDecoder {
...
+            .format
+            .tracks()
+            .iter()
+            .find(|track| track.id == track_id)
+            .unwrap();
 
-        let mut decoder = symphonia::default::get_codecs()
-            .make(&track.codec_params, &decode_opts)?;
+        let mut decoder =
+            symphonia::default::get_codecs().make(&track.codec_params, &Default::default())?;
```
dvdsk, line 96: another nitpick (feel free to ignore if you disagree). I personally find: `Type::trait_fn` clearer then `Trait::trait_fn()`. I think the former would be `DecoderOpt::default()` here.

src/decoder/symphonia.rs
```diff
@@ -97,7 +104,10 @@ impl SymphoniaDecoder {
         let decoded = loop {
             let current_frame = match probed.format.next_packet() {
                 Ok(packet) => packet,
-                Err(_) => break decoder.last_decoded() // IoError end of stream is expected
+                Err(e) => match e {
```
dvdsk, line 107: You can inline this match so:
Err(Error::IoError(_)) => break etc etc,
Err(other) => return Err(other),

src/decoder/symphonia.rs
apolunar: I'm not sure what error to return / where to put the error string, as I think it's bad practice to just put it in the file as I did in the most recent commit.

src/decoder/symphonia.rs
```diff
@@ -77,13 +77,19 @@ impl SymphoniaDecoder {
...
         // Select the first supported track
-        let track_id = probed
+        let track_id = match probed
             .format
             .tracks()
             .iter()
             .find(|t| t.codec_params.codec != CODEC_TYPE_NULL)
-            .unwrap()
-            .id;
+        {
```
dvdsk, line 85: Looking good, thanks again for the investigation and fix!

One more tweak you could try (if you want to) is to replace the match with something using [`Option::ok_or`](https://doc.rust-lang.org/std/option/enum.Option.html#method.ok_or), that would look something like this:

```rust
.find(...)
.ok_or(symphonia::core::errors::Error::Unsupported(...))?
.id;
```

But if you rather leave it thats fine too! This code is better then most in rodio so we are good to merge it without any more changes :)

apolunar:
YIPPEE! Thanks for your help with this :D

dvdsk (approved):

Cargo.toml
```diff
@@ -15,7 +15,7 @@ claxon = { version = "0.4.2", optional = true }
 hound = { version = "3.3.1", optional = true }
 lewton = { version = "0.10", optional = true }
 minimp3_fixed = { version = "0.5.4", optional = true}
-symphonia = { version = "0.5.2", optional = true, default-features = false }
+symphonia = { version = "0.5.4", optional = true, default-features = false, features = ["aac", "isomp4"] }
```
dvdsk, line 18: oeps, these should be gone too, no worries ill do it after the merge

dvdsk: done

PR #585, opened 2024-06-09
Add TrackPosition to Source

### LINKED ISSUE #457: How to obtain the current playing time and position when rodio plays music files
How to obtain the current playing time and position when rodio plays music files

At present, I want to try to build a simple music playing software,

Music playing software needs the most basic

"Total duration" of music files

Current play position "Play time position"

Jump to the specified position "Time Position"

I don't know how to get the "play time position" at present

Thank you for your help

Gusted:
This adds a new TrackPosition wrapper which counts the amount of times `next()` is called (where it returns `Some`) and provide a function to get the position, which is a simple calculation.

This is added to Sink by default and the accuracy of this depends on the interval of `periodic_access`.

I wasn't able to add testing to the sink counter part, because I wanted to also test `try_seek`, but it seems like I would need to have some threading code to call `next` on sink in another thread, which didn't look that good and resulted in a flaky test so only a 'unit test' in ``position.rs`.

Resolves #457
Closes #510

### DIFF
--- a/src/sink.rs
+++ b/src/sink.rs
@@ -64,6 +64,7 @@ struct Controls {
     speed: Mutex<f32>,
     to_clear: Mutex<u32>,
     seek: Mutex<Option<SeekOrder>>,
+    position: Mutex<f64>,
 }
 
 impl Sink {
@@ -90,6 +91,7 @@ impl Sink {
                 speed: Mutex::new(1.0),
                 to_clear: Mutex::new(0),
                 seek: Mutex::new(None),
+                position: Mutex::new(0.0),
             }),
             sound_count: Arc::new(AtomicUsize::new(0)),
             detached: false,
@@ -119,6 +121,7 @@ impl Sink {
 
         let source = source
             .speed(1.0)
+            .trackable()
             .pausable(false)
             .amplify(1.0)
             .skippable()
@@ -127,12 +130,16 @@ impl Sink {
             .periodic_access(Duration::from_millis(5), move |src| {
                 if controls.stopped.load(Ordering::SeqCst) {
                     src.stop();
+                    *controls.position.lock().unwrap() = 0.0;
                 }
                 {
                     let mut to_clear = controls.to_clear.lock().unwrap();
                     if *to_clear > 0 {
                         src.inner_mut().skip();
                         *to_clear -= 1;
+                        *controls.position.lock().unwrap() = 0.0;
+                    } else {
+                        *controls.position.lock().unwrap() = src.inner().inner().inner().inner().get_pos();
                     }
                 }
                 let amp = src.inner_mut().inner_mut();
@@ -140,6 +147,7 @@ impl Sink {
                 amp.inner_mut()
                     .set_paused(controls.pause.load(Ordering::SeqCst));
                 amp.inner_mut()
+                    .inner_mut()
                     .inner_mut()
                     .set_factor(*controls.speed.lock().unwrap());
                 if let Some(seek) = controls.seek.lock().unwrap().take() {
@@ -308,6 +316,12 @@ impl Sink {
     #[inline]
     pub fn len(&self) -> usize {
         self.sound_count.load(Ordering::Relaxed)
+    }
+
+    /// Returns the position of the sound that's being played.
+    #[inline]
+    pub fn get_pos(&self) -> f64 {
+        *self.controls.position.lock().unwrap()
     }
 }
 
--- a/src/sink.rs
+++ b/src/sink.rs
@@ -1,4 +1,4 @@
-use std::sync::atomic::{AtomicBool, AtomicU64, AtomicUsize, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
@@ -12,28 +12,6 @@ use crate::stream::{OutputStreamHandle, PlayError};
 use crate::{queue, source::Done, Sample, Source};
 use cpal::FromSample;
 
-#[derive(Debug)]
-pub struct AtomicF64 {
-    storage: AtomicU64,
-}
-
-impl AtomicF64 {
-    pub fn new(value: f64) -> Self {
-        let as_u64 = value.to_bits();
-        Self {
-            storage: AtomicU64::new(as_u64),
-        }
-    }
-    pub fn store(&self, value: f64, ordering: Ordering) {
-        let as_u64 = value.to_bits();
-        self.storage.store(as_u64, ordering)
-    }
-    pub fn load(&self, ordering: Ordering) -> f64 {
-        let as_u64 = self.storage.load(ordering);
-        f64::from_bits(as_u64)
-    }
-}
-
 /// Handle to a device that outputs sounds.
 ///
 /// Dropping the `Sink` stops all sounds. You can use `detach` if you want the sounds to continue
@@ -44,7 +22,6 @@ pub struct Sink {
 
     controls: Arc<Controls>,
     sound_count: Arc<AtomicUsize>,
-    position: Arc<AtomicF64>,
 
     detached: bool,
 }
@@ -87,6 +64,7 @@ struct Controls {
     speed: Mutex<f32>,
     to_clear: Mutex<u32>,
     seek: Mutex<Option<SeekOrder>>,
+    position: Mutex<f64>,
 }
 
 impl Sink {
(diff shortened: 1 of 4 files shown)


src/source/position.rs
```diff
@@ -0,0 +1,133 @@
...
+}
+
+impl<I> Source for TrackPosition<I>
+where
+    I: Source,
+    I::Item: Sample,
+{
+    #[inline]
+    fn current_frame_len(&self) -> Option<usize> {
+        self.input.current_frame_len()
```
dvdsk, line 82: after `current_frame_len` the sample rate may change. 
One way to deal with that is to:
- collect samples until the frame rate changes 
- convert the collected samples using the old frame-rate and add them to a duration
- change `get_pos()` to return the newly collected samples * new frame rate + duration

Feel free to experiment with other approaches to find what fits best. Performance matters but as always _premature optimization is the root of all evil_. An if statement that fires every `frame_len` in `next()` might very well pulled up by the optimizer and merged with the iterator in some more optimal fashion. Better to measure the impact before spending time on optimizing.

src/sink.rs
```diff
@@ -127,19 +130,24 @@ impl Sink {
...
+                    *controls.position.lock().unwrap() = 0.0;
                 }
                 {
                     let mut to_clear = controls.to_clear.lock().unwrap();
                     if *to_clear > 0 {
                         src.inner_mut().skip();
                         *to_clear -= 1;
+                        *controls.position.lock().unwrap() = 0.0;
+                    } else {
+                        *controls.position.lock().unwrap() = src.inner().inner().inner().inner().get_pos();
```
dvdsk, line 142: I am planning to overhaul this whole module, the every 5ms check for things to do feels very iffy. One might want more precision. Though that is for later right now. If you have ideas on how to redo this module I would love to hear them.

src/source/position.rs
Gusted: Do you happen to know which decoder currently handle this correctly? I've tried with FLAC using https://github.com/ietf-wg-cellar/flac-test-files/blob/main/uncommon/01%20-%20changing%20samplerate.flac with the symphonia decoder and after switching to the next sample rate it stops working and [claxon](https://github.com/ruuda/claxon) doesn't stop playing but doesn't seem to record the new sample rate (just reports the 32 kHz all the way trough) and it ends up 'pitching' the audio. (FWIW, VLC also borks on this only mpv seems to somewhat handle it but has incorrect position after switching and wrong total duration). Shouldn't be hard to handle these cases correctly, but at least the FLAC decoder is stopping me from testing this and I don't know of any other format that supports variable sample rate 😅

dvdsk: nice! thanks a lot for that :+1: I had no idea these where broken (rodio's test suit is kinda sparse atm)

Gusted: Once the decoders are fixed https://github.com/RustAudio/rodio/pull/585/commits/0e6d5a1da2d75d462f145dcd4ad54d5821599deb should make the position tracking work correctly with. Doing a few local tests doesn't hint at performance problems (but only a profiler would be able to confirm that) and tracking the position of https://github.com/ietf-wg-cellar/flac-test-files/blob/main/uncommon/01%20-%20changing%20samplerate.flac works flawlessly.

src/source/position.rs
```diff
@@ -43,10 +51,20 @@ where
...
     I::Item: Sample,
 {
-    /// Returns the inner source.
+    /// Returns the position of the source.
     #[inline]
     pub fn get_pos(&self) -> f64 {
-        self.samples_elapsed as f64 / self.input.sample_rate() as f64 / self.input.channels() as f64
+        self.samples_collected as f64
+            / self.input.sample_rate() as f64
+            / self.input.channels() as f64
```
dvdsk, line 59: ah yes channels I forgot about those :sweat_smile:

src/source/position.rs
```diff
@@ -59,9 +77,28 @@ where
...
     fn next(&mut self) -> Option<I::Item> {
+        // This should only be executed once at the first call to next.
+        if self.current_frame_len.is_none() {
+            self.set_current_frame();
+        }
+
         let item = self.input.next();
         if item.is_some() {
-            self.samples_elapsed += 1;
+            self.samples_collected += 1;
```
dvdsk, line 87: `samples_counted` might fit better?

dvdsk (commented):
I have also been thinking about the API. Most use cases will take the form of some kind of music player with a UI showing the current position. In that case the position will need to be updated once every second. 

Right now you need access to the source to get the position which in turn requires the `every_x_ms` closure thing in *Sink*. 

So here is a wild (and possibly bad) idea, lets store the current position in an Arc<Struct with atomics>. We can use relaxed ordering so the performance impact of using atomics should be negligible. The Arc<struct of atomics> is Clone, Send and Sync, it could be returned by the factory function* that creates the stream. The Sink could store it in its handle, but also readily hand it out. Users could even put it in their UI loop.

Is this worth exploring?

*thats how to refer to members that create Self right?

Gusted:
I've implemented the `Arc` idea in https://github.com/RustAudio/rodio/pull/585/commits/ca43ad088526d29612c25bb8e83063f1c5657d1b I'm not sure if we should hand out references to the position `Arc` or if we should just provide the `get_pos` function. As the latter works fine in my case (music player) and avoids having users of this library having to handle that extra `Arc` for just position.

src/sink.rs
```diff
@@ -12,6 +12,28 @@ use crate::stream::{OutputStreamHandle, PlayError};
 use crate::{queue, source::Done, Sample, Source};
 use cpal::FromSample;
 
+#[derive(Debug)]
+pub struct AtomicF64 {
```
dvdsk, line 16: nice solution to not having an atomicf64!

dvdsk:
> I'm not sure if we should hand out references to the position `Arc` or if we should just provide the `get_pos` function. 

I do not think we should expose the fact there is an `Arc` involved, wrapping it with a struct that provides a `get_pos` would be a good solution to that.

Fundamentally the question is whether you need to own the source to get the position (no arc) or if there is a position handle (the arc) that you can freely move around your program. 

Right now rodio uses the first approach for everything. Concretely if you want interact with any source you need to own that source*. To kinda work around that rodio introduces a smart wrapper the `periodic_access` wrapper. That swaps the _need to own_ disadvantage for a _slight delay before the interaction_ or _outdated info_ disadvantage. This is luckily hidden by the `Sink` for most users. The `Sink` itself uses a ton synchronization primitives to communicate between a handle and the closure in `periodic access`.

We (probably I :sweat_smile: ) could improve the `Sink` by swapping out `periodic_access` for `signaled_access`, that would run a closure on a signal. That would get rid of the _delay_ and _outdated info_.

> As the latter works fine in my case (music player) and avoids having users of this library having to handle that extra Arc for just position.

If you think its more of a hassle then we should not go for it. The only advantage the Arc is that is would be clone (the sink is not clone atm). That could make using the arc easier for the end user. If you think it is actually easier to query the `Sink` then it is better do away with the arc. 

I have some ideas of how to make the `Sink` clone, so even that disadvantage can be negated in the future.

Thanks for trying it out though! 


*A source being a stack of wrappers (a trackposition wrapping pausable wrapping decoder)

Gusted:
> If you think its more of a hassle then we should not go for it. The only advantage the Arc is that is would be clone (the sink is not clone atm). That could make using the arc easier for the end user. If you think it is actually easier to query the Sink then it is better do away with the arc.

My use case just relies on the fact that `Sink` exists and it (ab)uses that. So having that and have it exposes functions like `get_pos` and `volume` etc. makes it a nice high level struct. I think you would end up with some boilerplate code for the end users if it was a low level struct where you will be given `Arc`s. 

> Fundamentally the question is whether you need to own the source to get the position (no arc) or if there is a position handle (the arc) that you can freely move around your program.

I assume users either have a `Source` or a `Sink` in their state that's available in an event loop (assuming there's one). From what I understand, it could become no Arc because in the `Source` case, you do own the `Source` and in the `Sink` case `Sink` owns the `Source` and could do the necessary code to return the position. I don't think owning an Arc would be an good idea per the previous paragraph.

dvdsk:
Agreed.

So lets revert the Arc<Atomic> solution and then we can merge I think. You can largely just revert the last commit though there are some changes in there unrelated to the Arc<Atomic> stuff.

src/source/mod.rs
```diff
@@ -333,6 +335,13 @@ where
         skippable::skippable(self)
     }
 
+    fn trackable(self) -> TrackPosition<Self>
```
dvdsk, line 338: I like the name `TrackPosition`, it communicates that it has to do with the position of the source and tracking it. Would you be okay with changing the name `trackable` to `track_position`? Then the factory function clearly communicates that what is tracked is the position.

dvdsk:
Merged! Thanks a lot for the work!

dvdsk:
I changed the return type from f64 to Duration, let me know if you disagree on that then we can discuss it

PR #595, opened 2024-07-11
Derive `Clone` for `buffer::SamplesBuffer`

drzymalanet:
This change applies a #derive attribute macro adding the `Clone` trait to the `SamplesBuffer` struct.


dvdsk:
Im kinda suprised this works without an additional bound on S, I guess the Sample trait needs S to be Clone already.

Anyways LGTM

PR #598, opened 2024-07-19
Add ALAC / AIFF Support

### LINKED ISSUE #193: ALAC/AIFF support
Taken from the readme, rodio doesn't support ALAC file playback (the compressed lossless file format used by Apple). Would be nice to have this too! Not sure if there are any crates that decode the data though.

Since this is kind of related to the prior question about file support, is AIFF supported as well?

Thanks!

Keinsleif:
Add ALAC/AIFF playback support.

The change is just add the following feature flags that enable symphonia's feature flags.
"symphonia-alac" enable alac playback
"symphonia-aiff" enable aiff playback

- Why these flags are not added in "symphonia-all" feature?
  "symphonia/alac" and "symphonia/aiff" features are not enabled by default in symphonia.

Fix #193


dvdsk:
Thanks for noticing and fixing this!

PR #602, opened 2024-08-10
Synthesizer Waveforms 

iluvcapra:
Adds a new Source, `Synth`, that can generate square, triangle, sawtooth and sine waveforms useful for additive music sound synthesis. Waveforms can be generated with arbitrary sample rates and frequencies.

Also the existing `SineWave` Source has been reimplemented as a client of `Synth`.


iluvcapra:
Can we hold on this for a moment? I'd like to add white and pink noise and I don't think it'd be too difficult.

iluvcapra:
Okay check that out.

src/source/noise.rs
```diff
@@ -0,0 +1,135 @@
...
+        Self { sample_rate }
+    }
+}
+
+impl Iterator for WhiteNoise {
+    type Item = f32;
+
+    #[inline]
+    fn next(&mut self) -> Option<Self::Item> {
+        let randf = rand::random::<f32>();
```
dvdsk, line 35: take a look at the random number generators in the rand crate. Most of them are very slow but in exchange output cryptographically secure numbers. I do not think that is needed here but we do pay the price for it.

iluvcapra: Will do. I'm using the `cpal::SampleRate` type for sample rate parameter where we ask for it. Is that okay or would you rather it's just a `u32`?

iluvcapra: In a015eb49a2

dvdsk: > Will do. I'm using the `cpal::SampleRate` type for sample rate parameter where we ask for it. Is that okay or would you rather it's just a `u32`?

I would match what the other signal generator/synthesizer sources do (Sine is one I believe?)

iluvcapra:
What if I just called it “TestSignal”, that’s why I did it, I wanted to make signals as test sources for other sources.On Aug 13, 2024, at 7:22 AM, David Kleingeld ***@***.***> wrote:﻿
@dvdsk commented on this pull request.



In src/source/noise.rs:
> +    sample_rate: cpal::SampleRate,
+}
+
+impl WhiteNoise {
+    /// Create a new white noise generator.
+    pub fn new(sample_rate: cpal::SampleRate) -> Self {
+        Self { sample_rate }
+    }
+}
+
+impl Iterator for WhiteNoise {
+    type Item = f32;
+
+    #[inline]
+        let randf = rand::random::<f32>();


In a015eb4

small_rng looks good! We might wanna put the whole synthesizer (and rand dependency) behind a cargo feature. Rodio is used in a ton of large projects that do not need synthesizers but without a feature the will pay the cost of compiling the rand crate.

iluvcapra:
What I'd propose is just renaming `Synth` into `TestSignal`.

I definitely think there should at least be a `WhiteNoise` and `PinkNoise`, those have a lot of applications as test signals, it's hard to subjectively judge loudness differences with sine waves, they're better for testing dynamic processes, they're obligatory for testing monitors etc. I added it because I was going to do some dynamic gain sources and it just makes more sense to test and do an example with pink noise for those.

Square and triangle waves I'm more equivocal on. A lot of signal generator plugins make them, and not in a particularly musical context. It's definitely useful to be able to see them on a scope, square waves can tell you about slew rate in a DA, triangle waves have a known spectrum and you can see aberrant harmonics easily.

If we're talking about test signals maybe I'm missing one, we could definitely use a chirp signal generator, that's really useful for evaluating frequency response.

dvdsk:
you clearly know what you are talking about which is very nice since I do not :sweat_smile:. I happily defer to your judgement :+1:. 

And I  agree on the Synth -> TestSignal rename, that is clearer. I see you have also already reimplemented the Sine source as Synth/TestSignal :+1: . 

We could deprecate it in the future (the Sine source), that would clean up the API a bit.

src/source/test_waveform.rs
```diff
@@ -0,0 +1,196 @@
+//! Generator sources for various periodic test waveforms.
+//!
+//! This module provides several periodic, deterministic waveforms for testing other sources and
+//! for simple additive sound synthesis. Every source is monoaural and in the codomain `[-1.0f32,
```
dvdsk, line 4: if its for testing *and* sound synthesis then neither *synth* or *test* are the right terms to describe it. Maybe waveform?

src/source/test_waveform.rs
```diff
@@ -0,0 +1,196 @@
...
+use crate::Source;
+
+/// Test waveform functions.
+#[derive(Clone, Debug)]
+pub enum TestWaveformFunction {
+    /// A sinusoidal waveform.
+    Sine,
+    /// A triangle wave.
+    Triangle,
+    /// A square wave, rising edge at t=0.
```
dvdsk, line 27: does it matter where the rising edge is? The more specific we document the behavior the less flexible the implementation can be. If we say the rising edge is at t=0 then one user can start depending on that and we can never change it without their code breaking. 

I do not know enough about audio to know if its useful to specify where the rising edge starts. If this is something that is needed then I am all for documenting it.

src/source/test_waveform.rs
```diff
@@ -0,0 +1,196 @@
...
+    Square,
+    /// A rising swatooth wave.
+    Sawtooth,
+}
+
+impl TestWaveformFunction {
+    /// Create a single sample for the given waveform
+    #[inline]
+    fn render(&self, i: u64, period: f32) -> f32 {
+        let i_div_p: f32 = i as f32 / period;
```
dvdsk, line 37: we need to fix this variable name. Its current name describes what it is but not what that means.

maybe `step`, `cycle` `position`?

src/source/test_waveform.rs
```diff
@@ -0,0 +1,196 @@
...
+
+        match self {
+            Self::Sine => (TAU * i_div_p).sin(),
+            Self::Triangle => 04.0f32 * (i_div_p - (i_div_p + 0.5f32).floor()).abs() - 1f32,
```
dvdsk, line 41: 04 -> 4?

src/source/test_waveform.rs
```diff
@@ -0,0 +1,196 @@
...
+        }
+    }
+}
+
+/// An infinite source that produces one of a selection of test waveforms.
+#[derive(Clone, Debug)]
+pub struct TestWaveform {
+    sample_rate: cpal::SampleRate,
+    period: f32,
+    f: TestWaveformFunction,
```
dvdsk, line 59: f -> function

a single letter is almost always confusing. Is it frequency, a function the fraction?

src/source/test_waveform.rs
```diff
@@ -0,0 +1,196 @@
...
+        }
+    }
+}
+
+impl Iterator for TestWaveform {
+    type Item = f32;
+
+    #[inline]
+    fn next(&mut self) -> Option<f32> {
+        let this_i = self.i;
```
dvdsk, line 87: I would just write:
```
let val = Some(self.f.render(i, self.period));
self.i += 1;
val
```

src/source/test_waveform.rs
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    #[inline]
+    fn total_duration(&self) -> Option<Duration> {
+        None
+    }
+
+    #[inline]
+    fn try_seek(&mut self, duration: Duration) -> Result<(), SeekError> {
+        self.i = (self.sample_rate.0 as f32 * duration.as_secs_f32()) as u64;
```
dvdsk, line 116: Does it make sense to implement `try_seek`? Seek was meant as a rough way to go to a  point in time. The duration is measured from the start of playback. 

You seek in a music file a podcast etc usually based on imprecise instructions from the user (go forward 5s go to 13m:23s). If you need precise narrow adjustments to playback (35ms forward) skipping samples seems the better approach (see [the skip duration method which is part of Source](https://docs.rs/rodio/0.19.0/rodio/source/trait.Source.html#method.skip_duration)).

Cargo.toml
```diff
@@ -18,6 +18,7 @@ minimp3_fixed = { version = "0.5.4", optional = true}
 symphonia = { version = "0.5.4", optional = true, default-features = false }
 crossbeam-channel = { version = "0.5.8", optional = true }
 thiserror = "1.0.49"
+rand = {version = "0.8.5", features = ["small_rng"]}
```
dvdsk, line 21: we do not want to depend rand if the noise waveform functions are not needed by the user. So rand should be an optional dependency enabled by some compile time feature. Lets call that feature `noise`?

dvdsk (changes requested):
I've left my opinion on the code here and there its almost exclusively about code style. Let me know if you disagree with anything, then we can discuss that in more detail.

and again thanks for the solid work!

iluvcapra:
I'll get on these in a few days, I'm finishing a Chirp source to complete the suite.

src/source/test_waveform.rs
iluvcapra: Is there a *precise* seeking method? I noticed this when I was chaining together sources for crossfades, sometimes you need to split a source/skip a source to a precise sample.

dvdsk: unfortunately not, the seek implementation is left over to the decoder. The decoder can be really precise, but it could also be off for a bit. Think off by milliseconds not seconds.

Cargo.toml
iluvcapra: Is there a place for feature documentation?

dvdsk: not yet, but that is a good idea :+1: 

best place would probably be here: https://github.com/RustAudio/rodio/blob/master/src/lib.rs

I am planning to overhaul the features related to decoders later so for now lets only document this one.

src/source/test_waveform.rs
iluvcapra: Jus getting back into this, would you like me to delete this from the implementation?

iluvcapra:
This is all passing again, it looks like all of the failures in nightly are outside the scope of the PR.

Cargo.toml
```diff
@@ -28,6 +28,7 @@ vorbis = ["lewton"]
 wav = ["hound"]
 mp3 = ["symphonia-mp3"]
 minimp3 = ["dep:minimp3_fixed"]
+noise = ["rand"]
```
dvdsk, line 31: :+1:

examples/noise_generator.rs
```diff
@@ -31,3 +33,9 @@ fn main() {
 
     thread::sleep(interval_duration);
 }
+
```
dvdsk, line 36: good idea! very nice and helpful!

src/lib.rs
```diff
@@ -112,6 +112,15 @@
 //! the number of sinks that can be created (except for the fact that creating too many will slow
 //! down your program).
 //!
+//! 
+//! ## Features 
+//!
+//! Rodio provides several optional features that are guarded with feature gates.
```
dvdsk, line 118: This is the only <thing> thats not a decoder that lives behind a feature gate. So its not `several` right?

dvdsk (commented):
I just noticed the `todo!()`. So I take it you are not done yet. Take your time! its looking good!

src/lib.rs
iluvcapra: I dunno I expected all the features to eventually be in here, I'll change the language to make it singular.

iluvcapra:
Oh that shouldn't be there, I thought I'd deleted that, I'm not going to support try_seek with any of the signal generators based on our previous convo.

iluvcapra:
Why isn't this building in `nightly` BTW? I've merged the upstream so I'm not sure what the problem is.

dvdsk:
> Why isn't this building in `nightly` BTW? I've merged the upstream so I'm not sure what the problem is.

Its building fine its just that the test builds are failing as there are undocumented methods. The wierd thing is, that lint never fired in the past... My guess is that the lint was broken and has been fixed in the latest nightly.

dvdsk:
I'll disable the lint or fix the docs

dvdsk:
if you merge main all the lint errors not related to this PR should disappear :+1:

src/lib.rs
dvdsk: You where right :laughing:, rodio got optional support for tracing (logging library) so we need that description. I'm copying it to main now. You can add the synthesizer note to that.

src/source/test_waveform.rs
```diff
@@ -0,0 +1,185 @@
...
+/// Waveform functions.
+#[derive(Clone, Debug)]
+pub enum Function {
+    /// A sinusoidal waveform.
+    Sine,
+    /// A triangle waveform.
+    Triangle,
+    /// A square wave, rising edge at t=0.
+    Square,
+    /// A rising swatooth wave.
```
UnknownSuperficialNight, line 29: 90% sure this is a typo

iluvcapra: Fixed in 56fcce74.

src/lib.rs
iluvcapra: I've fixed the documentation here so both flags are in the same header block.

iluvcapra:
OMG this thread is so long, I'm so so sorry.

iluvcapra (commented):
Good for me.

src/source/test_waveform.rs
dvdsk: nah its okay, its not that complex. Fine to keep it in.

dvdsk: I was on the fence but I thought about it and I really think we do not want ppl to rely on try_seek for such precise 'seeking'. Skip samples is the correct approach for that. Do you agree?

src/source/test_waveform.rs
```diff
@@ -0,0 +1,185 @@
...
+/// An infinite source that produces one of a selection of test waveforms.
+#[derive(Clone, Debug)]
+pub struct TestWaveform {
+    sample_rate: cpal::SampleRate,
+    period: f32,
+    function: Function,
+    i: u64,
+}
+
+impl TestWaveform {
```
dvdsk, line 63: To me `Test` implies its only for debugging, however I think this is more useful then that. I'am also afraid of ppl skipping it when they see `Test` in the name, assuming its not reliable/performant/stable. We could call it `SimpleWaveform`? How does that sound to you?

src/source/sine.rs
```diff
@@ -59,11 +57,7 @@ impl Source for SineWave {
     }
 
     #[inline]
-    fn try_seek(&mut self, _: Duration) -> Result<(), SeekError> {
-        // This is a constant sound, normal seeking would not have any effect.
-        // While changing the phase of the sine wave could change how it sounds in
-        // combination with another sound (beating) such precision is not the intend
-        // of seeking
-        Ok(())
```
dvdsk, line 67: see the thread on try_seek. I would propose make the original comment a doc comment and mentioning `skip_samples` as the solution for things like beating

dvdsk (changes requested):
Its a great PR, and brings a lot of functionality to rodio. Thank you very much for adding all this :+1: 

I still have some small things, if you disagree that is fine too! Then we can merge it as is.

Have a great day!

src/source/test_waveform.rs
iluvcapra: Other suitable terms might be `SignalGenerator`, `WaveformGenerator`, `FunctionGenerator`, `ToneGenerator`. Maybe ToneGenerator...

dvdsk: all very good names, damn.. 

My personal favorite is `SignalGenerator`, I have a background in physics and thats how we called the device that generates sawtooth/sine etc. The choice is yours :+1:

iluvcapra: `SignalGenerator` is good, we don't reserve the word "signal" for anything special to the framework so yeah that works for me.

src/source/sine.rs
iluvcapra: In [322bdd5](https://github.com/RustAudio/rodio/pull/602/commits/322bdd5f5f66d1930f09e193fc1f48314ec11626)

dvdsk:
LGTM! :partying_face:

iluvcapra:
Epic.

Issue #619, opened 2024-09-16
Tracking issue for redesigning the Sink API


dvdsk:
There are currently two seperate PR's that expand on the Sink's functionality: 
- https://github.com/RustAudio/rodio/pull/600
- https://github.com/RustAudio/rodio/pull/506 

This issue is to discuss a new Sink which supports both those applications, maybe by merging those PR's or by making breaking changes to the current Sink.

This should take https://github.com/RustAudio/rodio/issues/613 into account.

teliosdev:
I think having `append` return a type that allows the user to register interest in when the sound is finished playing, whether that's sync or async. You may be able to avoid exposing channels at all with this, as the returned (opaque) type can have `fn wait(&self)` (which blocks the calling thread until the sound is done playing), and `async fn stopped(&self)`, which only completes when the sound is done.

This also allows extension in the future to perform acts upon the specific sound, if that's desired, as well.

You can have this type take `&self` instead of `&mut self` by replacing the use of `std::sync::mpsc` with a `Mutex`/`Condvar` for the blocking API.  For the async API, you may be able to get away with an `AtomicBool`/`AtomicWaker`.

PetrGlad:
A problem I see with the Sink implementation is that forces to have all the functions embedded every time even if users do not need them. So maybe that makes `Sink::append(...)` implementation so cumbersome. It combines all the layers at once and then unwraps them in reverse in the callback. For example, to update the speed from a control it takes `src.inner_mut().inner_mut().inner_mut().inner_mut()`.

I think, atomics can be used to control parameters without using a periodic callback. For `f32` one can use `AtomicCell<T>` from `crossbeam-utils`. It is lock-free for `f32`.  While sink is constructed the control handles can be collected and used directly to adjust parameters. Currently parameters are plain numeric values, one can even support both dynamic controls with atomics and plain values by allowing any `Deref` as the parameter value. So for dynamic control it could be for example `Arc<AtomicCell<f32>>` and for static config f32.

dvdsk:
> A problem I see with the Sink implementation is that forces to have all the functions embedded every time even if users do not need them. 

Ahh yes, that could possibly slow down code. I thought of that a while back one of the solutions I came up with is to make Sink generic over a ton of const generic parameters like `Pausable`, `Amplifyable` these are then true or false. The member pausable will only be available on `Sink<true , .. >` (given `Sink<Pausable, Amplifyable, other things..>`).

[...]

PR #621, opened 2024-09-26
Add Automatic Gain Control

### LINKED ISSUE #620: Idea of implementing AGC (Automatic Gain Control)
I was thinking for the sink: would you be okay with me looking into making a pull request for implementing AGC for the sink? This would allow it to dynamically adjust the audio volume by increasing the gain if the audio level is too low or decreasing it if the level is too high. 

Apologies if this is already implemented somewhere; I checked the documentation but could not find a mention of it.

UnknownSuperficialNight:
Implemented automatic gain control (AGC)

I implemented a [to_f32](https://github.com/UnknownSuperficialNight/rodio/blob/feature/automatic-gain-control/src/conversions/sample.rs#L83C1-L84C28) conversion for the sample as it was missing. I'm not entirely certain if this is optimal, so I welcome feedback on the implementation.

This pull request is in relation to. Closes #620 

**Here is my `main.rs` and `Cargo.toml` i used to test while developing:**
```rust
use rodio::source::Source;
use rodio::{Decoder, OutputStream, Sink};
use std::fs::File;
use std::io::BufReader;

fn main() {
    // Get a output stream handle to the default physical sound device
    let (_stream, stream_handle) = OutputStream::try_default().unwrap();
    // Create a sink
    let sink = Sink::try_new(&stream_handle).unwrap();

    // Load a sound from a file, using a path relative to Cargo.toml
    let file =
        BufReader::new(File::open(env!("CARGO_MANIFEST_DIR").to_owned() + "/OOTW.mp3").unwrap());
    // Decode that sound file into a source
    let source = Decoder::new(file).unwrap();

    // Amplify the source
    let amplified_source = source.automatic_gain_control(1.0, 0.3, 3.0);

    // Add the amplified source to the sink
    sink.append(amplified_source);

    // The sound plays in a separate thread.
    sink.sleep_until_end();
}
```

```rust
[workspace]
members = [".", "rodio"]

[package]
name = "test_area"
version = "0.1.0"
edition = "2021"

[dependencies]
rodio = { path = "rodio" }
```

Id suggest an attack_time of around 0.5 to 2.5 that seems to be a sweet spot for songs I tested.
~~Also, target_level I adjusted to be 0.9 just to make sure it does not clip the audio; it was clipping for me at 1.0.~~

Don't know where to write documentation or if I should be the one writing it. Idk, let me know; thanks :)

### DIFF
--- a/benches/effects.rs
+++ b/benches/effects.rs
@@ -64,7 +64,6 @@ fn agc_enabled(bencher: Bencher) {
 }
 
 #[cfg(feature = "experimental")]
-// Do note you will need to pass --features experimental to run this benchmark
 #[divan::bench]
 fn agc_disabled(bencher: Bencher) {
     bencher
--- a/benches/effects.rs
+++ b/benches/effects.rs
@@ -54,10 +54,10 @@ fn agc(bencher: Bencher) {
         .bench_values(|source| {
             source
                 .automatic_gain_control(
-                    1.0,  // target_level
-                    2.0,  // attack_time (in seconds)
-                    0.01, // release_time (in seconds)
-                    5.0,  // absolute_max_gain
+                    1.0,   // target_level
+                    4.0,   // attack_time (in seconds)
+                    0.005, // release_time (in seconds)
+                    5.0,   // absolute_max_gain
                 )
                 .for_each(divan::black_box_drop)
         })
--- a/benches/effects.rs
+++ b/benches/effects.rs
@@ -63,6 +63,8 @@ fn agc_enabled(bencher: Bencher) {
         })
 }
 
+#[cfg(feature = "experimental")]
+// Do note you will need to pass --features experimental to run this benchmark
 #[divan::bench]
 fn agc_disabled(bencher: Bencher) {
     bencher
--- a/benches/effects.rs
+++ b/benches/effects.rs
@@ -48,7 +48,7 @@ fn amplify(bencher: Bencher) {
 }
 
 #[divan::bench]
-fn agc(bencher: Bencher) {
+fn agc_enabled(bencher: Bencher) {
     bencher
         .with_inputs(|| TestSource::music_wav().to_f32s())
         .bench_values(|source| {
@@ -60,5 +60,27 @@ fn agc(bencher: Bencher) {
                     5.0,   // absolute_max_gain
                 )
                 .for_each(divan::black_box_drop)
+        })
+}
+
+#[divan::bench]
+fn agc_disabled(bencher: Bencher) {
+    bencher
+        .with_inputs(|| TestSource::music_wav().to_f32s())
+        .bench_values(|source| {
+            // Create the AGC source
+            let amplified_source = source.automatic_gain_control(
+                1.0,   // target_level
+                4.0,   // attack_time (in seconds)
+                0.005, // release_time (in seconds)
+                5.0,   // absolute_max_gain
+            );
+
+            // Get the control handle and disable AGC
+            let agc_control = amplified_source.get_agc_control();
+            agc_control.store(false, std::sync::atomic::Ordering::Relaxed);
+
+            // Process the audio stream with AGC disabled
+            amplified_source.for_each(divan::black_box_drop)
         })
 }
--- a/benches/effects.rs
+++ b/benches/effects.rs
@@ -46,3 +46,19 @@ fn amplify(bencher: Bencher) {
         .with_inputs(|| TestSource::music_wav().to_f32s())
         .bench_values(|source| source.amplify(0.8).for_each(divan::black_box_drop))
 }
+
+#[divan::bench]
+fn agc(bencher: Bencher) {
+    bencher
+        .with_inputs(|| TestSource::music_wav().to_f32s())
+        .bench_values(|source| {
+            source
+                .automatic_gain_control(
+                    1.0,  // target_level
+                    2.0,  // attack_time (in seconds)
+                    0.01, // release_time (in seconds)
+                    5.0,  // absolute_max_gain
+                )
+                .for_each(divan::black_box_drop)
+        })
+}
--- a/src/source/agc.rs
+++ b/src/source/agc.rs
@@ -59,6 +59,10 @@ pub struct AutomaticGainControl<I> {
 }
 
 #[cfg(not(feature = "experimental"))]
+/// Automatic Gain Control filter for maintaining consistent output levels.
+///
+/// This struct implements an AGC algorithm that dynamically adjusts audio levels
+/// based on both **peak** and **RMS** (Root Mean Square) measurements.
 #[derive(Clone, Debug)]
 pub struct AutomaticGainControl<I> {
     input: I,
--- a/src/source/agc.rs
+++ b/src/source/agc.rs
@@ -1,3 +1,17 @@
+//
+//      Automatic Gain Control (AGC) Algorithm
+//      Designed by @UnknownSuperficialNight
+//
+//   Features:
+//   • Adaptive peak detection
+//   • RMS-based level estimation
+//   • Asymmetric attack/release
+//
+//   Optimized for smooth and responsive gain control
+//
+//   Crafted with love. Enjoy! :)
+//
+
 use super::SeekError;
 use crate::{Sample, Source};
 use std::time::Duration;
--- a/src/source/agc.rs
+++ b/src/source/agc.rs
@@ -15,7 +15,9 @@
 
 use super::SeekError;
 use crate::{Sample, Source};
+#[cfg(feature = "experimental")]
 use atomic_float::AtomicF32;
+#[cfg(feature = "experimental")]
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 use std::time::Duration;
@@ -36,6 +38,7 @@ const fn power_of_two(n: usize) -> usize {
 /// A larger size provides more stable RMS values but increases latency.
 const RMS_WINDOW_SIZE: usize = power_of_two(8192);
 
+#[cfg(feature = "experimental")]
 /// Automatic Gain Control filter for maintaining consistent output levels.
 ///
 /// This struct implements an AGC algorithm that dynamically adjusts audio levels
@@ -52,6 +55,20 @@ pub struct AutomaticGainControl<I> {
     peak_level: f32,
     rms_window: CircularBuffer,
     is_enabled: Arc<AtomicBool>,
+}
+
+#[cfg(not(feature = "experimental"))]
+#[derive(Clone, Debug)]
+pub struct AutomaticGainControl<I> {
+    input: I,
+    target_level: f32,
+    absolute_max_gain: f32,
+    current_gain: f32,
+    attack_coeff: f32,
+    release_coeff: f32,
+    min_attack_coeff: f32,
+    peak_level: f32,
+    rms_window: CircularBuffer,
 }
 
 /// A circular buffer for efficient RMS calculation over a sliding window.
@@ -124,17 +141,35 @@ where
     let attack_coeff = (-1.0 / (attack_time * sample_rate as f32)).exp();
     let release_coeff = (-1.0 / (release_time * sample_rate as f32)).exp();
 
-    AutomaticGainControl {
-        input,
-        target_level: Arc::new(AtomicF32::new(target_level)),
-        absolute_max_gain: Arc::new(AtomicF32::new(absolute_max_gain)),
-        current_gain: 1.0,
-        attack_coeff: Arc::new(AtomicF32::new(attack_coeff)),
-        release_coeff: Arc::new(AtomicF32::new(release_coeff)),
-        min_attack_coeff: release_time,
-        peak_level: 0.0,
-        rms_window: CircularBuffer::new(),
-        is_enabled: Arc::new(AtomicBool::new(true)),
+    #[cfg(feature = "experimental")]
+    {
+        AutomaticGainControl {
+            input,
+            target_level: Arc::new(AtomicF32::new(target_level)),
+            absolute_max_gain: Arc::new(AtomicF32::new(absolute_max_gain)),
+            current_gain: 1.0,
+            attack_coeff: Arc::new(AtomicF32::new(attack_coeff)),
+            release_coeff: Arc::new(AtomicF32::new(release_coeff)),
+            min_attack_coeff: release_time,
+            peak_level: 0.0,
+            rms_window: CircularBuffer::new(),
+            is_enabled: Arc::new(AtomicBool::new(true)),
+        }
+    }
+
+    #[cfg(not(feature = "experimental"))]
+    {
+        AutomaticGainControl {
+            input,
+            target_level,
+            absolute_max_gain,
+            current_gain: 1.0,
+            attack_coeff,
+            release_coeff,
+            min_attack_coeff: release_time,
+            peak_level: 0.0,
+            rms_window: CircularBuffer::new(),
+        }
     }
 }
 
@@ -143,6 +178,7 @@ where
     I: Source,
     I::Item: Sample,
 {
+    #[cfg(feature = "experimental")]
     /// Access the target output level for real-time adjustment.
     ///
     /// Use this to dynamically modify the AGC's target level while audio is processing.
@@ -152,6 +188,7 @@ where
         Arc::clone(&self.target_level)
     }
 
+    #[cfg(feature = "experimental")]
     /// Access the maximum gain limit for real-time adjustment.
     ///
     /// Use this to dynamically modify the AGC's maximum allowable gain during runtime.
@@ -161,6 +198,7 @@ where
         Arc::clone(&self.absolute_max_gain)
     }
 
+    #[cfg(feature = "experimental")]
     /// Access the attack coefficient for real-time adjustment.
     ///
     /// Use this to dynamically modify how quickly the AGC responds to level increases.
@@ -171,6 +209,7 @@ where
         Arc::clone(&self.attack_coeff)
     }
 
+    #[cfg(feature = "experimental")]
     /// Access the release coefficient for real-time adjustment.
     ///
     /// Use this to dynamically modify how quickly the AGC responds to level decreases.
@@ -181,6 +220,7 @@ where
         Arc::clone(&self.release_coeff)
     }
 
+    #[cfg(feature = "experimental")]
     /// Access the AGC on/off control for real-time adjustment.
     ///
     /// Use this to dynamically enable or disable AGC processing during runtime.
@@ -199,6 +239,7 @@ where
     /// more accurately while maintaining smoother behavior for gradual changes.
     #[inline]
     fn update_peak_level(&mut self, sample_value: f32) {
+        #[cfg(feature = "experimental")]
         let attack_coeff = if sample_value > self.peak_level {
             self.attack_coeff
                 .load(Ordering::Relaxed)
@@ -206,6 +247,14 @@ where
         } else {
             self.release_coeff.load(Ordering::Relaxed)
         };
+
+        #[cfg(not(feature = "experimental"))]
+        let attack_coeff = if sample_value > self.peak_level {
+            self.attack_coeff.min(self.min_attack_coeff) // User-defined attack time limited via release_time
+        } else {
+            self.release_coeff
+        };
+
         self.peak_level = attack_coeff * self.peak_level + (1.0 - attack_coeff) * sample_value;
     }
 
@@ -225,11 +274,23 @@ where
     /// The peak level helps prevent sudden spikes in the output signal.
     #[inline]
     fn calculate_peak_gain(&self) -> f32 {
-        if self.peak_level > 0.0 {
-            (self.target_level.load(Ordering::Relaxed) / self.peak_level)
-                .min(self.absolute_max_gain.load(Ordering::Relaxed))
-        } else {
-            self.absolute_max_gain.load(Ordering::Relaxed)
+        #[cfg(feature = "experimental")]
+        {
+            if self.peak_level > 0.0 {
+                (self.target_level.load(Ordering::Relaxed) / self.peak_level)
+                    .min(self.absolute_max_gain.load(Ordering::Relaxed))
+            } else {
+                self.absolute_max_gain.load(Ordering::Relaxed)
+            }
+        }
+
+        #[cfg(not(feature = "experimental"))]
+        {
+            if self.peak_level > 0.0 {
+                (self.target_level / self.peak_level).min(self.absolute_max_gain)
+            } else {
+                self.absolute_max_gain
+            }
         }
     }
 
@@ -245,10 +306,19 @@ where
         let rms = self.update_rms(sample_value);
 
         // Compute the gain adjustment required to reach the target level based on RMS
+        #[cfg(feature = "experimental")]
         let rms_gain = if rms > 0.0 {
             self.target_level.load(Ordering::Relaxed) / rms
         } else {
             self.absolute_max_gain.load(Ordering::Relaxed) // Default to max gain if RMS is zero
+        };
+
+        // Compute the gain adjustment required to reach the target level based on RMS
+        #[cfg(not(feature = "experimental"))]
+        let rms_gain = if rms > 0.0 {
+            self.target_level / rms
+        } else {
+            self.absolute_max_gain // Default to max gain if RMS is zero
         };
 
         // Calculate the peak limiting gain
@@ -288,13 +358,25 @@ where
         };
 
         // Gradually adjust the current gain towards the desired gain for smooth transitions
-        self.current_gain = self.current_gain * attack_speed.load(Ordering::Relaxed)
-            + desired_gain * (1.0 - attack_speed.load(Ordering::Relaxed));
+        #[cfg(feature = "experimental")]
+        {
+            self.current_gain = self.current_gain * attack_speed.load(Ordering::Relaxed)
+                + desired_gain * (1.0 - attack_speed.load(Ordering::Relaxed));
 
-        // Ensure the calculated gain stays within the defined operational range
-        self.current_gain = self
-            .current_gain
-            .clamp(0.1, self.absolute_max_gain.load(Ordering::Relaxed));
+            // Ensure the calculated gain stays within the defined operational range
+            self.current_gain = self
+                .current_gain
+                .clamp(0.1, self.absolute_max_gain.load(Ordering::Relaxed));
+        }
+
+        #[cfg(not(feature = "experimental"))]
+        {
+            self.current_gain =
+                self.current_gain * attack_speed + desired_gain * (1.0 - attack_speed);
+
+            // Ensure the calculated gain stays within the defined operational range
+            self.current_gain = self.current_gain.clamp(0.1, self.absolute_max_gain);
+        }
 
         // Output current gain value for developers to fine tune their inputs to automatic_gain_control
         #[cfg(feature = "tracing")]
@@ -313,6 +395,7 @@ where
     type Item = I::Item;
 
     #[inline]
+    #[cfg(feature = "experimental")]
     fn next(&mut self) -> Option<I::Item> {
         self.input.next().map(|sample| {
             if self.is_enabled.load(Ordering::Relaxed) {
@@ -321,6 +404,12 @@ where
                 sample
             }
         })
+    }
+
+    #[inline]
+    #[cfg(not(feature = "experimental"))]
+    fn next(&mut self) -> Option<I::Item> {
+        self.input.next().map(|sample| self.process_sample(sample))
     }
 
     #[inline]
--- a/src/source/agc.rs
+++ b/src/source/agc.rs
@@ -183,6 +183,66 @@ where
     I: Source,
     I::Item: Sample,
 {
+    #[inline]
+    fn target_level(&self) -> f32 {
+        #[cfg(feature = "experimental")]
+        {
+            self.target_level.load(Ordering::Relaxed)
+        }
+        #[cfg(not(feature = "experimental"))]
+        {
+            self.target_level
+        }
+    }
+
+    #[inline]
+    fn absolute_max_gain(&self) -> f32 {
+        #[cfg(feature = "experimental")]
+        {
+            self.absolute_max_gain.load(Ordering::Relaxed)
+        }
+        #[cfg(not(feature = "experimental"))]
+        {
+            self.absolute_max_gain
+        }
+    }
+
+    #[inline]
+    fn attack_coeff(&self) -> f32 {
+        #[cfg(feature = "experimental")]
+        {
+            self.attack_coeff.load(Ordering::Relaxed)
+        }
+        #[cfg(not(feature = "experimental"))]
+        {
+            self.attack_coeff
+        }
+    }
+
+    #[inline]
+    fn release_coeff(&self) -> f32 {
+        #[cfg(feature = "experimental")]
+        {
+            self.release_coeff.load(Ordering::Relaxed)
+        }
+        #[cfg(not(feature = "experimental"))]
+        {
+            self.release_coeff
+        }
+    }
+
+    #[inline]
+    fn is_enabled(&self) -> bool {
+        #[cfg(feature = "experimental")]
+        {
+            self.is_enabled.load(Ordering::Relaxed)
+        }
+        #[cfg(not(feature = "experimental"))]
+        {
+            true
+        }
+    }
+
     #[cfg(feature = "experimental")]
     /// Access the target output level for real-time adjustment.
     ///
@@ -244,20 +304,10 @@ where
     /// more accurately while maintaining smoother behavior for gradual changes.
     #[inline]
     fn update_peak_level(&mut self, sample_value: f32) {
-        #[cfg(feature = "experimental")]
         let attack_coeff = if sample_value > self.peak_level {
-            self.attack_coeff
-                .load(Ordering::Relaxed)
-                .min(self.min_attack_coeff) // User-defined attack time limited via release_time
+            self.attack_coeff().min(self.min_attack_coeff) // User-defined attack time limited via release_time
         } else {
-            self.release_coeff.load(Ordering::Relaxed)
-        };
-
-        #[cfg(not(feature = "experimental"))]
-        let attack_coeff = if sample_value > self.peak_level {
-            self.attack_coeff.min(self.min_attack_coeff) // User-defined attack time limited via release_time
-        } else {
-            self.release_coeff
+            self.release_coeff()
         };
 
         self.peak_level = attack_coeff * self.peak_level + (1.0 - attack_coeff) * sample_value;
@@ -279,23 +329,10 @@ where
     /// The peak level helps prevent sudden spikes in the output signal.
     #[inline]
     fn calculate_peak_gain(&self) -> f32 {
-        #[cfg(feature = "experimental")]
-        {
-            if self.peak_level > 0.0 {
-                (self.target_level.load(Ordering::Relaxed) / self.peak_level)
-                    .min(self.absolute_max_gain.load(Ordering::Relaxed))
-            } else {
-                self.absolute_max_gain.load(Ordering::Relaxed)
-            }
-        }
-
-        #[cfg(not(feature = "experimental"))]
-        {
-            if self.peak_level > 0.0 {
-                (self.target_level / self.peak_level).min(self.absolute_max_gain)
-            } else {
-                self.absolute_max_gain
-            }
+        if self.peak_level > 0.0 {
+            (self.target_level() / self.peak_level).min(self.absolute_max_gain())
+        } else {
+            self.absolute_max_gain()
         }
     }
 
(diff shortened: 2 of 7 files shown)


dvdsk:
Since I do not feel fully qualified to review the algorithmic side of this, I am going to ask the wider rodio community for help in reviewing. I will also read up on AGC and see if I can find any problem with the implementation.

- @iluvcapra you are working on synthesizers and those looked pretty complicated, would you be willing to give this a look too?

src/source/agc.rs
```diff
@@ -0,0 +1,175 @@
+use super::SeekError;
+use crate::{Sample, Source};
+use std::time::Duration;
+
+/// Constructs an `AutomaticGainControl` object with specified parameters.
+///
```
dvdsk, line 6: for the novice like me, could you explain here the difference between gain, level and volume?

> Id suggest an attack_time of around 0.3 to 0.5 that seems to be a sweet spot for songs I tested.
> Also, target_level I adjusted to be 0.9 just to make sure it does not clip the audio; it was clipping for me at 1.0.

Its probably useful to suggest some values. Users can tweak them if they want they do not need to be perfect.

src/source/agc.rs
```diff
@@ -0,0 +1,175 @@
...
+
+impl<I> Iterator for AutomaticGainControl<I>
+where
+    I: Source,
+    I::Item: Sample,
+{
+    type Item = I::Item;
+
+    #[inline]
+    fn next(&mut self) -> Option<I::Item> {
```
dvdsk, line 78: a nitpick (Its fine not to change this) but I was wondering how you would feel about extracting some of this code into member functions. To me the first 3 comments almost beg to become a function name.

self.update_peak_level();
self.update_rms_level();
let gain = self.desired_gain();

src/source/mod.rs
```diff
@@ -232,6 +234,20 @@ where
         amplify::amplify(self, value)
     }
 
+    /// Applies automatic gain control to the sound.
```
dvdsk, line 237: I usually copy the documentation from the the source here (in this case agc::automatic_gain_control). This is what most users will see so the docs should be as extensive as at the source.

src/source/agc.rs
```diff
@@ -0,0 +1,175 @@
...
+            // Smoothly adjust current gain towards the target
+            let adjustment_speed = 0.05; // Balance between responsiveness and stability
+            self.current_gain = self.current_gain * (1.0 - adjustment_speed)
+                + (desired_gain * target_gain) * adjustment_speed;
+
+            // Constrain gain within predefined limits
+            self.current_gain = self.current_gain.clamp(0.1, self.absolute_max_gain);
+
+            // Uncomment for debugging:
+            println!("Current gain: {}", self.current_gain);
```
dvdsk, line 126: think we can do without these now :)

dvdsk (changes requested):
Very readable thanks! I have some minor things here and there. The algorithm makes sense to me though ill need to read up on AGC and look closer tomorrow. Did you come up with it yourself or did you find it somewhere? In either case we should attribute it in a comment somewhere.

Regarding the `to_32`. That seems like a fine solution to me. I am wondering though would the algorithm still work if you had only i16? We could implement subtract, add multiply_by_f32, div_by_usize and modulo for Sample.

That might save some conversions to f32.

src/source/agc.rs
UnknownSuperficialNight: ### Gain
Gain is the amount of amplification applied to an audio signal. It is typically adjusted at the input stage of an audio device, such as a microphone preamp. Increasing the gain boosts the signal strength, which is particularly useful for capturing quieter sounds.

### Volume
Volume describes the perceived loudness of sound to the listener. It is typically controlled by the output of an audio device, such as speakers or amplifiers. While gain and level are more technical terms related to signal processing, volume is about how loud the sound is in the environment. Adjusting the volume affects how we perceive the sound without changing the underlying signal strength.

## The key difference
The key difference is that gain is about the strength of the audio signal, while volume is about the perceived loudness of the sound. Gain is adjusted at the input, while volume is controlled at the output.

So here i'm adjusting the audio strength in order to boost the entire signal in order to be able to hear those quiet sounds or to be able to hear the entire track if the entire track is very low on gain

### Regarding level:
Level refers to the strength of an audio signal at various points in the audio chain. It can be measured in decibels (dB) and indicates how strong the signal is, whether it's at the input of a mixer, the output of a track, or any other point in the processing chain. Levels can be adjusted to ensure that different audio elements work well together without distortion.

src/source/agc.rs
UnknownSuperficialNight: That's a good idea thanks :)

Im working on it now and cleaning it up a bit + better comments

src/source/agc.rs
UnknownSuperficialNight: Ill leave them in for now so myself and others can test until everything is set in stone, then remove them before the final merge if thats ok?

UnknownSuperficialNight:
> I am wondering though would the algorithm still work if you had only i16? We could implement subtract, add multiply_by_f32, div_by_usize and modulo for Sample. There are trait in the std for all those I think.

Ill look into it; thanks :)

> Did you come up with it yourself or did you find it somewhere?

I put this together myself, but I definitely used some algorithms and ideas I found online. So while I wrote the code, I built on what others have done before.

src/source/agc.rs
```diff
@@ -77,55 +144,41 @@ where
...
 
-            // Choose the more conservative gain adjustment
+            // Select the lower of peak and RMS gains to ensure conservative adjustment
             let desired_gain = peak_gain.min(rms_gain);
 
-            // Set target gain to the middle of the allowable range
-            let target_gain = 1.0; // Midpoint between 0.1 and 3.0
-
-            // Smoothly adjust current gain towards the target
-            let adjustment_speed = 0.05; // Balance between responsiveness and stability
```
UnknownSuperficialNight: Setting adjustment_speed to self.attack_time may cause issues with the algorithm; I couldn't hear or detect anything wrong, but it's something to look out for.

src/source/agc.rs
```diff
@@ -77,55 +144,41 @@ where
...
-                1.0
+                1.0 // Default to unity gain if RMS is zero to avoid division by zero
             };
 
-            // Choose the more conservative gain adjustment
+            // Select the lower of peak and RMS gains to ensure conservative adjustment
             let desired_gain = peak_gain.min(rms_gain);
 
-            // Set target gain to the middle of the allowable range
-            let target_gain = 1.0; // Midpoint between 0.1 and 3.0
```
UnknownSuperficialNight: Removed `target_gain` since it seems redundant. However, it's a potential area to monitor for any issues that may arise that I couldn't identify/find.

src/source/mod.rs
UnknownSuperficialNight: Thanks i added docs there

UnknownSuperficialNight:
Integrated it with my GUI earlier, and it's working really well—Tested on a -10db audio file. Here’s a video showing the results: 

(**Left is using sink without AGC**)
(**Right is using sink with AGC**)

Video:


https://github.com/user-attachments/assets/77691699-be3d-41fd-af7d-4946d95d45ed

dvdsk:
> It could be a premature optimization

you got a good point there. It might very well be, its a real bummer rodio has no benchmarks/profile runs setup. We should not optimize without benchmark or profile.

dvdsk:
> Where?

maybe the top of the source file? If you can list your inspiration and put your name on it, you made an algorithm for AGC you deserve some credit for that. Something along the likes of:

```
// Automatic gain control by @UnknownSuperficialNight, 
// this is a combination of <concept> other <concept> it builds upon 
// the work of <name> as seen here <blog/book/something>.
```

Its entirely optional if you do not feel like it you can leave this out.

dvdsk:
> Integrated it with my GUI earlier, and it's working really well—Tested on a -10db audio file. Here’s a video showing the results:

Great to hear! unfortunately the video is not playing for me (neither in browser or downloaded in VLC).

dvdsk (changes requested):
Honestly, this is the most well documented code I have seen in a long while :heart:. I am gonna show this to my friends. If I where you I would be really proud of myself. 

The only thing that could be improved is the `--debug-gain` flag, I think checking an argument as a library is sup-optimal.

Your comment there mentions its for both debugging and monitoring. Do you mean its useful for the end user to be able to see the number? For example so they can tweak the parameters better?

If that's true then maybe a callback would be a good idea. Automatic gain control could get an extra argument (maybe an option?). 

It would look like this then:
```rust
    fn automatic_gain_control(
        self,
        target_level: f32,
        attack_time: f32,
        absolute_max_gain: f32,
        gain_monitor: Option<Box<dyn Fn(f32)>>,
    ) -> AutomaticGainControl<Self>
```

The current monitoring code in `fn next` could then become:
```rust
if let Some(callback) = self.gain_monitor {
    callback(self.current_gain);
}
```

I have used `Box<dyn Fn>` to prevent AGC getting a generic argument. That will incur a performance cost when using monitoring, its a tradeoff. What we could do is add a second source: `MonitoredAutomaticGainControl` which is generic over the monitor callback. Then implement the current `AutomaticGainControl` by just calling `MonitoredAutomaticGainControl` with as callback a function that does nothing. The optimizer will optimize that function out for us.

Regarding logging for debugging purposes. Its probably a good idea to add either tracing or log to Rodio. We can enable them using a compile time feature so they are zero cost for any user that does not need them. They also integrate nicely in whatever application someone is writing.

dvdsk:
I should really add some benchmarking to Rodio. AGC is now running for every sample, might be an idea to make it run only once every n samples to lower the perf hit of AGC. But that is definitely not for now but for a future PR/issue once we have benchmarking integrated into Rodio's CI.

UnknownSuperficialNight:
> might be an idea to make it run only once every n samples to lower the perf hit of AGC

Running every n samples could work, but we risk issues with the audio not being able to adjust when needed. For example, a bass drop or a loud kick would possibly have a gain that is too high, making a crackling sound.

Instead, we can make a threshold that logs the previous values and detects a large enough change.


**This is my testing while developing (AGC responsiveness threshold)**
```
Performance results from testing based on cpu usage:
Every single thread (baseline):
1.4% to 4.2% of 1 core (avg 2.8%-4.2%)

Using a advanced detection method:
1.4% to 4.2% of 1 core (avg 2.8%)

Using updated and simplified advanced detection method:
1.4% to 4.2% of 1 core (same as above but no perceivable change in audio)
```

Although it most of the time indicates a 2.8% cpu usage, it's still such a small difference that I don't know if it's worth it. Plus, this is all using one core. By that, I mean 100% equals one core, and 200% equals two cores fully maxed out. In this case, we are using 1/36th of a core on average. So, maybe it's not worth it; the more delay we have, the more potential audio issues we will encounter, as it might not be able to react in time to lower it enough reasonably.

In any case here is the code for that if we want to explore it in the future
```rust
/// Checks if the peak level change is significant enough to trigger a gain update
///
/// This method calculates the absolute difference between the current peak level
/// and the last significant peak level. It considers a change to be significant
/// if this difference is 0.01 or greater, which is equivalent to a 1% change
/// in the signal level.
#[inline]
fn is_significant_change(&self) -> bool {
// Calculate the absolute difference directly using f32
let difference = (self.peak_level - self.last_significant_peak).abs();

// Consider a change significant if it's 0.01 or more (1% change)
 difference >= 0.01
}
```

UnknownSuperficialNight:
> I should really add some benchmarking to Rodio

definitely a good idea

> Honestly, this is the most well documented code I have seen in a long while ❤️. I am gonna show this to my friends. You have set a high bar for all the other PR's now :)

Thank you so much! That really means a lot to me☺️

> The only thing that could be improved is the --debug-gain flag, I think checking an argument as a library is sup-optimal.
  Your comment there mentions its for both debugging and monitoring. Do you mean its useful for the end user to be able to see the number? For example so they can tweak the parameters better?
  If that's true then maybe a callback would be a good idea. Automatic gain control could get an extra argument (maybe an option?).
  
Originally, it was just so I could view the changes while developing. However, now that I've refined the algorithm, it's become quite handy to read the values visually. This way, you don't have to guess using your ears. Realizing this, I'm planning to add a toggle for it, so we don't need to parse and loop over all arguments.

> maybe the top of the source file? If you can list your inspiration and put your name on it, you made an algorithm for AGC you deserve some credit for that. Something along the likes of:

I don’t mind. ☺️ However, to clarify, my AGC algorithm is inspired by various online resources related to gain control concepts, including discussions on the RMS algorithm. While I didn't base it on a specific algorithm, I synthesized ideas from multiple sources to create my implementation.

UnknownSuperficialNight:
Force pushed so i don't add redundant commits to the history as i just merged the commits together

UnknownSuperficialNight:
> I have used Box<dyn Fn> to prevent AGC getting a generic argument. That will incur a performance cost when using monitoring, its a tradeoff. What we could do is add a second source: MonitoredAutomaticGainControl which is generic over the monitor callback. Then implement the current AutomaticGainControl by just calling MonitoredAutomaticGainControl with as callback a function that does nothing. The optimizer will optimize that function out for us.

This is a good idea but i wonder since our users will only do this once probably in debug mode why not just do

```rust
// Output current gain value for debugging purposes
#[cfg(debug_assertions)]
println!("Current gain: {}", self.current_gain);
```

Then it only compiles in debug mode but the downside here is that it will always compile in debug_mode

for now ill go with your suggestion as its most flexible and if we want we can revert to #[cfg(debug_assertions)]

dvdsk:
> Although it most of the time indicates a 2.8% cpu usage, it's still such a small difference that I don't know if it's worth it.

I completely agree, thats not worth the effort and increased code complexity :+1:

dvdsk:
> However, the downside is that if we do call `automatic_gain_control`, it will always compile the println in debug mode. What do you think?

If anyone needs access to the value via a callback they can always open an issue, and then its easy enough to add. Always printing in debug mode seems like a bad idea, imagine having a tui app and in debug mode the view gets corrupted :sweat_smile:. But the idea of determining compile time whether to print it is good.

I see two options there (maybe there are more):
- add a feature to rodio, something like "print-agc-gain". Would still mess up tui-apps.
- integrate [tracing](https://docs.rs/tracing/latest/tracing/#) or [log](https://docs.rs/log/latest/log/#) into rodio. They are what the rust community has standardised on. They can be disabled using a compile-time-feature, and allow the app developer to determine where the log messages go (stdout/file/some online log service). They can even be configured at runtime whether to display logs from a certain dependency.

It seems best to go for the second option, though that is out of the scope for this PR, ill see if I can quickly add tracing to rodio.

UnknownSuperficialNight:
> It seems best to go for the second option, though that is out of the scope for this PR, ill see if I can quickly add tracing to rodio.

Yea that seems best also look into [env_logger](https://docs.rs/env_logger/latest/env_logger) it's integrated directly with log and allows control over logging through environment variables

Like this:
`DEBUG_AUDIO=1 cargo run`

dvdsk:
> Clarification I've never used any of these crates before, but I've heard of env_logger and the others.

tracing builds upon log and env_logger, its a bit heavier to compile but I think that is okay since you have to opt into it. These crates are always split into two parts: the core that only gives you the `println` equivalent and the part that uses those macro's to actually log somewhere. The app developer should use the latter and the libs the former.

I have added tracing on main now, see 95a466e03a7997ce91b114dc1062e48c5a306756 on how to use it. And to see those traces in your tests/application you will want to do something like this: https://github.com/tokio-rs/tracing/blob/master/examples/examples/fmt.rs. For logging in AGC I would suggest `tracing::trace("some msg/status")`.

UnknownSuperficialNight:
Done added it in check if its ok

idk if i should add a msg or documentation somewhere to use tracing_subscriber like this
```rust

use tracing_subscriber::fmt::format::{FmtSpan, Format};

tracing_subscriber::fmt()
        .event_format(Format::default().compact())
        .without_time()
        .with_target(false)
        .with_span_events(FmtSpan::NONE)
        .with_env_filter("rodio=debug")
        .init();
```

call like this `RUST_LOG=debug cargo run`

The reason I say this is because the default tracing_subscribers is so verbose that when using RUST_LOG=debug, it causes a delay in AGC that causes it to cut out audio constantly.

Here is a video showing that:

**Using default init of tracing_subscriber**

https://github.com/user-attachments/assets/d76143a3-313c-4b39-833a-fd66ee61aeb2

**Using tracing_subscriber::fmt**

https://github.com/user-attachments/assets/98681cdc-d493-4de8-a038-75dfefee9a81

src/source/mod.rs
```diff
@@ -241,15 +241,28 @@ where
     ///
     /// # Parameters
```
UnknownSuperficialNight, line 242: From some research the documentation for docs.rs will be generated of these comments correct?

UnknownSuperficialNight:
> The only thing that could be improved is the --debug-gain flag

Wrong :)

New improvements added in the latest commit:

Replaced Vec-based RMS calculation with CircularBuffer:
- Allocates a few hundred more bytes on the stack but eliminates heap usage
- Prioritizes lower latency over small memory increase
- Should improve CPU performance
- CircularBuffer provides O(1) updates and mean calculations

Adjusted RMS algorithm:
- Now relies more on RMS than peaks compared to previous version
- Allows for smoother audio while still preventing clipping
- Uses peaks to ensure audio never clips
- Should result in smoother overall audio output

dvdsk:
(I have not looked at the new changes yet)

I am working on adding benchmarks, and I notice that some sources accept any Sample type while others want f32. I would like to unify that in the future. As long as AGC has to cast samples to f32 it might be best to just demand that the Sample is f32 in the bounds. (see high_pass source as an example of this).

There should probably be a contributors guide that mentions stuff like this... ill add it to my todo :)

dvdsk:
>As long as AGC has to cast samples to f32 it might be best to just demand that the Sample is f32 in the bounds. (see high_pass source as an example of this).

Okay I really do not know whats best now, I just noticed amplify using `Sample::amplify` but thats just `u16 as f32 * value`. I'll add a benchmark on the cost of a conversion like that. Then get back to you. Its probably basically free, on the other hand there is the `sample_converter` `Source` must be there for a reason.

UnknownSuperficialNight:
> There should probably be a contributors guide that mentions stuff like this... ill add it to my todo :)

Yea a decent contributors guide can go a long way

> Okay I really do not know what's best now, I just noticed amplify using Sample::amplify but that's just u16 as f32 * value. I'll add a benchmark on the cost of a conversion like that. Then get back to you. Its probably basically free, on the other hand there is the sample_converter Source must be there for a reason.

Yea probably costs nothing but unification is probably a good idea in the long run for simplicity

dvdsk:
> The reason I say this is because the default tracing_subscribers is so verbose that when using RUST_LOG=debug, it causes a delay in AGC that causes it to cut out audio constantly.

Thats with a release build? Kinda suprises me tbh. Yeah adding such a not is worth it even if its only on debug builds.

src/source/mod.rs
dvdsk: yes rust is great like that :)

You can also open/generate the docs locally (will check some things for you too) using: `cargo doc --open`.

UnknownSuperficialNight:
> Thats with a release build? Kinda suprises me tbh. Yeah adding such a not is worth it even if its only on debug builds.

Nope thats on a debug build on release the init it is ok

dvdsk:
> New improvements added in the latest commit:

Very nice, ill have a look :)

Its a pitty I did not find time to do the benchmarks earlier. They are done now though :+1:. You can merge them in from main and add your own in `benches/effects.rs` It should be pretty easy, just copy the `high_pass` bench.

You run the benchmarks with `cargo bench`.

From what I can see the conversions to f32 are, as expected, basically free:

```
conversions     fastest       │ slowest       │ median        │ mean          │ samples │ iters
╰─ from_i16_to                │               │               │               │         │
   ├─ f32       254.9 µs      │ 487.7 µs      │ 256.5 µs      │ 266.1 µs      │ 100     │ 100
   ├─ i16       189.8 µs      │ 412.9 µs      │ 191.1 µs      │ 198.5 µs      │ 100     │ 100
   ╰─ u16       245.3 µs      │ 435.8 µs      │ 247.3 µs      │ 257.8 µs      │ 100     │ 100

```

dvdsk:
next review will have to wait until after the weekend. Have a great sunday btw :+1:

src/source/mod.rs
UnknownSuperficialNight: Thanks working on making the styling better for the method automatic_gain_control

UnknownSuperficialNight:
> next review will have to wait until after the weekend. Have a great sunday btw 👍

Ok check u then ill be here refining it more 👋

UnknownSuperficialNight:
When you get back have a look at this let me know if its to verbose or if there should be any tweaks

UnknownSuperficialNight:
I can't think of any other improvements I could make to the code; so far, the current state is near final.

dvdsk:
> I figured id add a tldr then a full explanation for those who want it

ohh nice idea, love it! :+1:

dvdsk:
> I don't know if you can hear the little clip when the beat drops

I asked a friend to listen in, we both did not hear it. So its not an issue for us, I would say keep it as is. If someone has a problem with it they are of course free to add that look-ahead buffer as a variation on the current AGC (maybe with some generics).

src/source/agc.rs
```diff
@@ -0,0 +1,345 @@
...
+struct CircularBuffer {
+    buffer: [f32; RMS_WINDOW_SIZE],
+    index: usize,
+    sum: f32,
+}
+
+impl CircularBuffer {
+    /// Creates a new CircularBuffer with a fixed size determined at compile time.
+    ///
+    /// The `_size` parameter is ignored as the buffer size is set by `RMS_WINDOW_SIZE`.
```
dvdsk, line 61: then why is it there? :)

src/source/agc.rs
```diff
@@ -0,0 +1,345 @@
...
+    is_enabled: Arc<AtomicBool>,
+}
+
+/// A circular buffer for efficient RMS calculation over a sliding window.
+///
+/// This structure allows for constant-time updates and mean calculations,
+/// which is crucial for real-time audio processing.
+#[derive(Clone, Debug)]
+struct CircularBuffer {
+    buffer: [f32; RMS_WINDOW_SIZE],
```
dvdsk, line 53: 32Kb is a rather large allocation on the stack. I am kinda surprised that clippy is not complaining here. Might be better to heap allocate this (vec<f32>). You could also try replacing buffer+index in `CircularBuffer` with [`VecDeque`](https://doc.rust-lang.org/std/collections/struct.VecDeque.html#). But that might have some overhead. Since you added a benchmark we can measure that now :stuck_out_tongue_closed_eyes:.

src/source/agc.rs
```diff
@@ -0,0 +1,345 @@
...
+        let old_value = self.buffer[self.index];
+        self.buffer[self.index] = value;
+        self.sum += value - old_value;
+        self.index = (self.index + 1) % self.buffer.len();
+        old_value
+    }
+
+    /// Calculates the mean of all values in the buffer.
+    ///
+    /// This operation is O(1) due to the maintained running sum.
```
dvdsk, line 85: :tada: :partying_face:

src/source/agc.rs
```diff
@@ -0,0 +1,345 @@
...
+///
+/// # Arguments
+///
+/// * `input` - The input audio source
+/// * `target_level` - The desired output level
+/// * `attack_time` - Time constant for gain increase
+/// * `release_time` - Time constant for gain decrease
+/// * `absolute_max_gain` - Maximum allowable gain
+#[inline]
+pub fn automatic_gain_control<I>(
```
dvdsk, line 102: This should be pub(crate) I am pretty sure this is wrong in most rodio sources, see: https://github.com/RustAudio/rodio/issues/622 for more info.

src/source/agc.rs
```diff
@@ -0,0 +1,345 @@
...
+        input,
+        target_level,
+        absolute_max_gain,
+        current_gain: 1.0,
+        attack_coeff: (-1.0 / (attack_time * sample_rate as f32)).exp(),
+        release_coeff: (-1.0 / (release_time * sample_rate as f32)).exp(),
+        min_attack_coeff: release_time,
+        peak_level: 0.0,
+        rms_window: CircularBuffer::new(RMS_WINDOW_SIZE),
+        is_enabled: Arc::new(AtomicBool::new(true)),
```
dvdsk, line 125: I see you added 'dynamic' controls in there. The current way we do controls in rodio is through the sink. @PetrGlad had the idea to try and do this with Arcs + Atomics just like you do here.  I have no idea what the performance impact of this is, its something I have been discussing in https://github.com/RustAudio/rodio/issues/619 and which will move to its own issue in the future. 

For now its a bit strange that `enabled` can be changed via a handle (gotten through `get_agc_control`) but the other variables can not. @PetrGlad AGC might be a good candidate for benchmarking dynamic controls (though atomics+arcs) since it has so many variables.

src/source/agc.rs
UnknownSuperficialNight: Oops, that's not supposed to be there that was originally used I commented it out and forgot about it

src/source/agc.rs
UnknownSuperficialNight: Tried VecDeque added 10ms, upped mean by 4ms

image

dvdsk: and putting the buffer on the heap with Box or Vec? Does that do the same?

UnknownSuperficialNight: Fixed in this [commit](https://github.com/RustAudio/rodio/pull/621/commits/86cb156e47bb0165a8463e75c70b284903189337)

src/source/agc.rs
UnknownSuperficialNight: Thanks didn't know pub(crate) existed, fixed now

src/source/agc.rs
UnknownSuperficialNight: > and putting the buffer on the heap with Box or Vec? Does that do the same? Did you pre-allocate the VecDeque during source creation? (filling it up with zero's)?

Sorry didn't see the reply before the commit yea I did not pre-allocate the VecDeque

here is pre alloc VecDeque
├─ agc_enabled   13.83 ms      │ 15.11 ms      │ 13.91 ms      │ 13.95 ms      │ 100     │ 100

here is using the commit I made with box
├─ agc_enabled   8.682 ms      │ 12.19 ms      │ 9.479 ms      │ 9.564 ms      │ 100     │ 100

still a quite bit worse

UnknownSuperficialNight:
> Then there is a stack size issue we might wanna adress

Fixed :)

> Finally we should add AGC to the sink so everyone can easily enable it provided that AGC has hardly any preformance impact once disabled

It's basically free
```
effects          fastest       │ slowest       │ median        │ mean          │ samples │ iters
├─ agc_disabled  484.5 µs      │ 960 µs        │ 585.4 µs      │ 622.6 µs      │ 100     │ 100
```

> For now its a bit strange that enabled can be changed via a handle (gotten through get_agc_control) but the other variables can not. @PetrGlad AGC might be a good candidate for benchmarking dynamic controls (though atomics+arcs) since it has so many variables.

True, I made them dynamic, but it makes the code a lot messier as I need to use `Arc<AtomicU32>`

that means iteractions with it look like this
```rust
    // Set the absolute maximum gain limit
    let hi = agc_source.set_absolute_max_gain();
    hi.store(10.0_f32.to_bits(), std::sync::atomic::Ordering::Relaxed);
```

as ill need to use f32::from_bits for everything to conver from AtomicU32 to f32

src/source/agc.rs
UnknownSuperficialNight: > idea to try and do this with Arcs + Atomics just like you do here. I have no idea what the performance impact of this is

Almost negligible especially with atomics it almost is around the same level of performance as an if statement or a conversion to f32 i.e. in µs/micro seconds

but more than welcome to benchmark it somehow

src/source/agc.rs
```diff
@@ -74,9 +72,11 @@ impl CircularBuffer {
     #[inline]
     fn push(&mut self, value: f32) -> f32 {
         let old_value = self.buffer[self.index];
+        // Update the sum by first subtracting the old value and then adding the new value; this is more accurate.
+        self.sum = self.sum - old_value + value;
         self.buffer[self.index] = value;
-        self.sum += value - old_value;
-        self.index = (self.index + 1) % self.buffer.len();
+        // Use bitwise AND for efficient index wrapping since RMS_WINDOW_SIZE is a power of two.
```
dvdsk, line 79: pretty sure the compiler will think of that :)
if your unsure benchmark it?

UnknownSuperficialNight: Did benchmark it, it's a little bit better with this, plus this makes sure the compiler is doing the optimization.

UnknownSuperficialNight:
Just tested playback while compiling other things on my system forcing 100% CPU usage here are the results

```
Timer precision: 20 ns
effects          fastest       │ slowest       │ median        │ mean          │ samples │ iters
├─ agc_disabled  727.4 µs      │ 2 ms          │ 1.014 ms      │ 1.073 ms      │ 100     │ 100
├─ agc_enabled   8.752 ms      │ 25.29 ms      │ 12.92 ms      │ 12.43 ms      │ 100     │ 100
├─ amplify       628.9 µs      │ 2.323 ms      │ 1.033 ms      │ 1.048 ms      │ 100     │ 100
├─ fade_out      5.324 ms      │ 7.242 ms      │ 5.477 ms      │ 5.541 ms      │ 100     │ 100
├─ high_pass     2.911 ms      │ 5.878 ms      │ 3.055 ms      │ 3.134 ms      │ 100     │ 100
╰─ reverb        19.95 ms      │ 43.81 ms      │ 25.95 ms      │ 27.12 ms      │ 100     │ 100
```

Still has smooth playback when compiled on `--release`

src/source/agc.rs
dvdsk: well since your experimenting (and having fun) anyways :)

I suspect that performance impact (9.5 vs 7.6 is a significant increase) does not come from the fact that we are using atomics but the Arc's. Where all these controls where in the same struct next to each-other now they are spread over the heap. That makes the optimizer struggle to optimize more and its difficult for the cpu to keep everything in registers/L1 cache.

So maybe we can adress them by using only one Arc over a struct of atomics?

dvdsk: On the other hind I kinda wanna merge this without any of the atomic/arc stuff (we need to hammer out a clear API and apply that to everything and this PR is not the place for that). So if you want to experiment be my guest :) but to merge it we'll need it plain and boring again.

src/source/agc.rs
dvdsk: oh you did, well now im impressed by your thoroughness and disappointed in the compiler :sweat_smile:

dvdsk:
> Let me know how you feel about adding [atomic_float](https://docs.rs/atomic_float/latest/atomic_float/index.html) as it makes the implementation a lot faster and also could be used by others in other parts of the program where they need a performant AtomicF32 (load/store) are basically free. Could be useful in relation to #624 and the [code](https://github.com/thomcc/atomic_float/blob/main/src/atomic_f32.rs) for atomic_float is quite small

See: https://github.com/RustAudio/rodio/pull/621#discussion_r1782421846
For now I would like to keep to using the Sink for all control. Then once we hammer out a good API we can modify AGC to allow for controls. Lets first get this merged. The PR is big enough as is :)

UnknownSuperficialNight:
> On the other hind I kinda wanna merge this without any of the atomic/arc stuff (we need to hammer out a clear API and apply that to everything and this PR is not the place for that). So if you want to experiment be my guest :) but to merge it we'll need it plain and boring again.

For me, the AtomicF32 does not matter for my use case however the atomicbool to disable and enable it dynamically I do need, and I need it to be able to be passed around which only an atomicbool can provide me in this case as I need to access it from outside the scope of sink in my own program.

dvdsk:
> for my use case however the atomicbool to disable and enable it dynamically I do need, and I need it to be able to be passed around which only an atomicbool can provide me in this case as I need to access it from outside the scope of sink in my own program.

Sink not being clone is annoying. We are going to adress the ergonimics of controlling sources for sure. Until then would wrapping the AGC (in your program) in a [`PeriodicAccess`](https://docs.rs/rodio/latest/rodio/source/struct.PeriodicAccess.html) work? The callback would be something along the lines of: 
```rust
|inner| inner.enabled =  shared_atomic_bool.load(Relaxed);
```
`shared_atomic_bool` can then be freely moved around.

UnknownSuperficialNight:
> alternatively we could put it under a compile time feature experimental and clearly document that feature (in lib.rs) as unstable and that it will break once the controls API lands. Then you can use it right now without having to go through PeriodicAccess.

This works for me then we are both happy should I put (get_release_coeff, get_attack_coeff, get_absolute_max_gain, get_target_level) as well under `experimental` as well future can tweak it for the new API when it lands then it's easier to adapt in the future and people can still use it now if they want?

dvdsk:
> `experimental`

lets just make it `experimental` then we can add other things under the same name later and reuse the readme section that will explain the risks of using experimental (mainly the feature can and will probably dissapear/break or change at any time)

dvdsk:
> instead, I can just make it optional and only enabled via the experimental flag

:+1:

dvdsk:
> If this is all good, ill finish the commit in around 5 min

take your time, there is no reason to rush.

Oh you got to merge main btw, CHANGELOG.md got updated by the merge of the syntesizer sources.

UnknownSuperficialNight:
Should be all done and ready to merge. 

This is my second ever PR and my first ever code PR, so I'm happy it's gone well.

src/source/agc.rs
```diff
@@ -0,0 +1,462 @@
...
+        let sample_value = sample.to_f32().abs();
+
+        // Dynamically adjust peak level using an adaptive attack coefficient
+        self.update_peak_level(sample_value);
+
+        // Calculate the current RMS (Root Mean Square) level using a sliding window approach
+        let rms = self.update_rms(sample_value);
+
+        // Compute the gain adjustment required to reach the target level based on RMS
+        #[cfg(feature = "experimental")]
```
dvdsk, line 314: lots of #[cfg(not(feature = "experimental"))] makes the algorithm a lot less readable. Have you thought about adding getters? instead of `self.target_level.load(Ordering::Relaxed) / rms` vs `self.target_level / rms` you would have one: `self.target_level() / rms`. Then the getter contains the "noisy" feature stuff.

src/source/agc.rs
```diff
@@ -0,0 +1,462 @@
...
+    input: I,
+    target_level: Arc<AtomicF32>,
+    absolute_max_gain: Arc<AtomicF32>,
+    current_gain: f32,
+    attack_coeff: Arc<AtomicF32>,
+    release_coeff: Arc<AtomicF32>,
+    min_attack_coeff: f32,
+    peak_level: f32,
+    rms_window: CircularBuffer,
+    is_enabled: Arc<AtomicBool>,
```
dvdsk, line 58: is_enabled is useful without an atomic too! its needed for the `periodic_callback` method of controlling sources.

dvdsk (commented):
I would do the cfg experimental differently, and I think we still need an enable/disable function for the non atomicbool approach is missing. the rest looks perfect

benches/effects.rs
```diff
@@ -46,3 +46,43 @@ fn amplify(bencher: Bencher) {
...
+                    4.0,   // attack_time (in seconds)
+                    0.005, // release_time (in seconds)
+                    5.0,   // absolute_max_gain
+                )
+                .for_each(divan::black_box_drop)
+        })
+}
+
+#[cfg(feature = "experimental")]
+// Do note you will need to pass --features experimental to run this benchmark
```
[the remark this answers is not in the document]
UnknownSuperficialNight: Yes, but if you don't explicitly pass `--features experimental` it won't run even if you pass it in the cargo.toml

dvdsk:
> This is my second ever PR and my first ever code PR, so I'm happy it's gone well. 🙂

and its a big one 600+ lines

> Btw, if you're working on remaking the sink API, I don't mind helping if you'd have me

whats needed most is ideas for API design. I have not used rodio a ton myself (ironic isnt it). So I am not familiar with all usecases.

src/source/agc.rs
UnknownSuperficialNight: Good idea working on it

benches/effects.rs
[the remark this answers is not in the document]
dvdsk: I do not know exactly what you mean here with pass it in the cargo.toml. But benchmarks should not run when using rodio as a dependency.

UnknownSuperficialNight:
> and its a big one 600+ lines

Don't know if you're saying that's a good or bad thing

> whats needed most is ideas for API design. I have not used rodio a ton myself (ironic isnt it). So I am not familiar with all usecases.

I use it quite a lot in my music player GUI, in fact I've built around it

benches/effects.rs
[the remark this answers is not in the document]
UnknownSuperficialNight: My bad got confused with something completely irrelevant, ill fix this

dvdsk:
> Don't know if you're saying that's a good or bad thing

well it means it was difficult/a lot of work most ppl start with a tiny PR :), you went big. So thats why it took a while.

UnknownSuperficialNight:
> well it means it was difficult/a lot of work most ppl start with a tiny PR

Well I learned a lot from it :smiley:

Btw, ready for another review.

src/source/agc.rs
```diff
@@ -183,6 +183,66 @@ where
...
     I::Item: Sample,
 {
+    #[inline]
+    fn target_level(&self) -> f32 {
+        #[cfg(feature = "experimental")]
+        {
+            self.target_level.load(Ordering::Relaxed)
+        }
+        #[cfg(not(feature = "experimental"))]
+        {
```
dvdsk, line 195: I do not think these `{}` are necessary, does it work without?

dvdsk (changes requested):
otherwise ready to merge

src/source/agc.rs
UnknownSuperficialNight: It does not work without it, I get errors with it wanting me to append ; to the end of the experimental one so for consistency I made them both the same easier to read in the future since one is needed anyway

dvdsk: might be a rust analyzer issue. W'll merge and then ill play around with it a little :)

dvdsk:
hype :+1:, loving having this in gonna be great for my podcast app, once I write it :)

UnknownSuperficialNight:
> hype 👍, loving having this in gonna be great for my podcast app, once I write it :)

Yay, its finally merged :relieved:

PR #630, opened 2024-10-05
Issue#566

ugochukwu-850:
In this PR,
I have made changes in the speed.rs module . Explaining how the entire module works 
I have also added a note on the set speed function - Specifying there would be an increase of pitch

I have also removed the note comment from @dvdsk on the try_seek method : As the reason for the commit is now fixed

### DIFF
--- a/src/source/mod.rs
+++ b/src/source/mod.rs
@@ -421,7 +421,7 @@ where
     /// 2. **Change in the speed would affect your total duration inversely**
     /// - if you set the speed by 0.5,  your total duration would be (2x the original total duration) longer.
     /// - Also if you set the speed to 2 the total duration would be (0.5 the original total_duration) shorter
-    /// 
+    ///
     /// See [`Speed`] for details
     #[inline]
     fn speed(self, ratio: f32) -> Speed<Self>
--- a/src/source/mod.rs
+++ b/src/source/mod.rs
@@ -413,11 +413,6 @@ where
 
     /// Changes the play speed of the sound. Does not adjust the samples, only the play speed.
     /// Creates a Speed struct that handles the speed control
-    ///
-    ///  In order to speed up a source, the speed struct:
-    /// - Increases the current sample rate by the given factor
-    /// - Updates the total duration function to cover for the new factor by dividing by the factor
-    /// - Update the pos function by multiplying the position by the factor
     #[inline]
     fn speed(self, ratio: f32) -> Speed<Self>
     where
--- a/src/source/mod.rs
+++ b/src/source/mod.rs
@@ -412,7 +412,17 @@ where
     }
 
     /// Changes the play speed of the sound. Does not adjust the samples, only the play speed.
-    /// Creates a Speed struct that handles the speed control
+    ///
+    ///  Creates a [`Speed`] struct that handles the speed control
+    /// #### Note:
+    /// 1. **Increasing the speed would also increase the pitch by the same factor**
+    /// - If you increased set the speed to 0.5,  the frequency would be slower (0.5x the original frequency) .
+    /// - Also if you set the speed to 1.5 the frequency would be faster ( 1.5x the original frequency).
+    /// 2. **Change in the speed would affect your total duration inversely**
+    /// - if you set the speed by 0.5,  your total duration would be (2x the original total duration) longer.
+    /// - Also if you set the speed to 2 the total duration would be (0.5 the original total_duration) shorter
+    /// 
+    /// See [`Speed`] for details
     #[inline]
     fn speed(self, ratio: f32) -> Speed<Self>
     where
--- a/src/source/speed.rs
+++ b/src/source/speed.rs
@@ -40,7 +40,7 @@
 //! std::thread::sleep(std::time::Duration::from_secs(5));
 //! ```
 //! Notice the increase in pitch as the factor increases
-//! 
+//!
 //! Since the samples are played faster the audio wave get shorter increasing their frequencies
 
 use std::time::Duration;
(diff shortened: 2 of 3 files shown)


dvdsk:
super! two small things, you need to run `cargo fmt` to fix the formatting and I would suggest moving some of the documentation to `source/mod.rs` specifically add it to `fn speed`. Since that is what most ppl will see.

ugochukwu-850:
@dvdsk , Ohh I forgot to run cargo fmt and yes your right . I should probably move the docs to the mod file, so it's more accessible.

Expect the commit, any hour 24hrs from now . I was just so busy enjoying the weekend . 😅.

ugochukwu-850:
Alright , I think this is good @dvdsk , please review and merge.

src/source/speed.rs
```diff
@@ -1,3 +1,50 @@
+//! Playback Speed control Module.
+//!
+//! The main concept of this module is the [`Speed`] struct, which
+//! encapsulates playback speed controls of the current sink.
+//!
+//! In order to speed up a sink, the speed struct:
+//! - Increases the current sample rate by the given factor
+//! - Updates the total duration function to cover for the new factor by dividing by the factor
+//! - Update the pos function by multiplying the position by the factor
```
dvdsk, line 9: I think you mean to describe how try_seek works for a sped up source here

src/source/speed.rs
```diff
@@ -1,3 +1,50 @@
...
+//! .take_duration(Duration::from_secs_f32(20.25))
+//! .amplify(0.20);
+//!
+//! //! let sink = Sink::try_new(&stream_handle)?;
+//! sink.set_speed(2.0);
+//! sink.append(source);
+//! std::thread::sleep(std::time::Duration::from_secs(5));
+//! ```
+//! Notice the increase in pitch as the factor increases
+//! This is due to the higher audio data stream increasing the frequency
```
dvdsk, line 46: I would say: "Since the samples are played faster the audio wave get shorter increasing their frequencies."

src/source/speed.rs
```diff
@@ -1,3 +1,50 @@
...
+//! // Decode that sound file into a source
+//! let source = Decoder::new(file).unwrap();
+//! // Play the sound directly on the device 2x faster
+//! stream_handle.play_raw(source.convert_samples().speed(2.0));
+//!
+//! // The sound plays in a separate audio thread,
+//! // so we need to keep the main thread alive while it's playing.
+//! std::thread::sleep(std::time::Duration::from_secs(5));
+//!
+//! // here is how you would do it using the sink
```
dvdsk, line 33: should probably be two separate code "blocks" (I mean the ``` things). Also you can hide the imports using `#`.

dvdsk (changes requested):
Few tiny issues otherwise looks good!

dvdsk:
I found a great resource on documenting items, I have added it to the CONTRIBUTING.md guidlines: https://github.com/RustAudio/rodio/blob/master/CONTRIBUTING.md#documentation

ugochukwu-850:
@dvdsk nice vets , I would take an extra look on the contribution guidelines you cited and by tomorrow , I should have this in optimum status .

Btw I wanted to propose a new feature on the output stream structure , could I create a new issue for that .

src/source/speed.rs
ugochukwu-850: Yeah maybe I should be try sounding more human :rofl:

src/source/speed.rs
ugochukwu-850: I did not quite get this though . Do you mean the examples

src/source/speed.rs
ugochukwu-850: Done

dvdsk:
> Btw I wanted to propose a new feature on the output stream structure , could I create a new issue for that .

yes please do

src/source/speed.rs
dvdsk: yes the examples, I figured you put them in one code block to save having to show the setup code (imports etc) twice. However using # you can hide that.

ugochukwu-850: Argh, I get it now 

I would separate the examples so it is easier to understand 

And maybe the imports aren't that necessary ; I was just trying to conform to your docs pattern

dvdsk: > And maybe the imports aren't that necessary ; I was just trying to conform to your docs pattern

rodio is pretty old so unfortunately a lot of existing documentation is from before imports could be hidden.

src/source/speed.rs
ugochukwu-850: @dvdsk I just made changes to this .

src/source/speed.rs
ugochukwu-850: Alright , I hid the documentation and separated the two examples

ugochukwu-850:
@dvdsk Thanks for the resources you added to the contribution guidelines ; they were very useful . I think we are good now , please review and merge if its ok.

src/source/mod.rs
```diff
@@ -412,6 +412,7 @@ where
     }
 
     /// Changes the play speed of the sound. Does not adjust the samples, only the play speed.
+    /// Creates a Speed struct that handles the speed control
```
dvdsk, line 415: Could you replace this with the documentation from the Sink? I especially would like the note that the pitch will change. Finally could you note how the pitch will change (so at 0.5 speed is it higher or lower). That is something that might not be clear for readers of the doc.

ugochukwu-850: @dvdsk I have made the required changes , I also added some part of the doc to the sink.set_speed function to increase accessibility . Please review and merge

dvdsk (approved):

dvdsk:
thank you very much!

Issue #628, opened 2024-10-04
Total Duration after speed up calculation is faulty


ugochukwu-850:
After a speed increase the total duration tends to increase instead of decrease and vice versa .
I believe a change to the total_duration function  ,on the speed.rs module , to divide by factor instead of multiply by factor , would solve this .

Correct me if I am wrong

dvdsk:
you are completely right, that should be a division! Good catch. If you want you can make a PR fixing it. Otherwise I'll fix it on main.

ugochukwu-850:
Great 😊.
I have already fixed that .
I would raise the PR tomorrow.
I am to tired rn to mess with git ( I made the commit on a branch I was working on for another PR) .

dvdsk:
> I am to tired rn to mess with git

I feel you, im trying to figure out how to edit a pr right now I should probably go to bed instead :yawning_face:

PR #631, opened 2024-10-05
Issue#628: Fix total_duration malfunction after speed mutation

ugochukwu-850:
-- updated the speed struct function total_duration to divide instead of multiply by speed factor
-- removed redundant comment on try_seek function in the speed.rs file #628


dvdsk:
:+1:

PR #632, opened 2024-10-06
Add AGC example

### LINKED ISSUE #629: add automatic_gain example?
Should I add an AGC example?

UnknownSuperficialNight:
Here is a quick AGC example

I don't know how to use set_enabled() outside the setup area, so I'm using the experimental atomicbool

close #629

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -67,3 +67,7 @@ required-features = ["symphonia-isomp4", "symphonia-aac"]
 [[example]]
 name = "noise_generator"
 required-features = ["noise"]
+
+[[example]]
+name = "automatic_gain_control"
+required-features = ["dep:atomic_float"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -70,4 +70,4 @@ required-features = ["noise"]
 
 [[example]]
 name = "automatic_gain_control"
-required-features = ["dep:atomic_float"]
+required-features = ["atomic_float"]
--- a/examples/automatic_gain_control.rs
+++ b/examples/automatic_gain_control.rs
@@ -16,12 +16,10 @@ fn main() {
     let agc_source = source.automatic_gain_control(1.0, 4.0, 0.005, 5.0);
 
     // Get a handle to control the AGC's enabled state (only when using experimental feature)
-    #[cfg(feature = "experimental")]
     let agc_control = agc_source.get_agc_control();
 
     // Disable AGC by default when using experimental feature
-    #[cfg(feature = "experimental")]
-    agc_control.store(false, std::sync::atomic::Ordering::Relaxed);
+    agc_control.store(true, std::sync::atomic::Ordering::Relaxed);
 
     // Add the AGC-processed source to the sink for playback
     sink.append(agc_source);
--- a/examples/automatic_gain_control.rs
+++ b/examples/automatic_gain_control.rs
@@ -1,0 +1,31 @@
+use rodio::source::Source;
+use rodio::Decoder;
+use std::fs::File;
+use std::io::BufReader;
+
+fn main() {
+    let (_stream, handle) = rodio::OutputStream::try_default().unwrap();
+    let sink = rodio::Sink::try_new(&handle).unwrap();
+
+    let file = BufReader::new(File::open("assets/music.flac").unwrap());
+
+    // Decode the sound file into a source
+    let source = Decoder::new(file).unwrap();
+
+    // Apply automatic gain control to the source
+    let agc_source = source.automatic_gain_control(1.0, 4.0, 0.005, 5.0);
+
+    // Get a handle to control the AGC's enabled state (only when using experimental feature)
+    #[cfg(feature = "experimental")]
+    let agc_control = agc_source.get_agc_control();
+
+    // Disable AGC by default when using experimental feature
+    #[cfg(feature = "experimental")]
+    agc_control.store(false, std::sync::atomic::Ordering::Relaxed);
+
+    // Add the AGC-processed source to the sink for playback
+    sink.append(agc_source);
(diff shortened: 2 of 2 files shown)


dvdsk:
> so I'm using the experimental atomicbool

I rather not have examples using experimental features, since those might go away at any time. 

Its clear the current method for changing a source outside of setup is hard to find or use. We will work on that, maybe by changing the API maybe with extensive documentation if no better API can be found. 

For now an example showing how to do it might be very useful. So lets try and do it without atomicbool?

What you will need is [`PeriodicAccess`](https://docs.rs/rodio/0.19.0/rodio/source/trait.Source.html#method.periodic_access) it is missing an example. The best we have now is looking at the source code of the [`Sink`](https://github.com/RustAudio/rodio/blob/a5a49fb5a23d5cd1d8d95cbe69657a20934cdff0/src/sink.rs#L131).

Now we get something along the lines of this:
```rust
   let agc_enabled = Arc::new(Atomicbool::new(true));
   

   // Apply automatic gain control to the source
    let agc_source = source.automatic_gain_control(1.0, 4.0, 0.005, 5.0);

   // Make the audio thread check every 5 millis if the agc needs 
   // to be enabled/disabled
   let controlled = agc_source.periodic_access(Duration::from_millis(5), 
     |periodic_access_source| {
        let agc_source = periodic_access_source.inner_mut()
        agc_source.set_enabled(agc_enabled.load(Relaxed));
     });

   // put it in sink & play
   // from some other part of the program:
   agc_enabled.store(true, Relaxed); // takes effect after 5 millis.
```

UnknownSuperficialNight:
> > so I'm using the experimental atomicbool
> 
> I rather not have examples using experimental features, since those might go away at any time.
> 
> Its clear the current method for changing a source outside of setup is hard to find or use. We will work on that, maybe by changing the API maybe with extensive documentation if no better API can be found.
> 
> For now an example showing how to do it might be very useful. So lets try and do it without atomicbool?
> 
> What you will need is [`PeriodicAccess`](https://docs.rs/rodio/0.19.0/rodio/source/trait.Source.html#method.periodic_access) it is missing an example. The best we have now is looking at the source code of the [`Sink`](https://github.com/RustAudio/rodio/blob/a5a49fb5a23d5cd1d8d95cbe69657a20934cdff0/src/sink.rs#L131).
> 
> Now we get something along the lines of this:
> 
> ```rust
>    let agc_enabled = Arc::new(Atomicbool::new(true));
>    
> 
>    // Apply automatic gain control to the source
>     let agc_source = source.automatic_gain_control(1.0, 4.0, 0.005, 5.0);
> 
>    // Make the audio thread check every 5 millis if the agc needs 
>    // to be enabled/disabled
>    let controlled = agc_source.periodic_access(Duration::from_millis(5), 
>      |periodic_access_source| {
>         let agc_source = periodic_access_source.inner_mut()
>         agc_source.set_enabled(agc_enabled.load(Relaxed));
>      });
> 
>    // put it in sink & play
>    // from some other part of the program:
>    agc_enabled.store(true, Relaxed); // takes effect after 5 millis.
> ```

Yea, when I try periodic_access I can just never get it to work 

Example:
```
error[E0599]: no method named `inner_mut` found for mutable reference `&mut AutomaticGainControl<Decoder<BufReader<File>>>` in the current scope
  --> examples/automatic_gain_control.rs:27:53
   |
27 |             let agc_source = periodic_access_source.inner_mut();
   |                                                     ^^^^^^^^^ method not found in `&mut AutomaticGainControl<Decoder<BufReader<File>>>`
```

The only thing I could get the compiler to agree with is this:
```
let controlled =
        agc_source.periodic_access(Duration::from_millis(5), move |src| {
            src.set_enabled(agc_enabled.load(Relaxed));
        });
```

But of course that's not very useful as you can't use it

dvdsk:
ahh so the compile error states the type given to the closure is already the `AutomaticGainControl`. So that `let agc_source` bit that I give you is unneeded. Try something like `|agc| agc.set_enabled(agc_enabled.load(Relaxed))`.

It also points to a small issue with the agc source. It should have an `inner mut` and `inner` method. They are not part of `Source` since a source can be generating sound, thats why I forgot them while reviewing :sweat_smile:.

UnknownSuperficialNight:
> If this is still on your list let me know, there is no rush to finish this. However if you would rather leave it ill take it over. I would like to have this merged before releasing 0.20

Sorry, I got a bit distracted with my own project and totally forgot about this. I did try to fiddle with the code a bit more to get the compiler to play nice, but no luck. If you want to take it over, go for it! I’d love to know what solution you find, so if you don't mind, let me know here when/if you get it working. For whatever reason, I can only get the atomicbool to agree with it. If it comes down to it, and we can’t get the compiler to cooperate, let’s just use the experimental one—it’s better than having no example at all.

UnknownSuperficialNight:
Marking it as ready for review so it can be edited and merged if you need to

dvdsk:
> I’d love to know what solution you find, so if you don't mind, let me know here when/if you get it working.

Ill jump on it, im gonna merge this first and then directly edit on main as thats a little easier then editing the PR. Ill let you know once thats done.

dvdsk:
see 48408bcfa6a6e4ff207fce5fa3b68ea96515bb41 for my approach

UnknownSuperficialNight:
Ah, I see what I was not doing 

I didn't add the `move` in     
```rust
let controlled = agc_source.periodic_access(Duration::from_millis(5), move |agc_source| {
        agc_source.set_enabled(agc_enabled_clone.load(Ordering::Relaxed));
    });
```

and I didn't add the inner_mut and inner to AGC itself

Thanks for the help :)

Much easier with an example of usage

Issue #584, opened 2024-06-07
Crackly audio on some mp3 files


Chazoshtare:
I'm writing a project that converts text to speech using Google TTS. The mp3 files it returns have some audible crackling noise in them while played with rodio. It's not an issue with the file itself, as it plays normally everywhere else, such as in music players and the like.
I tried:
- playing it through stream_handle or a sink
- decoding it with different decoders
- playing it through different audio interfaces and outputs
- all that on Linux and Windows

In all cases it was the same, so I'm rather convinced it's a problem on rodio's side. I'm attaching a simple project with a sample sound that demonstrates the issue.
[rodio-crackling-issue.tar.gz](https://github.com/user-attachments/files/15744050/rodio-crackling-issue.tar.gz)

dvdsk:
First off all thanks for the demonstrator project, that saves me a some time :)

Now the bad news, I can not reproduce it locally on my machine, which makes things a bit more difficult.

Lets start by checking if there could be an issue in cpal. Could you try the [beep example](https://github.com/RustAudio/cpal/blob/master/examples/beep.rs) from cpal?

If that works without issues ill see if I can find some more systems to try the demonstrator project on and reproduce the crackling.

Chazoshtare:
The beep example you provided produces a clean beep. Generally, rodio plays most of the sounds I have without any problems, I only observe the issue with mp3 files returned by google TTS.
Just in case so you could hear what it's like on my side, I recorded how it sounds using a loopback from my interface.
[crackles.zip](https://github.com/user-attachments/files/15749323/crackles.zip)

dvdsk:
> The beep example you provided produces a clean beep. Generally, rodio plays most of the sounds I have without any problems, I only observe the issue with mp3 files returned by google TTS. Just in case so you could hear what it's like on my side, I recorded how it sounds using a loopback from my interface. [crackles.zip](https://github.com/user-attachments/files/15749323/crackles.zip)

That is pretty interesting (useful?) actually since... I do not hear the crackling in that recording either... 

Ill be honest I have no idea at this point. And without a way to replicate it on my side I can not really try things out. If you need to get this working fast your best bet is probably to try kira (a different rust audio lib) or build something on the [symphonia play example](https://github.com/pdeljanov/Symphonia/tree/master/symphonia-play).

On the other hand I am open to trying things on your side until we get a better idea (if we ever do, this one is really confusing me).

[...]

PR #640, opened 2024-11-13
Test fixes, upgrade quickcheck

PetrGlad:
1. `quickcheck`  1.x input values vary more than in 0.9 which finds overflow cases and excessive memory allocation problems in `Sample::lerp`.  Calculations now use f64 to avoid that. Conversion back to 16 bit resolution is done explicitly since `as` conversion may silently discard most significant bits. Alternatively, interpolation coefficient (`numerator/dividor`) may use e.g. `u16` instead of `u32`, or maybe even floating point. I believe `u32` precision is unnecessary there. But that requires updating sample rate converter which at the moment is the only user of this interpolation.
2. Documentation for `Sample::lerp` was incorrect. It said that calculations should follow `c * first + (c - 1) * second`. which gives `first` at `c==1` and `second` at `c==0` but actual implementations use the opposite `first` at `c==0` and `second` at `c==1`.
3. Fix documentation examples. Some of the examples are marked as `no_run` since they require audio devices and may actually make sounds while `cargo test` command is running.
4. Include documentation tests into CI. `--all-targets` switch excludes doc tests, so those are executed separately.
5. Non `experimental` builds exclude some code and in some code parts use mutually exclusive implementations, so I included those in CI too to keep experimental code compilable.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,7 +47,10 @@ jobs:
         if: matrix.toolchain == 'stable' && matrix.os == 'ubuntu-latest'
 
       - run: cargo test --all-targets
-      - run: cargo test --features=symphonia-all --all-targets
+      - run: cargo test --all-targets --features=experimental
+      - run: cargo test --all-targets --features=symphonia-all
+      - run: cargo test --doc
+      - run: cargo test --doc --features=experimental
   cargo-publish:
     if: github.event_name == 'push' && github.ref == 'refs/heads/master'
     env:
--- a/examples/automatic_gain_control.rs
+++ b/examples/automatic_gain_control.rs
@@ -18,18 +18,28 @@ fn main() {
     // Apply automatic gain control to the source
     let agc_source = source.automatic_gain_control(1.0, 4.0, 0.005, 5.0);
 
-    // Make it so that the source checks if automatic gain control should be
-    // enabled or disabled every 5 milliseconds. We must clone `agc_enabled`
-    // or we would lose it when we move it into the periodic access.
-    let agc_enabled = Arc::new(AtomicBool::new(true));
-    let agc_enabled_clone = agc_enabled.clone();
-    let controlled = agc_source.periodic_access(Duration::from_millis(5), move |agc_source| {
-        agc_source.set_enabled(agc_enabled_clone.load(Ordering::Relaxed));
-    });
+    let agc_enabled : Arc<AtomicBool>;
 
-    // Add the source now equipped with automatic gain control and controlled via
-    // periodic_access to the sink for playback
-    sink.append(controlled);
+    #[cfg(not(feature = "experimental"))]
+    {
+        agc_enabled = Arc::new(AtomicBool::new(true));
+        // Make it so that the source checks if automatic gain control should be
+        // enabled or disabled every 5 milliseconds. We must clone `agc_enabled`,
+        // or we would lose it when we move it into the periodic access.
+        let agc_enabled_clone = agc_enabled.clone();
+        let controlled = agc_source.periodic_access(Duration::from_millis(5), move |agc_source| {
+            agc_source.set_enabled(agc_enabled_clone.load(Ordering::Relaxed));
+        });
+        
+        // Add the source now equipped with automatic gain control and controlled via
+        // periodic_access to the sink for playback
+        sink.append(controlled);
+    }
+    #[cfg(feature = "experimental")]
+    {
+        agc_enabled = agc_source.get_agc_control();
+        sink.append(agc_source);
+    }
 
     // after 5 seconds of playback disable automatic gain control using the
     // shared AtomicBool `agc_enabled`. You could do this from another part
@@ -39,6 +49,11 @@ fn main() {
     // Note that disabling the AGC takes up to 5 millis because periodic_access
     // controls the source every 5 millis.
     thread::sleep(Duration::from_secs(5));
+    #[cfg(not(feature = "experimental"))]
+    agc_enabled.store(false, Ordering::Relaxed);
+
+    // AGC on/off control using direct access to the boolean variable.
+    #[cfg(feature = "experimental")]
     agc_enabled.store(false, Ordering::Relaxed);
 
     // Keep the program running until playback is complete
--- a/examples/automatic_gain_control.rs
+++ b/examples/automatic_gain_control.rs
@@ -18,7 +18,7 @@ fn main() {
     // Apply automatic gain control to the source
     let agc_source = source.automatic_gain_control(1.0, 4.0, 0.005, 5.0);
 
-    let agc_enabled : Arc<AtomicBool>;
+    let agc_enabled: Arc<AtomicBool>;
 
     #[cfg(not(feature = "experimental"))]
     {
@@ -30,7 +30,7 @@ fn main() {
         let controlled = agc_source.periodic_access(Duration::from_millis(5), move |agc_source| {
             agc_source.set_enabled(agc_enabled_clone.load(Ordering::Relaxed));
         });
-        
+
         // Add the source now equipped with automatic gain control and controlled via
         // periodic_access to the sink for playback
         sink.append(controlled);
--- a/src/conversions/sample.rs
+++ b/src/conversions/sample.rs
@@ -243,7 +243,7 @@ mod test {
             let b = second as f64;
             let c = numerator as f64 / denominator as f64;
             if c < 0.0 || c > 1.0 { return TestResult::discard(); };
-            let reference = a + (b - a) * c;
+            let reference = a * (1.0 - c) + b * c;
             let x = Sample::lerp(first, second, numerator, denominator) as f64;
             let d = x - reference;
             TestResult::from_bool(d.abs() < 1.0)
--- a/src/conversions/sample.rs
+++ b/src/conversions/sample.rs
@@ -55,8 +55,7 @@ where
     I: ExactSizeIterator,
     I::Item: Sample,
     O: FromSample<I::Item> + Sample,
-{
-}
+{}
 
 /// Represents a value of a single sample.
 ///
@@ -93,10 +92,10 @@ pub trait Sample: CpalSample {
 impl Sample for u16 {
     #[inline]
     fn lerp(first: u16, second: u16, numerator: u32, denominator: u32) -> u16 {
-        let a = first as i32;
-        let b = second as i32;
-        let n = numerator as i32;
-        let d = denominator as i32;
+        let a = first as i64;
+        let b = second as i64;
+        let n = numerator as i64;
+        let d = denominator as i64;
         (a + (b - a) * n / d) as u16
     }
 
@@ -176,5 +175,28 @@ impl Sample for f32 {
     #[inline]
     fn zero_value() -> f32 {
         0.0
+    }
+}
+
+
+#[cfg(test)]
+mod test {
+    use quickcheck::{quickcheck, TestResult};
+    use super::*;
+
+    quickcheck! {
+        fn lerp_u16(first: u16, second: u16, numerator: u32, denominator: u32) -> TestResult {
+            if denominator == 0 { return TestResult::discard(); }
+            let a = first as f64;
+            let b = second as f64;
+            let c = numerator as f64 / denominator as f64;
+            // if c < 0.0 || c > 1.0 { return TestResult::discard(); };
+            // let reference = a * c + b * (1.0 - c);
+            let reference = a + (b - a) * c;
+            let x = Sample::lerp(first, second, numerator, denominator) as f64;
+            let d = x - reference;
+            dbg!(x, reference, d);
+            TestResult::from_bool(d.abs() < 1.0)
+        }
     }
 }
--- a/src/conversions/sample.rs
+++ b/src/conversions/sample.rs
@@ -79,7 +79,7 @@ pub trait Sample: CpalSample {
     /// Multiplies the value of this sample by the given amount.
     fn amplify(self, value: f32) -> Self;
 
-    /// Converts the sample to an f32 value.
+    /// Converts the sample to a f32 value.
     fn to_f32(self) -> f32;
 
     /// Calls `saturating_add` on the sample.
@@ -92,11 +92,9 @@ pub trait Sample: CpalSample {
 impl Sample for u16 {
     #[inline]
     fn lerp(first: u16, second: u16, numerator: u32, denominator: u32) -> u16 {
-        let a = first as i64;
-        let b = second as i64;
-        let n = numerator as i64;
-        let d = denominator as i64;
-        (a + (b - a) * n / d) as u16
+        let d = first as i64 + (second as i64 - first as i64) * numerator as i64 / denominator as i64;
+        u16::try_from(d)
+            .expect("numerator / denominator is within [0, 1] range")
     }
 
     #[inline]
@@ -124,8 +122,9 @@ impl Sample for u16 {
 impl Sample for i16 {
     #[inline]
     fn lerp(first: i16, second: i16, numerator: u32, denominator: u32) -> i16 {
-        (first as i32 + (second as i32 - first as i32) * numerator as i32 / denominator as i32)
-            as i16
+        let d = first as i64 + (second as i64 - first as i64) * numerator as i64 / denominator as i64;
+        i16::try_from(d)
+            .expect("numerator / denominator is within [0, 1] range")
     }
 
     #[inline]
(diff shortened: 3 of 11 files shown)


src/conversions/sample.rs
```diff
@@ -178,3 +177,76 @@ impl Sample for f32 {
         0.0
     }
 }
+
```
PetrGlad, line 180: Were the tests removed? These can be shorter but actually useful to check for edge cases. Type conversions' behavior is not always obvious.

src/conversions/sample_rate.rs
```diff
@@ -50,10 +50,14 @@ where
         let from = from.0;
         let to = to.0;
 
+        assert!(num_channels >= 1);
+        assert!(num_channels <= 128);
         assert!(from >= 1);
+        assert!(from <= 500_000);
         assert!(to >= 1);
+        assert!(to <= 500_000);
```
PetrGlad, line 58: Just picked some large numbers, if exceeded the error may only be found by memory allocation errors, which are not helpful. Maybe these can be some global constants so they can be checked separately.

dvdsk (commented):
Nice fixes and improvements, thank you very much. This must have taken a while! 

Few things I am unsure about as you'll see in the discussion. Its a bit too late for me to check on the performance impact of the try_from. Ill see if I can add a new interpolate benchmark to rodio tomorrow.

examples/automatic_gain_control.rs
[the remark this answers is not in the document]
PetrGlad: Experimental API was already used in the example. I have only fixed compilation. It should be removed or changed to older API.

src/conversions/sample.rs
PetrGlad: Is this comment about the tests or the calculation change?
Another option to avoid overflows would be to use smaller integer types for the coefficient,  I think `u32` is too much for the use case. Or, maybe use a float as coefficient.

src/conversions/sample_rate.rs
PetrGlad: I would not insist on this, but in my experience it makes sense to check for obviously unusable values, they may be there by mistake or some malicious misconfiguration.

src/conversions/sample.rs
PetrGlad: How one would test the performance here? Setting up a lot of conversions in a row?

examples/automatic_gain_control.rs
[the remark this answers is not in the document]
PetrGlad: Updated.

src/conversions/sample_rate.rs
```diff
@@ -50,10 +50,11 @@ where
         let from = from.0;
         let to = to.0;
 
+        assert!(num_channels >= 1);
         assert!(from >= 1);
         assert!(to >= 1);
```
PetrGlad, line 55: Sample rate converter allocates a buffer that is proportional to number of channels (65k * sample size).
Also `SampleRateConverter::size_hint` does not seem to be accurate and it is used to collect items into Vecs in test.
There's a TODO note in it BTW. It seems to be rather complicated, maybe there can be a simpler but imprecise version.

dvdsk:
Love the work here, can not look at it the next few days unfortunately. After the weekend I'll see if I can come up with a benchmark for the resampler. We need one anyway (I want to introduce a highfy resampler in the future). Ill do a detailed review and answer all the questions that came up then.

PetrGlad:
Looking a other filters in `rodio` I am now curious if rodio should only provide basic functions and use https://github.com/RustAudio/dasp for complex processing instead. Some of the rodio functionality is already implemented there.

As I understand rodio in most cases can use generic parameters to avoid dyn references (but Sync and Mixer still use them). That probably can be slightly  more efficient. Any other reason rodio cannot use some existing external libraries for sound processing graph?

dvdsk:
> I am now curious if rodio should only provide basic functions and use [RustAudio/dasp](https://github.com/RustAudio/dasp) for complex processing instead.

Rodio and dasp have fundamentally different goals, that makes it hard and maybe unwise to merge them or require users to be familiar with both. 
While rodio is an audio playback crate dasp is a signal processing suite. When Rodio is used its usually not the main feature of the application. The app might be a simple game, ui with sounds on click or podcast app. So Rodio should focus on doing most things efficient while getting out of the way of the user. That is why for example `Sink` exists. You do not need it, it adds nothing that can not be done with the other parts of rodio. The `Sink` API does however cover most use-cases and is easy to understand/use. (Its also far from perfect and has its own issues which we are going to address).

> Some of the rodio functionality is already implemented there.

Rodio predates dasp, so its easy to turn this around and ask, why do they not depend on rodio. Again its a difference in goals, dasp wants to have zero dependencies, rodio just wants to be easy to use (so no C-deps if possible).
 
> use [RustAudio/dasp](https://github.com/RustAudio/dasp) for complex processing instead

I also think rodio has most features dasp has since #602 landed. I might be mistaken there. And a Source that makes it easy to use dasp from rodio might be interesting.

The sinc interpolator in dasp is interesting, rodio needs an option for a slower but more hifi interpolator. See  https://github.com/RustAudio/rodio/issues/584.

.github/workflows/ci.yml
```diff
@@ -47,7 +47,9 @@ jobs:
         if: matrix.toolchain == 'stable' && matrix.os == 'ubuntu-latest'
 
       - run: cargo test --all-targets
-      - run: cargo test --features=symphonia-all --all-targets
+      - run: cargo test --all-targets --features=experimental
+      - run: cargo test --all-targets --features=symphonia-all
+      - run: cargo test --doc
```
dvdsk, line 52: doc tests run be default with just cargo test. The only thing `--doc` does is only run doc tests

examples/automatic_gain_control.rs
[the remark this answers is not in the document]
dvdsk: > Experimental API was already used in the example.

its not on main: https://github.com/RustAudio/rodio/blob/master/examples/automatic_gain_control.rs

I point to this example whenever we get questions about how to use things like periodic access. Maybe it was different on main when you branched off?

dvdsk (changes requested):
Only AGC example and what is possibly a superfluous --doc argument in ci.yml left and then we can merge this :+1:

src/conversions/sample_rate.rs
dvdsk: Could you document that in the assert message? Otherwise one might wonder where the number comes from.

dvdsk:
Still have not found the time for the benchmark, as soon as I've done that ill post the results here and we can see what we do based on them.

.github/workflows/ci.yml
PetrGlad: Yes, `cargo test` runs doc tests, but `cargo test --all-targets` does not. 
```
 cargo test --help
Execute all unit and integration tests and build examples of a local package
...
Target Selection:
      ...
      --all-targets       Test all targets (does not include doctests)
      --doc               Test only this library's documentation
...
```

Then maybe just run `cargo test` without `--all-targets`?

src/source/speed.rs
```diff
@@ -4,37 +4,36 @@
...
 //!# use std::io::BufReader;
 //!# use rodio::{Decoder, Sink, OutputStream, source::{Source, SineWave}};
 //!
 //! // Get an output stream handle to the default physical sound device.
-//! // Note that no sound will be played if _stream is dropped
+//! // Note that no sound will be played if _stream is dropped.
 //! let (_stream, stream_handle) = OutputStream::try_default().unwrap();
 //! // Load a sound from a file, using a path relative to Cargo.toml
-//! let file = BufReader::new(File::open("examples/music.ogg").unwrap());
+//! let file = BufReader::new(File::open("assets/music.ogg").unwrap());
```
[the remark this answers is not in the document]
dvdsk: correct, thanks for the fix though!

.github/workflows/ci.yml
PetrGlad: Updated.

src/conversions/sample_rate.rs
PetrGlad: I decided to remove those for now. Maybe this should be a separate task to make rodio more resilient to misconfiguration, as there are many similar places where earlier diagnostics via assertions can be helpful.

examples/automatic_gain_control.rs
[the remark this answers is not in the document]
PetrGlad: I have used `cargo test --lib --bins --tests --benches --features=experimental` (see https://github.com/PetrGlad/rodio/commit/bf60185d4bb97489ce2a0c2fc02ccb5413a678e2). Since it keeps examples without switches. If one day there's need to have examples for experimental features then one may put back `#[cfg()]` switches or use some other solution.

src/conversions/sample_rate.rs
dvdsk: > I decided to remove those for now. Maybe this should be a separate task to make rodio more resilient to misconfiguration, as there are many similar places where earlier diagnostics via assertions can be helpful.

That could be useful, as long as they do not impact perf I do not see a disadvantage.

examples/automatic_gain_control.rs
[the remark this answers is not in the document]
dvdsk: that works for now :+1: 

In my opinion its better not to have examples for experimental features. Those features are for *some* users to experiment with. With an example the barrier to using an experimental feature decreases. That leads to more users affected by us 'experimenting' with the experimental API.

dvdsk:
benchmarks this pr on my machine:
```
Timer precision: 20 ns
resampler         fastest       │ slowest       │ median        │ mean          │ samples │ iters
├─ no_resampling  2.414 ms      │ 2.701 ms      │ 2.508 ms      │ 2.515 ms      │ 100     │ 100
╰─ resample_to                  │               │               │               │         │
   ├─ 8000        2.671 ms      │ 2.755 ms      │ 2.692 ms      │ 2.693 ms      │ 100     │ 100
   ├─ 11025       2.416 ms      │ 2.682 ms      │ 2.433 ms      │ 2.436 ms      │ 100     │ 100
   ├─ 16000       3.723 ms      │ 3.812 ms      │ 3.775 ms      │ 3.772 ms      │ 100     │ 100
   ├─ 22050       3.505 ms      │ 3.62 ms       │ 3.527 ms      │ 3.53 ms       │ 100     │ 100
   ├─ 44100       2.384 ms      │ 3.039 ms      │ 2.456 ms      │ 2.486 ms      │ 100     │ 100
   ├─ 48000       7.421 ms      │ 7.512 ms      │ 7.435 ms      │ 7.439 ms      │ 100     │ 100
   ├─ 88200       11.22 ms      │ 11.38 ms      │ 11.31 ms      │ 11.31 ms      │ 100     │ 100
   ├─ 96000       12.87 ms      │ 13.2 ms       │ 13.18 ms      │ 13.17 ms      │ 100     │ 100
   ├─ 176400      20.75 ms      │ 20.91 ms      │ 20.85 ms      │ 20.85 ms      │ 100     │ 100
   ├─ 192000      23.73 ms      │ 27.32 ms      │ 23.94 ms      │ 23.99 ms      │ 100     │ 100
   ├─ 352800      40.9 ms       │ 41.99 ms      │ 40.95 ms      │ 41 ms         │ 100     │ 100
   ╰─ 384000      46.4 ms       │ 47.51 ms      │ 46.45 ms      │ 46.48 ms      │ 100     │ 100

```

main branch:
```
Timer precision: 20 ns
resampler         fastest       │ slowest       │ median        │ mean          │ samples │ iters
├─ no_resampling  2.294 ms      │ 5.027 ms      │ 2.498 ms      │ 2.591 ms      │ 100     │ 100
╰─ resample_to                  │               │               │               │         │
   ├─ 8000        2.129 ms      │ 2.307 ms      │ 2.151 ms      │ 2.153 ms      │ 100     │ 100
   ├─ 11025       2.386 ms      │ 2.444 ms      │ 2.4 ms        │ 2.402 ms      │ 100     │ 100
   ├─ 16000       3.005 ms      │ 5.783 ms      │ 3.042 ms      │ 3.075 ms      │ 100     │ 100
   ├─ 22050       3.479 ms      │ 4.052 ms      │ 3.49 ms       │ 3.497 ms      │ 100     │ 100
   ├─ 44100       2.283 ms      │ 3.367 ms      │ 2.504 ms      │ 2.574 ms      │ 100     │ 100
   ├─ 48000       6.254 ms      │ 7.076 ms      │ 6.26 ms       │ 6.272 ms      │ 100     │ 100
   ├─ 88200       9.795 ms      │ 20.01 ms      │ 10.09 ms      │ 10.24 ms      │ 100     │ 100
   ├─ 96000       11.52 ms      │ 21.97 ms      │ 11.53 ms      │ 11.74 ms      │ 100     │ 100
   ├─ 176400      19.02 ms      │ 19.15 ms      │ 19.03 ms      │ 19.03 ms      │ 100     │ 100
   ├─ 192000      22.13 ms      │ 22.26 ms      │ 22.14 ms      │ 22.15 ms      │ 100     │ 100
   ├─ 352800      38.63 ms      │ 38.76 ms      │ 38.65 ms      │ 38.66 ms      │ 100     │ 100
   ╰─ 384000      44.03 ms      │ 44.17 ms      │ 44.06 ms      │ 44.06 ms      │ 100     │ 100
```

So that's a 5 to 10% slowdown. A bit too much. Lets see what causes it

PetrGlad:
@dvdsk Security-wise there is no strict need in checked cast but u32 * u32  may overflow in some cases. AFAIK overlapped value will wrap over zero in release build, and panic in debug mode. I'd expected some clicks or distortion in case of overflows and panics in debug mode (which was triggered by quickcheck). The final interpolated value should be in range, though.

One could use smaller integers for the ratios (I doubt u32 precision is actually necessary there), or use f32 as interpolation coefficient. I think smaller integers, like u16 can be an approximation, maybe even something crude, like dropping some least significant bits in the u32 numerator and divider when one of those has significant bits beyond 16 position. Or finding some other approximate value pair.

PetrGlad:
@dvdsk Security-wise there is no strict need in checked cast but u32 * u32  may overflow in some cases. AFAIK overlapped value will wrap over zero in release build, and panic in debug mode. I'd expected some clicks or distortion in case of overflows and panics in debug mode (which was triggered by quickcheck). The final interpolated value should be in range, though.

One could use smaller integers for the ratios (I doubt u32 precision is actually necessary there), or use f32 as interpolation coefficient. I think smaller integers, like u16 can be an approximation, maybe even something crude, like dropping some least significant bits in the u32 numerator and divider when one of those has significant bits beyond 16 position. Or finding some other approximate value pair.
E.g. `num` crate provides such algorithm:
```rust
let ratio = num::rational::Rational32::approximate_float(2.3f32).unwrap();
dbg!(ratio.numer(), ratio.denom());
```
Maybe making this more reliable can be a separate task...

dvdsk:
> I'd expected some clicks or distortion in case of overflows and panics in debug mode (which was triggered by quickcheck).

might be related to #584, I thought that was a property of the linear interpolator and could not be fixed. You might have stumbled on the fix for it :tada:

> Maybe making this more reliable can be a separate task...

I think that is the best, not only can we try fixing #584 then but we might also look at how to enable users to pick their own re-sampler and introduce more hifi alternatives like rubato or dasps sinc resampler. That should be pretty easy, I think an extra option on the OutputStream builder?

On a separate note, how would you feel about joining me as maintainer of rodio?

PetrGlad:
Yes, I would like to help. I cannot promise much, but I do have some free time for this.

PetrGlad:
Regarding interpolation, now I think I understand it better. I looks like there are only some specific combinations of conversion rates where it can overflow. I have removed the explicit overflow check from interpolation and tried to clarify the limitations the docs.
Regarding #584, it looks like the output sample is converted from 2400Hz to 44100Hz in that case, the above overflows should not occur in this case.
If this is indeed the rate conversion problem then I think that is linear interpolation causing distortions.

PetrGlad:
Well resampler is just another filter, yes output stream builder can have this option.

dvdsk:
CI wants another cargo fmt run :stuck_out_tongue:.

dvdsk:
tldr: Same kind of perf regression as before. Pretty strange, I would expect perf to be the same now I'll hunt for the cause tomorrow.

this pr:
```
Timer precision: 20 ns
resampler         fastest       │ slowest       │ median        │ mean          │ samples │ iters
├─ no_resampling  2.399 ms      │ 2.795 ms      │ 2.494 ms      │ 2.505 ms      │ 100     │ 100
╰─ resample_to                  │               │               │               │         │
   ├─ 8000        2.137 ms      │ 2.385 ms      │ 2.187 ms      │ 2.197 ms      │ 100     │ 100
   ├─ 11025       2.377 ms      │ 2.436 ms      │ 2.392 ms      │ 2.395 ms      │ 100     │ 100
   ├─ 16000       3.085 ms      │ 3.173 ms      │ 3.12 ms       │ 3.122 ms      │ 100     │ 100
   ├─ 22050       3.501 ms      │ 5.609 ms      │ 3.52 ms       │ 3.543 ms      │ 100     │ 100
   ├─ 44100       2.432 ms      │ 3.042 ms      │ 2.557 ms      │ 2.585 ms      │ 100     │ 100
   ├─ 48000       6.456 ms      │ 6.505 ms      │ 6.476 ms      │ 6.477 ms      │ 100     │ 100
   ├─ 88200       10.71 ms      │ 10.82 ms      │ 10.75 ms      │ 10.75 ms      │ 100     │ 100
   ├─ 96000       12.14 ms      │ 14.13 ms      │ 12.18 ms      │ 12.21 ms      │ 100     │ 100
   ├─ 176400      20.01 ms      │ 23.26 ms      │ 20.07 ms      │ 20.12 ms      │ 100     │ 100
   ├─ 192000      22.95 ms      │ 26.81 ms      │ 23.02 ms      │ 23.16 ms      │ 100     │ 100
   ├─ 352800      40.27 ms      │ 40.79 ms      │ 40.4 ms       │ 40.4 ms       │ 100     │ 100
   ╰─ 384000      45.8 ms       │ 46.39 ms      │ 45.94 ms      │ 45.95 ms      │ 100     │ 100
```

main:
```
Timer precision: 20 ns
resampler         fastest       │ slowest       │ median        │ mean          │ samples │ iters
├─ no_resampling  2.283 ms      │ 2.75 ms       │ 2.392 ms      │ 2.426 ms      │ 100     │ 100
╰─ resample_to                  │               │               │               │         │
   ├─ 8000        2.342 ms      │ 2.423 ms      │ 2.356 ms      │ 2.36 ms       │ 100     │ 100
   ├─ 11025       2.518 ms      │ 2.88 ms       │ 2.533 ms      │ 2.54 ms       │ 100     │ 100
   ├─ 16000       3.207 ms      │ 3.283 ms      │ 3.25 ms       │ 3.248 ms      │ 100     │ 100
   ├─ 22050       3.599 ms      │ 3.635 ms      │ 3.613 ms      │ 3.614 ms      │ 100     │ 100
   ├─ 44100       2.326 ms      │ 3.04 ms       │ 2.48 ms       │ 2.51 ms       │ 100     │ 100
   ├─ 48000       6.368 ms      │ 6.407 ms      │ 6.384 ms      │ 6.384 ms      │ 100     │ 100
   ├─ 88200       9.995 ms      │ 10.83 ms      │ 10.28 ms      │ 10.28 ms      │ 100     │ 100
   ├─ 96000       11.92 ms      │ 12 ms         │ 11.95 ms      │ 11.95 ms      │ 100     │ 100
   ├─ 176400      19.33 ms      │ 36.44 ms      │ 19.4 ms       │ 19.57 ms      │ 100     │ 100
   ├─ 192000      22.5 ms       │ 22.71 ms      │ 22.57 ms      │ 22.57 ms      │ 100     │ 100
   ├─ 352800      39.32 ms      │ 54.21 ms      │ 39.43 ms      │ 39.67 ms      │ 100     │ 100
   ╰─ 384000      44.85 ms      │ 53.31 ms      │ 44.98 ms      │ 45.07 ms      │ 100     │ 100
```

src/conversions/sample_rate.rs
```diff
@@ -83,10 +77,13 @@ where
             (first, next)
         };
 
+        // Reducing nominator to avoid numeric overflows during interpolation.
+        let (to, from) = Ratio::new(to, from).into_raw();
```
dvdsk, line 81: I took a look at the Ratio source, and this does the same as the original code that did `from: from/gcd` and `to: to/gcd`. See: https://docs.rs/num-rational/latest/src/num_rational/lib.rs.html#108-112. Though I had now idea why they did that until now.

Given that what is the motivation for using `Ratio` over `/ gcd`?

PetrGlad:
Is there a way to compare the benchmark snapshots? This is the part I liked in `criterion` that it tries to clearly show changes between versions. Although its API is more cumbersome.

PetrGlad:
Cargo.lock must be in git. It is important for builds to be reproducible.

dvdsk:
> Cargo.lock should be in git somewhere, at least for benches and tests. It is necessary for builds to be reproducible. Projects that have `rodio` as a dependency will ignore it anyway. I see other Rust library projects do track it.
> Alternatively benches can be in a separate binary which will have a `Cargo.lock`.

good call, agreed!

src/conversions/sample_rate.rs
dvdsk: > Even if this is an extra call it happens only once at the converter creation.

pretty sure that will be in-lined so i'll be okay

> It is also used in tests, to limit input values, I thought it makes sense to re-use that and have a bit less code to test.

:+1:

dvdsk:
> tldr: Same kind of perf regression as before. Pretty strange, I would expect perf to be the same now I'll hunt for the cause tomorrow.

as soon as I've handled the other PR's/issues ill go hunting

dvdsk:
ran the benchmarks again side by side and now this pr is faster..... I am gonna merge and not worry about perf :)

dvdsk (approved):

PR #645, opened 2024-11-24
Fix wrong total duration caused by divided by zero conversion

### LINKED ISSUE #601: Incorrect total_duration while using symphonia features
While using rodio with symphonia features enabled values reported by [Decoder::total_duration](https://docs.rs/rodio/latest/rodio/decoder/struct.Decoder.html#method.total_duration) are incorrect. It works fine while using other backends. Calculating it with symphonia directly also return correct value. Minimum example with 7s long, flac file:

<details>
  <summary>Cargo.toml</summary>
   
  ```toml
  [dependencies]
  rodio = { version = "0.19.0", default-features = false, features = [
      "symphonia-all",
  ] }
  ```
  
</details>

<details>
  <summary>main.rs</summary>
   
  ```rust
  use std::{fs::File, io::BufReader, path::PathBuf, str::FromStr, time::Duration};

  use rodio::{Decoder, Source};

  fn main() {
      let track_path = PathBuf::from_str("test.flac").unwrap();
      let file = BufReader::new(File::open(track_path).unwrap());
      let track_source = Decoder::new(file).unwrap();
      let track_duration = track_source.total_duration();
      let real_duration = Some(Duration::from_secs(7));
  
      assert_eq!(track_duration, real_duration);
  }
  ```
  
</details>

<details>
  <summary>Output</summary>
   
  ```plain
  thread 'main' panicked at src/main.rs:12:5:
  assertion `left == right` failed
    left: Some(11.294967295s)
   right: Some(7s)
  ```
  
</details>

[Audio file](https://github.com/user-attachments/files/16552223/test.zip)
[More complex example and comparison with symphonia](https://github.com/user-attachments/files/16552267/rodio-issue.zip)

Methapon2001:
fixes #601

I came into this issue when trying to create music player as my learning project and notice that total duration is not quite right as the music ended before total duration reached.

The cause seems to be related to 0 division and convert to u32.

I also added test case for the conversion, I'm not sure if it is needed.
Any feedback is welcome.

Code:
```rs
        let total_duration = track
            .codec_params
            .time_base
            .zip(track.codec_params.n_frames)
            .map(|(time_base, n_frames)| time_base.calc_time(n_frames));

        total_duration.map(|time| {
            let hours = time.seconds / (60 * 60);
            let mins = (time.seconds % (60 * 60)) / 60;
            let secs = f64::from((time.seconds % 60) as u32) + time.frac;

            println!("Time: {}", format!("{}:{:0>2}:{:0>6.3}", hours, mins, secs))
        });

        total_duration.map(|Time { seconds, frac }| {
            let dur = Duration::new(seconds, (1f64 / frac) as u32).as_secs();

            let hours = dur / (60 * 60);
            let mins = dur / 60;
            let secs = dur % 60;

            println!(
                "Duration: {}",
                format!("{}:{:0>2}:{:0>2}", hours, mins, secs)
            );
        });

        total_duration.map(|Time { seconds, frac }| {
            let dur =
                Duration::new(seconds, if frac > 0.0 { (1f64 / frac) as u32 } else { 0 }).as_secs();

            let hours = dur / (60 * 60);
            let mins = dur / 60;
            let secs = dur % 60;

            println!(
                "Fixed Divided by 0 Duration: {}",
                format!("{}:{:0>2}:{:0>2}", hours, mins, secs)
            );
        });
```

Result:
```
Time: 0:00:07.000
Duration: 0:00:11
Fixed Divided by 0 Duration: 0:00:07
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Breaking: `Sink::try_new` renamed to `connect_new` and does not return error anymore.
             `Sink::new_idle` was renamed to `new`.
 
+### Fixed
+
+- Symphonia decoder `total_duration` incorrect value caused by conversion from `Time` to `Duration`.
+
 # Version 0.20.1 (2024-11-08)
 
 ### Fixed
--- a/src/decoder/symphonia.rs
+++ b/src/decoder/symphonia.rs
@@ -170,8 +170,9 @@ impl Source for SymphoniaDecoder {
 
     #[inline]
     fn total_duration(&self) -> Option<Duration> {
-        self.total_duration
-            .map(|Time { seconds, frac }| Duration::new(seconds, (1f64 / frac) as u32))
+        self.total_duration.map(|Time { seconds, frac }| {
+            Duration::new(seconds, if frac > 0.0 { (1f64 / frac) as u32 } else { 0 })
+        })
     }
 
     fn try_seek(&mut self, pos: Duration) -> Result<(), source::SeekError> {
--- a/src/decoder/symphonia.rs
+++ b/src/decoder/symphonia.rs
@@ -170,9 +170,7 @@ impl Source for SymphoniaDecoder {
 
     #[inline]
     fn total_duration(&self) -> Option<Duration> {
-        self.total_duration.map(|Time { seconds, frac }| {
-            Duration::new(seconds, if frac > 0.0 { (1f64 / frac) as u32 } else { 0 })
-        })
+        self.total_duration.map(time_to_duration)
     }
 
     fn try_seek(&mut self, pos: Duration) -> Result<(), source::SeekError> {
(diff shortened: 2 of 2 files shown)


dvdsk:
This seems to be a different issue. But definitely a bug and very nice to have a fix for it. Thank you!

I do not think we need the unit tests anymore now they pass. 

They only test the `time_to_duration` function and there is no reason for that code to ever change. If it does the tests will need to be re-written too. While that could be fixed by testing the `SymphoniaDecoder` with the field `total_duration` as Time with frac part zero. That is still overkill for this.

That does not mean writing unit tests is bad! They are useful as they show the code is correct, which is pretty important!

dvdsk:
One thing this does still need: a changelog entry! Cause you fixed a bug :)

Methapon2001:
Thanks for your reply, and sorry about a little lack in explanation.
I updated CHANGELOG.md now. let me know if the message is alright.

dvdsk (approved):

dvdsk:
merged :partying_face:

Issue #208, opened 2018-11-17
SineWave generates heavy amounts of ringing because of improper downsampling


mlindner:
This is one of the two issues found in #201.

With the fix in #202 what still occurs is when the input sinewave is 48000 (the default) and the automatically selected sound output device is using 44100 Hz and the rodio::sink or rodio::queue (fed through play_raw) are used then the output has heavy amounts of aliasing/ringing from improper downsampling to 44100 Hz.

This should be a perfect sine wave at 3000 Hz but instead has the banding seen which is quite audible.

<img width="615" alt="fixed roundnoff bad resampling" src="https://user-images.githubusercontent.com/2280117/48659222-b5755580-ea02-11e8-808c-8c0551011a72.png">

mlindner:
This is caused by the fact that the sample rate converter doesn't use a fourier transform in its sampling. I'm not sure what the intention of the sample rate converter is but it doesn't appear to be for converting sample rates.

Issue #626, opened 2024-10-03
User stories


dvdsk:
A collection of user stories (work in progress) to be used to inform (breaking) api changes.

dvdsk:
please do, we can always move it to its own issue if we have questions. And thank you very much!

dvdsk:
Something that seems very hard/impossible with rodio now:

> A user adds two songs to a queue. Then they enable cross-fade while the first song is almost at the end. Crossfade the works and the two songs fade into each-other.  
>
> The music plays without any stuttering while crossfade is toggled on and off.

bluenote10:
I'm simply adding a few use cases I had in the past -- without re-checking if they would currently be possible already.

#### Sample perfect scheduling

Think of a metronome app that needs to schedule its click sound dynamically (the user can control bpm e.g. via a slider), but with utmost temporal precision (you don't want any fluctuation in a metronome). This requires a to have sample/frame perfect scheduling, i.e., one would need to express _start playing this sound at sample index i=xxx_.

#### Sample perfect looping

Sometimes I had to loop sounds also without any gap in the audio output (note that waiting for the sound to finish, and re-schedule it again introduces a gap). This would require to set up a loop information that specifies same `(loop_sample_index_from, loop_sample_index_upto)` so that the playback can implicitly wrapped-around once the "upto" sample is reached. Ideally the playback interface would expose e.g. a `loop_counter` to allow users to query once a loop-wrap occurred. The most obvious use case for that are "infinite" background sounds that are constructed in a way that they loop around seamlessly, and simply setting the loop specification to `(0, last_sample)` would repeat the entire sound indefinitely.

#### Low-latency scheduling

I was once trying to implement a kind of drum synthesizer in rodio, where certain key-presses triggered certain sounds. However I noticed that the input latency was way too large for musical purposes. I would assume that mainly comes down to the large default buffer size, and #512 should improve that.

dvdsk:
> I'm simply adding a few use cases I had in the past 

Very helpful thank you!

dvdsk:
Thanks for all the feedback!

> I add my sources more 'dynamically'. Where once the playing song ends [a new one is appended](https://github.com/nednoodlehead/punge/blob/8019345295d667562758fb46293125333328af85/src/gui/subscription.rs#L444). So there is no point where I can apply it.

This is an issue I ran into myself too, good to hear others have trouble with it too. We should make this easy/doable and add a full example.

tramhao:
[...]

PR #646, opened 2024-12-01
Introduce a public constant for Rodio's default sample rate

PetrGlad:
This should help to reduce number of sample rate conversions. Setting it to current de-facto default (44.1kHz).

Also partially relieves #208. The rate converter has to be fixed still but with default sample rate less resampling is needed.

### DIFF
--- a/examples/mix_multiple_sources.rs
+++ b/examples/mix_multiple_sources.rs
@@ -2,10 +2,11 @@ use rodio::mixer;
 use rodio::source::{SineWave, Source};
 use std::error::Error;
 use std::time::Duration;
+use rodio::constants::DEFAULT_SAMPLE_RATE;
 
 fn main() -> Result<(), Box<dyn Error>> {
     // Construct a dynamic controller and mixer, stream_handle, and sink.
-    let (controller, mixer) = mixer::mixer::<f32>(2, 44_100);
+    let (controller, mixer) = mixer::mixer::<f32>(2, DEFAULT_SAMPLE_RATE);
     let stream_handle = rodio::OutputStreamBuilder::open_default_stream()?;
     let sink = rodio::Sink::connect_new(&stream_handle.mixer());
 
--- a/examples/noise_generator.rs
+++ b/examples/noise_generator.rs
@@ -1,6 +1,7 @@
 //! Noise generator example. Use the "noise" feature to enable the noise generator sources.
 
 use std::error::Error;
+use rodio::constants::DEFAULT_SAMPLE_RATE;
 
 #[cfg(feature = "noise")]
 fn main() -> Result<(), Box<dyn Error>> {
@@ -14,7 +15,7 @@ fn main() -> Result<(), Box<dyn Error>> {
     let interval_duration = Duration::from_millis(1500);
 
     stream_handle.mixer().add(
-        white(cpal::SampleRate(48000))
+        white(cpal::SampleRate(DEFAULT_SAMPLE_RATE))
             .amplify(0.1)
             .take_duration(noise_duration),
     );
@@ -23,7 +24,7 @@ fn main() -> Result<(), Box<dyn Error>> {
     thread::sleep(interval_duration);
 
     stream_handle.mixer().add(
-        pink(cpal::SampleRate(48000))
+        pink(cpal::SampleRate(DEFAULT_SAMPLE_RATE))
             .amplify(0.1)
             .take_duration(noise_duration),
     );
--- a/examples/signal_generator.rs
+++ b/examples/signal_generator.rs
@@ -1,6 +1,7 @@
 //! Test signal generator example.
 
 use std::error::Error;
+use rodio::constants::DEFAULT_SAMPLE_RATE;
 
 fn main() -> Result<(), Box<dyn Error>> {
     use rodio::source::{chirp, Function, SignalGenerator, Source};
@@ -11,7 +12,7 @@ fn main() -> Result<(), Box<dyn Error>> {
 
     let test_signal_duration = Duration::from_millis(1000);
     let interval_duration = Duration::from_millis(1500);
-    let sample_rate = cpal::SampleRate(48000);
+    let sample_rate = cpal::SampleRate(DEFAULT_SAMPLE_RATE);
 
     println!("Playing 1000 Hz tone");
     stream_handle.mixer().add(
--- a/src/constants.rs
+++ b/src/constants.rs
@@ -1,0 +1,6 @@
+//! Rodio common constants
+
+/// The sample rate (Hz) output stream uses by default.
+/// Prefer this one in the processing chain to minimize sample rate conversions.
+pub const DEFAULT_SAMPLE_RATE: u32 = 44_100;
+
--- a/src/conversions/mod.rs
+++ b/src/conversions/mod.rs
@@ -2,7 +2,6 @@
 This module contains function that will convert from one PCM format to another.
 
 This includes conversion between sample formats, channels or sample rates.
-
 */
 
 pub use self::channels::ChannelCountConverter;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -166,6 +166,7 @@ pub mod mixer;
 pub mod queue;
 pub mod source;
 pub mod static_buffer;
+pub mod constants;
 
 pub use crate::conversions::Sample;
 pub use crate::decoder::Decoder;
--- a/src/source/sine.rs
+++ b/src/source/sine.rs
@@ -1,5 +1,5 @@
 use std::time::Duration;
-
+use crate::constants::DEFAULT_SAMPLE_RATE;
 use crate::source::{Function, SignalGenerator};
 use crate::Source;
 
@@ -7,19 +7,17 @@ use super::SeekError;
 
 /// An infinite source that produces a sine.
 ///
-/// Always has a rate of 48kHz and one channel.
+/// Always has default sample rate and one channel.
 #[derive(Clone, Debug)]
 pub struct SineWave {
     test_sine: SignalGenerator,
 }
 
 impl SineWave {
-    const SAMPLE_RATE: u32 = 48000;
-
     /// The frequency of the sine.
     #[inline]
     pub fn new(freq: f32) -> SineWave {
-        let sr = cpal::SampleRate(Self::SAMPLE_RATE);
+        let sr = cpal::SampleRate(DEFAULT_SAMPLE_RATE);
         SineWave {
             test_sine: SignalGenerator::new(sr, freq, Function::Sine),
         }
@@ -48,7 +46,7 @@ impl Source for SineWave {
 
     #[inline]
     fn sample_rate(&self) -> u32 {
-        Self::SAMPLE_RATE
+        self.test_sine.sample_rate()
     }
 
     #[inline]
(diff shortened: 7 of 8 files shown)


PetrGlad:
A more advanced solution can be for processing steps to query current preferred  sample rate and use that. With such API one could for example open output stream with 48kHz and all processing steps attached to it will adjust accordingly. Or have sample rate a parameter for every processing step that may be affected by it. Which cumbersome but could help keeping 48kHz everywhere if one needs that.
But these are a lot bigger changes, and I do not have a clear idea how to implement that.

dvdsk:
> A more advanced solution could allow processing steps to query current preferred sample rate and use that.

that could look like:

```rust
impl trait Source for Something {
    fn negotiate_samplerate(&mut self, requested_by_parent: usize) {
        self.inner.negotiate_samplerate(requested_by_parent)
    }
}
```

Then the output calls `negotiate_samplerate` and the source(s) adjusts if they can.

We would need to add an api that allows specifying a different sample rate. That could nicely fit into the `Source` model too. It could be something like:

```rust
struct FixedSampleRate<S>{ 
   rate: usize,
   inner: S,
}

impl Source for FixedSampleRate {
    fn negotiate_samplerate(&mut self, requested_by_parent: usize) {
        debug!("FixedSampleRate set, ignoring samplerate requested by parent")
        self.inner.negotiate_samplerate(self.rate)
    }
}

trait Source {
    fn with_samplerate(self, rate: usize) -> Self {
        FixedSampleRate {
            rate,
            inner: self,
        }
    }
}
```

This might even allow us to improve mixer (which currently resamples both inputs) to use one sample-rate without resampling if possible. For example when mixing a Sine wave or Noise with a decoder (song playing from disk).

dvdsk:
related: https://github.com/RustAudio/rodio/issues/208

PetrGlad:
The intention for the constant was to document the state of affairs (see default stream initialization), It may not make sense for v1.0. I hoped that fully qualified name `rodio::constants::DEFAULT...` would help to emphasis that it is the rodio's default. Until we have dynamic way to negotiate sample rates, this constant may be helpful. I would not insist on this change, though.

PetrGlad:
Yes, empty callback does nothing, but the receiving side may still want to setup conversion before receiving samples which would be unnecessary if rates match.

dvdsk:
> Until we have dynamic way to negotiate sample rates

I want to play with this soon given it might also be an answer to #493 (having empty get the channel count instructed from the parent/mixer its part of).

PetrGlad:
I'd prefer to have a constructor or builder to handle that since it allows users to have complete control over it if they want. E.g.
```rust
// Or maybe can use rodio::stream::StreamConfig here instead.
struct StreamShape { 
  sample_rate: u32,
  channels: u8
}

impl ASource {
  pub fn new(params: &ASourceConf, conf: &StreamShape) -> ASource {
    todo!()
  }

  pub fn new_connected(params: &ASourceConf, sink: &mut Sink) {
        mixer.add(Self::new(mixer.input_shape()));
  }
}
```

dvdsk:
> I'd prefer to have a constructor or builder to handle that since it allows users to have complete control over it if they want. Something like
> 
> ```rust
> // Or maybe can use rodio::stream::StreamConfig here instead.
> struct StreamShape { 
>   sample_rate: u32,
>   channels: u8
> }
> 
> impl ASource {
>   pub fn new(params: &ASourceConf, conf: &StreamShape) -> ASource {
>     todo!()
>   }
> 
>   pub fn connect_new(params: &ASourceConf, sink: &mut Sink) {
>         mixer.add(Self::new(mixer.input_shape()));
>   }
> }
> ```

Complete control is still possible, in principle every source should try to adhere to whatever its parent/predecessor/sink/wrapping source (we really got to note down what we call this) decided. But a source can decide to pass on its own values. We can add a source that does this with a value given by the end user. That is what I meant here: https://github.com/RustAudio/rodio/pull/646#issuecomment-2511677332

dvdsk:
a more advanced design could present multiple _sample-rates_ during _negotiation_. In some order of preference. Maybe then the sources could negotiate with the playback hardware and find the optimal sample rate.

PetrGlad:
I'd prefer to have sample rate as a construction parameter. In case a source needs frames/chunks with output formats varying over time, that can be plugged into `UniformSourceIterator` or its equivalent. Do you have in mind any other cases when a source could not accept sample rate as a parameter?

Given your example a stream parameters can be passed as:
```rust
let output = OutputStreamBuilder::open_default_stream();

let sine = Signalgenerator(Sine);
// Can be "shape" instead of "format", maybe. I like the word but frequency is not really a shape...
// Note that here is only one conversion step.
let source = Decoder::new("file.mp3") .format(output.format()) .pausable() .amplify() .mix(sine);

output.mix(source);
```
A bit inconvenient, maybe, but the parameter setting on each step is automatic. Each step's constructor knows format of its inputs and all the linking can be checked at construction time.

It can be done in reverse, maybe more convenient but more complex, especially for sinks that have dynamic set of inputs. `output.mix(source)` is called last so the outputs will have to try to call format setters on inputs recursively.

PetrGlad:
CPAL does give a set of possible audio formats so in theory one can calculate intersection between one's preferences and what is available, but I suspect it may be a bit complicated.

dvdsk:
> CPAL does give a set of possible audio formats so in theory one can calculate intersection between one's preferences and what is available, but I suspect it may be a bit complicated.

I like to try, it would be the ultimate solution right? We would minimize unneeded re-sampling and the user would not even notice (except less artifacts more perf).

One hard thing might be decoders, they might change samplerate mid song. The negotiation might need to happen multiple times to support that....

PetrGlad:
@dvdsk Yes, to get ultimate reduction in conversion one can do that from the both ends, enumerate all the inputs and see what output device can do and then pick some stream format that satisfies them all, but it will not always be possible. As I see it requiring inputs to convert to some common format and then just using that would be a meaningful improvement already. Alternatively if inputs are all of the same format then, use that and add a converter only at the end of the chain just before the output stream only if necessary.
What I do not like at the moment is that the code is unsure what comes as input and tries to adapt to that on every step all the time. This complicates code significantly since now every sink implementation has to handle format changes (see `.current_frame_len(&self)`). And this likely has noticeable performance cost, if we want to keep an eye on that.

I believe all the sources that do change format are somehow external to rodio, are its inputs or close to inputs. That is why as I mentioned earlier I am for having also simpler `Souce` version that has predictable format. So anyone consuming it can rely on it, and simplify its processing accordingly. 
Internal cases probably may change number of channels dynamically, so it should be dedicated processing step (like `ChannelCountConverter`)

PetrGlad:
CPAL format enumeration is cumbersome so some sort of query function could be welcome (when we can filter formats with some criteria or a predicate). And ideally it should be in CPAL not in rodio.

A crazy implementation would insert all formats into SQLite and query that....

dvdsk:
hope your okay with the rename, you and I have said some valuable stuff here and I want it easier to find. Thats why I renamed it

dvdsk:
from the user stories [thread](https://github.com/RustAudio/rodio/issues/626#issuecomment-2525441320)

> Also, I've used the crate [Hodaun](https://github.com/kaikalii/hodaun) as an alternative to Rodio in the past for some of the features it has, mostly automation of parameters. Could be a good place to look for desirable features.

from hodaun readme, section _differences to rodio_:

> In Hodaun, the sample rate is passed to Source, and it is up to the Source to decide what to do with it. In practice, this means that the sample rate of Sources adapts to the sample rate of the audio output device.

That sounds like my proposal here: https://github.com/RustAudio/rodio/pull/646#issuecomment-2511677332.

and this:

> Sources that can have various possible sample rates, like audio input devices or audio files, implement UnrolledSource instead of Source. This trait corresponds more closely to Rodio's Source trait. An UnrolledSource can be converted to a Source using UnrolledSource::resample.

sounds like what you proposed.

PetrGlad:
Yes, I'd also like to pass preferred stream parameters to a source during construction. I do not like the "negotiation" part (if I understand it correctly), that is potentially changing parameters at some point later after construction is complete. I am afraid that may complicate the logic a lot.

PetrGlad:
I would also like to process sample frames instead. At least it should be a strict requirement that source always produces whole frames (does not end after passing only some of the channels).

dvdsk:
> that is potentially changing parameters at some point later after construction is complete

It is complex but it would allow us to eliminate resampling when the hardware can take care of it for us. I think the complexity is not that high rodio's sources are a simple tree. The negotiation does not need to be perfect, if it can cover the most common case (song playing through decoder) I think it might be worth it already. Not having to resample will save quite some battery on low power devices.

dvdsk:
I'll work on it once we have a better resampler in rodio, and we can always decide not te merge. But I think we need to see how it would work out before deciding if this is something we want in rodio. I only have a vague plan right now and most of the complexity will only become clear once I start implementing it.

PetrGlad:
@dvdsk  The formats are "suggested", so to say, by both sources and the output device. A source's format can change but stream has to be re-initialized to change that. In my opinion the approach when explicit stream format is passed to the sources is the best (see https://github.com/RustAudio/rodio/pull/646#issuecomment-2521376225). Chasing perfect score does not seem worth it as dynamic changes will complicate the logic. It is not about topology but difficulties similar to handling `.current_frame_len()` (I still suspect that #493 is caused by this in either rodio's or the user's code) which otherwise can be completely avoided. 
Also the questions remain for me. For example, what if we have several mp3 sources each having own format and a particular time how do we handle that? The sources and output stream should vote which format is the best? How do we change output format? By transparently re-opening output stream?

PetrGlad:
As a side note it is possible to have a generic adapter that helps for any source to handle `.current_frame_len()`. Namely, it would need to get the wrapped Source constructor function and create new source instance when the input format changes. So it would pass `current_frame_len` of samples to the wrapped source, and then re-create it with the new input stream format. Maybe it is cumbersome for rodio's embedded sources but may help users.

dvdsk:
> As a side note it is possible to have a generic adapter that helps for any source to handle `.current_frame_len()`. Namely, it would need to get the wrapped Source constructor function and create new source instance when the input format changes. So it would pass `current_frame_len` of samples to the wrapped source, and then re-create it with the new input stream format. Maybe it is cumbersome for rodio's embedded sources but may help users.

I do not really get how that would work, is it something you can easily demonstrate with example code?

dvdsk:
> but stream has to be re-initialized to change that.

I did not know that, it makes senses though.

> Chasing perfect score does not seem worth it as dynamic changes will complicate the logic.

you are completely right here. Perfect is the enemy of good. I think if we do this, we should do it as a slight optimization on the current situation. Most sources are a constant sample rate we just do not know the rate at compile time. So if we do a single negotiation at creation we can determine a good `sample_rate` for most use-cases. We can add a method to override easily. We should be able to make that override do what you propose (inject a sample_rate from the output). better (less unneeded resampling) for the vast majority of users.

PetrGlad:
@dvdsk Yes, now I admit that in Rust such an adapter will be more complicated then I thought or maybe I do not understand Rust well enough yet. In a garbage collected language this would be more straightforward. The filter that exists only for a span should be given temporary source instance since it is supposed to own it, and from other side of it there should be some source that can be dynamically attached to another source (when span changes). So I stopped half way :) Maybe will give it another try if I get a clearer idea.

Issue #647, opened 2024-12-01
Adapt a higher quality resampler


PetrGlad:
Currently implemented sample rate converter uses simple linear interpolation algorithm and is prone to producing high frequency noise in output. See for example #584 and #316 related to the resampler implementation. However the algorithm is fast and requires little resources, so we should keep it at least as an option.

We should integrate an existing sample rate converter and let users to pick an implementation depending on their needs. A simple optional low-pass filter may be also implemented as post-processing step for existing Rodio's SampleRateConverter to improve its audio quality.

It seems that there are 2 implementations that we can use:
1. [Rubato](https://github.com/HEnquist/rubato)
2. [rust-samplerate](https://github.com/Prior99/rust-samplerate) which uses C bindings to [libsamplerate](http://www.mega-nerd.com/SRC/api.html)

dvdsk:
> A simple optional low-pass filter may be also implemented as post-processing step for existing Rodio's SampleRateConverter to improve its audio quality.

I like this idea :+1: might give a good middle ground between low quality fast and high quality but slow. Lets first see what the perf difference the different resamplers. 

I've already looked at rubato and it looks great. I'm not sure about rust-samplerate. I do not like adding a C-dependency. Those complicate cross-compiling a lot.....

Instead we could try the resampler in dasp: https://docs.rs/dasp/latest/dasp/interpolate/sinc/struct.Sinc.html.

Losses:
This is, indeed a issue.

In our investigation into the playback quality of Rune, our player, which used Rodio as the playback backend. To rule out software-related influences, we controlled all external parameters:

- We generated an audio file with a sample rate of 44100Hz, consisting of a 1000Hz mono sine wave.
- Using OBS for recording, we set the recording parameters to 44100Hz to avoid resampling issues.
- During the recording process, both the audio player's volume and the system volume were set to 100%.

This setup resulted in a clean, smooth spectrum, confirming that there are no inherent development errors within Rune affecting audio quality.

![Clean Spectrum](https://github.com/user-attachments/assets/a0fbadf8-b1c7-417e-832f-f5ccbc7616f8)

However, we observed that Rodio's resampling process has flaws, particularly when downsampling from 48000Hz to 44100Hz, leading to suboptimal signal quality.

![Resampling Issue](https://github.com/user-attachments/assets/845a571e-0344-4324-a51a-e11d790c7a91)

[...]

PR #655, opened 2024-12-07
[not for merge] Resampler low pass prototype

PetrGlad:
For a context, we plan to introduce a high quality resampler (#647). This change is an attempt to maybe also improve existing default resampler by keeping it light but improving its sound quality slightly.

Here are some resampler changes prototype that I wanted trying. See `low_pass::tests::test_low_pass`, if this will be deemed sufficiently good, down-sampling step can be optimized (just use decimation).
I would like to have some objective evidence on how it performs, probably need some test harness for that. Or someone with better ears could listen to the output to compare asses it subjectively.
The low pass filter there is primitive has very slope edge, has cut-off frequency at about sample rate / 2 and may distort audio by boosting lows so the audio is up-sampled to 2 x target frequency first, and output is amplified by 0.5.
Low pass filter can be improved, but that will be slower as well. Maybe there are some people who are experienced in DSP so suggest better quick fix ideas.

### DIFF
--- a/src/source/low_pass.rs
+++ b/src/source/low_pass.rs
@@ -1,0 +1,115 @@
+use std::time::Duration;
+use num_rational::Ratio;
+use crate::conversions::SampleRateConverter;
+use crate::{Sample, Source};
+
+pub struct LowPass<I>
+where
+    I: Iterator,
+{
+    input: I,
+    prev: Option<I::Item>
+}
+
+impl<I> LowPass<I>
+where
+    I: Iterator,
+    I::Item: Sample,
+{
+    #[inline]
+    pub fn new(
+        mut input: I,
+    ) -> LowPass<I> {
+        LowPass {
+            input,
+            prev: None
+        }
+    }
+}
+
+impl<I> Source for LowPass<I>
+where
+    I: Source,
+    I::Item: Sample,
+{
+    fn current_frame_len(&self) -> Option<usize> {
+        None
+    }
+
+    fn channels(&self) -> u16 {
+        self.input.channels()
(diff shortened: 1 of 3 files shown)


src/source/low_pass.rs
dvdsk, line 1: luckily we already have a low_pass filter, see:
https://docs.rs/rodio/latest/rodio/source/trait.Source.html#method.low_pass

PetrGlad:
I have added the prototypes into src/conversions/sample_raters tests, see [conversions::sample_rate::test::resampler_tweaks()](https://github.com/PetrGlad/rodio/blob/e13d3c4429a4802f3c2594c806a56d7e13240c62/src/conversions/sample_rate.rs#L454). You can play with the implementation and the input parameters. The test runs 4 re-sampler versions and stores output to a wav file. The versions tested:
* Bare existing implementation, for comparison.
* With primitive low-pass filter.
* With primitive low pass filter but with 2x up-sampling before filtering
* With existing bi-quad filter from `blt` module.

I hoped that primitive filter although not being good still can make meaningful difference. This does not seem to be the case.

// I wonder if we should add a chirp (frequency sweep) signal to be used in testing.

yara-blue:
Hey Petr, sorry for letting this one sit for so long. 

We are about to add rubato as a high quality resampler however there are a few others out there build for performance. I agree there is absolutely a use case for a lower quality but higher performance resampler. Maybe it makes senses to add some of those other resamplers as an option. Roderick has some context on that.

For now I'm closing this prototype to clean up the PR list.

Issue #653, opened 2024-12-06
Channel router


dvdsk:
@PetrGlad you brought this up somewhere else and I just saw it has actually been requested a few weeks ago: https://github.com/RustAudio/rodio/issues/626#issuecomment-2470118600

So lets discuss that here.

iluvcapra:
Can I write this? I have opinions on this.

PetrGlad:
Yes, this is something I had in mind. We do have a channel number converter (`ChannelCountConverter`) but its primary function seems to be mono->stereo or just discarding channels with higher numbers. It does not look very useful otherwise.
So here I see 3 options.

:one: The most generic channel router would be to have volume/amplification matrix so every output is a weighed sum of inputs. That is for $n$ input channels, and  $m$ output channels, there would be a matrix $A_{m,n}$ with floating point weights (those could even be negative :) This version could handle arbitrary routing, mixing, channel duplication, mono conversion, etc. But this one would be the slowest of three, although I would not worry about that much as it likely can be optimized. This version can probably needs coefficients to be floating point or at least fractional, because it can sum channels together which may cause clipping so one may want to adjust input weights.

:two: A simpler implementation could use just single weighted input for every output. In that case an output is specified as pair of `{input_channel_index, weight}`, and the router would have a list of those pairs one for every output channel. This version can still handle channel routing (including duplication), and adjusting per-channel volume.

:three:  A simpler still implementation of this would just route channel (equivalent to limiting all coefficients in previous version to `1.0`). This only would have to use a list of input channel indexes, one for every output channel. This version is useful for arbitrary channel swaps, duplication, discarding channels but not for mixing.

@iluvcapra All of these versions look useful to me, so maybe you can pick whatever solves your problem the best. Maybe there are something else that I have missed.

Even simpler but inconvenient would be to use single channel converter and some kind of splitter, then single channel converter router, and then mixer. But this sounds rather cumbersome even to me :)

dvdsk:
> Can I write this? I have opinions on this.

have at it :)

It seems like Petr also has some thoughts so they'll probably be able to help out if you need some input along the way.

PetrGlad:
By the way https://github.com/RustAudio/rodio/issues/626#issuecomment-2470118600

PetrGlad:
After thinking more my personal preference would be :two: without limiting how many pairs can be added. Since it would support most if not all cases I saw so far. If there will be need for more complex processing :one: can be implemented separately.

[...]

PR #656, opened 2024-12-07
ChannelRouter feature

iluvcapra:
This is an implementation of a multichannel router/mixer source. See #653.


examples/channel_routing.rs
```diff
@@ -0,0 +1,43 @@
...
+    stream_handle.mixer().add(router);
+
+    for i in 0..1000 {
+        thread::sleep(interval_duration);
+        let n = i % 20;
+        match n {
+            0 => println!("Left speaker ramp up"),
+            1..10 => {
+                _ = controller.map(0, 0, n as f32 / 10.0);
+                _ = controller.map(0, 1, 0f32);
```
dvdsk: this is not very readable right? I can guess its channel from, channel too, gain? But we might be able to do better. I see a number of options, there might well be more:
1️⃣ make map take a struct like: `controller.map(Mapping { from: 0, to: 1, gain: 0f32 })`

2️⃣ make map a builder: `controller.map().from(0).to(1).with_gain(0.f32).appy()`. Here you could make the gain optional and do a default of 1.0

3️⃣ rename it so the argument names are in the name: `controller.map_from_to_with_gain(0, 1, 0f32)`.

I think 2 is my favorite, though its a little more code.

iluvcapra:
BTW I'm working on a separate project right now to try to profile all these different approaches, once I have that I'll post what I find and then we can move forward with this.

examples/channel_routing.rs
iluvcapra: I like the builder too.

src/source/channel_router.rs
```diff
@@ -0,0 +1,306 @@
...
+                self.input.current_frame_len().unwrap_or(usize::MAX),
+            );
+
+            // fill the input buffer. If the input is exhausted and returning None this will make
+            // the input buffer zero length
+            self.input_buffer = self.inner_mut().take(samples_to_take).collect();
+
+            self.current_channel = 0;
+
+            for change in self.receiver.try_iter() {
```
PetrGlad, line 222: Yes, it may do as first version, but polling channel is relatively expensive. In other places an additional flag is used to skip the loop altogether when there are no changes (e.g. `mixer::Mixer::has_pending`). See also https://github.com/RustAudio/rodio/issues/658.

src/source/channel_router.rs
```diff
@@ -0,0 +1,306 @@
...
+            .iter()
+            .zip(&self.channel_map)
+            .map(|(in_sample, input_gains)| {
+                // the way this works, the input_buffer need not be totally full, the router will
+                // work with whatever samples are available and the missing samples will be assumed
+                // to be equilibrium.
+                let gain = input_gains[self.current_channel as usize];
+                in_sample.amplify(gain)
+            })
+            .reduce(|a, b| a.saturating_add(b));
```
PetrGlad, line 239: Since this is real-time I'd prefer this to be more explicit. Maybe the matrix should be transposed to streamline this. This way instead of zip it would select a gain vector and then do the dot product of coefficients by input_buffer (it can be even optimized with SIMD).

src/source/channel_router.rs
```diff
@@ -0,0 +1,306 @@
...
+    /// channel in the output may be a mix of any number of channels in the input.
+    ///
+    /// Successive calls to `map` with the same `from` and `to` arguments will replace the
+    /// previous gain value with the new one.
+    pub fn map(
+        &mut self,
+        from: u16,
+        to: u16,
+        gain: f32,
+    ) -> Result<(), ChannelRouterControllerError> {
```
PetrGlad, line 77: :question:  If the routing has to change one will have to remember to clear all the gains that are not needed anymore. 
Maybe a "reset" function that sets all the gains to zero would be helpful.

:question: I wonder if we should allow specifying matrix at the initialization time, or let users to pass a list of coefficients.

src/source/channel_router.rs
```diff
@@ -0,0 +1,306 @@
...
+    sync::mpsc::{channel, Receiver, Sender},
+};
+
+/// A matrix to map inputs to outputs according to a gain
+///
+/// A two-dimensional matrix of `f32`s:
+/// - The first dimension is respective to the input channels
+/// - The second is respective to the output channels
+///
+/// Thus, if a value at `map[1,1]` is 0.2, this signifies that the signal on
```
PetrGlad, line 17: It is still confusing, I'd picked different numbers for input and output, or maybe just use names instead of constants (e.g. `map[in,out]`)

src/source/channel_router.rs
```diff
@@ -0,0 +1,306 @@
...
+/// channel 1 should be mixed into channel 1 with a coefficient of 0.2.
+pub type ChannelMap = Vec<Vec<f32>>;
+// doing this as Vec<Vec<atomic_float::AtomicF32>> would require feature=experimental, so I decided
+// to just use a channel to do updates.
+//
+// Doing it as a HashMap<(u16,u16), f32> is an option too but there's a penalty hashing these
+// values, there's ways to speed that up though. It'd be great if the object upgraded its
+// implementation if it got sufficiently big.
```
PetrGlad, line 25: No we do not need complex containers, if a sparse matrix is preferable I have an [example](https://github.com/RustAudio/rodio/issues/653#issuecomment-2526344738) how that can be implemented.

src/source/channel_router.rs
```diff
@@ -0,0 +1,306 @@
...
+
+    #[inline]
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.current_channel >= self.channel_count {
+            // We've reached the end of the frame, time to grab another one from the input
+            let input_channels = self.input.channels() as usize;
+
+            // This might be too fussy, a source should never break a frame in the middle of an
+            // audio frame.
+            let samples_to_take = min(
```
PetrGlad, line 211: Yes, I'd prefer to require all the sources to produce only complete frames. But I think it is not actually enforced anywhere so far. This can be an assertion, though.

src/source/channel_router.rs
PetrGlad: Probably that is why I wanted to set the whole gain set at once: users may have the matrix on their side process and inspect it a they prefer and then use its to update the router.

src/source/mod.rs
```diff
@@ -346,6 +348,103 @@ where
...
+    #[inline]
+    fn mono_to_stereo(self) -> ChannelRouterSource<Self>
+    where
+        Self: Sized,
+    {
+        let mapping: ChannelMap = vec![vec![0.5f32, 0.5f32]];
+        channel_router::channel_router(self, 2, mapping).1
+    }
+
+    /// Creates a [`ChannelRouter`] that mixes a 5.1 source in SMPTE channel order (L, R, C, Lfe, Ls,
```
dvdsk, line 423: very nice example use-case :+1:

src/source/channel_router.rs
dvdsk: I do not want to overburden @iluvcapra, this is already quite a big effort! So we can always leave optimizations for later and merge this first. Someone else can work on them too (maybe @PetrGlad). Let me know what you prefer @iluvcapra.

dvdsk:
> BTW I'm working on a separate project right now to try to profile all these different approaches, once I have that I'll post what I find and then we can move forward with this.

very cool, I'm curious to see if what ends up as best lines up with our suspicions

PetrGlad:
For mixing several input streams we can use similar class that uses the same algorithm but accepts several Sources at once. Our existing mixer only sums all inputs without any volume adjustments,

dvdsk:
> For mixing several input streams we can use similar class that uses the same algorithm but accepts several Sources at once. Our existing mixer only sums all inputs without any volume adjustments,

Isn't the better way to do this to have users wrap the streams they are mixing in a `Volume` struct?

PetrGlad:
Yes, it is a simpler solution. Maybe we could add and example or make some kind of cookbook.

I was just pondering whether an über-mixer might be useful (the one like in this PR but allowing multiple inputs). For sparse matrix representation this would be straightforward to adapt (or have another more advances version).

By he way, the router in this PR should also handle situation when number of input channels change. At least for current API this is required.

dvdsk:
> make some kind of cookbook.

I really like that idea! Sink/Player will be easy to use but the power of rodio lies in connecting the sources in an audio pipeline. We could have a bunch of examples but a cookbook is way nicer!

iluvcapra:
I've pulled this up to the upstream, we can merge it or I can keep working on it. I've been doing my active tinkering in my (unpublished) benchmark project but this all works, I've included an example as well.

dvdsk:
> I've pulled this up to the upstream, we can merge it or I can keep working on it. I've been doing my active tinkering in my (unpublished) benchmark project but this all works, I've included an example as well.

It sounds like your still having fun? As long as that's true keep it up. We can merge it whenever your done with it, we have all the time in the world. If this PR becomes a chore then let us know, we don't want that open source has to be fun.

examples/channel_routing.rs
PetrGlad: In case of builder `.from(0).to(1).with_gain(0.f32)` should always go together at least it should be single call with 3 parameters.

I'd prefer a shorter name than `controller.map_from_to_with_gain(0, 1, 0f32)` since it is the only function supported by the router. maybe even `controller.link(0, 1, 0f32)` what it links can be figured out from the parameter names:
```rust
fn link(from_channel: ChannelNumber, to_channel: ChannelNumber, gain: f32)
```

src/source/channel_router.rs
PetrGlad: In my benchmarks mpsc polling was tens of microseconds, atomic read - several nanoseconds.

PetrGlad:
Just to show what I had in mind in my proposals before. #671

PR #657, opened 2024-12-08
Feature: Channel bitmasks

iluvcapra:
Create a channel bitmask mechanism to allow sources to tell their consumers how their channels may map to speakers.

### DIFF
--- a/src/channel_bitmask.rs
+++ b/src/channel_bitmask.rs
@@ -1,0 +1,49 @@
+
+pub type ChannelBitmask = u64;
+
+const FRONT_LEFT: ChannelBitmask = 0x1;
+const FRONT_RIGHT: ChannelBitmask = 0x2;
+const FRONT_CENTER: ChannelBitmask = 0x4;
+const LFE: ChannelBitmask = 0x8;
+const BACK_LEFT: ChannelBitmask = 0x10;
+const BACK_RIGHT: ChannelBitmask = 0x20;
+const FRONT_LEFT_CENTER: ChannelBitmask = 0x40;
+const FRONT_RIGHT_CENTER: ChannelBitmask = 0x80;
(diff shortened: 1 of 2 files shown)


dvdsk:
Might need an example since I do not completely get how this will work. 

Is the idea that decoders are going to implement the new trait? In that case: https://docs.rs/symphonia-core/0.5.4/symphonia_core/codecs/struct.CodecParameters.html#structfield.channel_layout might be relevant. This matches the order assumed by channelcountconverter. I do not know what the other decoders do...

iluvcapra:
Oh good, I was looking for a file source that made this information available.

Wave files should make this available if they have > 2 channels, and the channel bitmaps are based on what Wave uses, but Hound doesn't seem to make this information available. I see what struct would have it and they just don't have a slot for it.

iluvcapra:
In fact this works almost exactly like symphonia's `Channels` type, I've just defined more values.
 https://docs.rs/symphonia-core/0.5.4/src/symphonia_core/audio.rs.html#29-90

PetrGlad:
Channel name mapping (name->channel_number) can be part of the stream format, so filters can use it if they need. We may require that filters pass it through. Just saying...

iluvcapra:
> Channel name mapping (name->channel_number) can be part of the stream format (it's metadata), so filters can use it if they need. We may require that filters pass it through. Just saying...

I'd see this PR as "transitional" but yes `channel_bitmask()` needs to be part of `trait Source`, and should be, it's very necessary for working with multichannel audio.

dvdsk:
> I'd see this PR as "transitional" but yes `channel_bitmask()` needs to be part of `trait Source`, and should be, it's very necessary for working with multichannel audio.

I'm okay with that, is there a default implementation you could give `Source`?

dvdsk:
let me know if this is ready for review.

PR #661, opened 2024-12-19
Add an option to output to a WAV file

PetrGlad:
Add a function that can store a Source's output to a WAV file. This will help with diagnostics and testing. For example to see whether a problem with playback is on `cpal`'s or `rodio`'s side.  The output file can then be analyzed with external software or played back. Also output is faster than real-time so this may speed-up development or testing.

Ideally this should be a part of the output stream builder but then we'll need to create a mixer and await for some input, which may be tricky. Also to make this transparent this would require to map `cpal` sample formats to `hound` parameters. It seems that this is not worth it until there are real use cases that need this.

### DIFF
--- a/src/file_output.rs
+++ b/src/file_output.rs
@@ -1,0 +1,39 @@
+use crate::{Sample, Source};
+use hound::{SampleFormat, WavSpec};
+use std::path;
+
+/// This procedure saves Source's output into a wav file. The output samples format is 32-bit float.
+/// This function is intended primarily for testing and diagnostics. It can be used to see
+/// the output without opening output stream to a real audio device.
+pub fn output_to_wav<S: Sample, P: AsRef<path::Path>, Src: Source<Item = S>>(
+    source: &mut Src,
+    wav_file: &P,
+) -> Result<(), Box<dyn std::error::Error>> {
+    let format = WavSpec {
+        channels: source.channels(),
+        sample_rate: source.sample_rate(),
+        bits_per_sample: 32,
+        sample_format: SampleFormat::Float,
+    };
+    let mut writer = hound::WavWriter::create(wav_file, format)?;
+    for sample in source {
+        writer.write_sample(sample.to_f32())?;
+    }
+    writer.finalize()?;
+    Ok(())
+}
+
+#[cfg(test)]
+mod test {
+    use super::output_to_wav;
+    use crate::Source;
+    use std::time::Duration;
+
+    #[test]
+    fn test_output_to_wav() {
+        let mut new_source = crate::source::SineWave::new(745.0)
+            .amplify(0.1)
+            .take_duration(Duration::from_secs(1));
+        output_to_wav(&mut new_source, &"target/tmp/save-to-wav-test.wav").unwrap();
+    }
+}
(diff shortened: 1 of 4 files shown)


src/file_output.rs
```diff
@@ -0,0 +1,39 @@
+use crate::{Sample, Source};
+use hound::{SampleFormat, WavSpec};
+use std::path;
+
+/// This procedure saves Source's output into a wav file. The output samples format is 32-bit float.
+/// This function is intended primarily for testing and diagnostics. It can be used to see
+/// the output without opening output stream to a real audio device.
+pub fn output_to_wav<S: Sample, P: AsRef<path::Path>, Src: Source<Item = S>>(
+    source: &mut Src,
+    wav_file: &P,
```
dvdsk, line 10: `wav_file: impl AsRef<Path>`, you can do the same for source, then you only need the generic `parameter S` for sample. Its more readable since you do not need to check what &P is

src/file_output.rs
```diff
@@ -0,0 +1,39 @@
...
+}
+
+#[cfg(test)]
+mod test {
+    use super::output_to_wav;
+    use crate::Source;
+    use std::time::Duration;
+
+    #[test]
+    fn test_output_to_wav() {
```
dvdsk, line 33: cool integration test could be: 
- generate some source in rodio (sine). 
- write it out to wav
- read it in 
- write the just read source out to another file
- check if they are identical (given wav is lossless they should be)

dvdsk: > Also to make this transparent this would require to map cpal sample formats to hound parameters. It seems that this is not worth it until there are real use cases that need this.

ahh yeah I understand what you meant here now. Lets not then. Lets keep this simple :+1:

dvdsk:
great idea! very valuable for debugging, I can also think of a few use cases that this will serve. I don't see any problems with this. Take a look at the parameter suggestion and then merge :+1:

src/file_output.rs
PetrGlad: Do you mean
```rust
pub fn output_to_wav<S: Sample>(
    source: &mut dyn Source<Item=S>,
    wav_file: &dyn AsRef<path::Path>,
) 
```
?

dvdsk: No normal static dispatch but using a different notation:
```rust
pub fn output_to_wav<S: Sample>(
    source: &mut impl Source<Item=S>,
    wav_file: &impl AsRef<path::Path>,
) 
```

dvdsk: actually I do not think you need the & in front of `impl AsRef<path::Path>`, since if AsRef is implemented for `thing` its also implemented for `&thing`

PetrGlad: // It was a misprint, it should have been `&dyn AsRef`

`AsRef` is a trait so it is unsized... If we want it to keep generic it should be either `dyn` or a template parameter? Can you give an example of what you have in mind?

PetrGlad: @dvdsk Updated.

iluvcapra:
Yay!

dvdsk:
@PetrGlad you can merge this when your done with it (you should be able to now). 

Alternatively we can make it a guideline that we try not merge our own changes (unless they are trivial). The advantage to that is it guarantees we both sign off on the changes. 

Let me know what you think of it then. I'll add `MAINTAINING.md` mentioning it we decide on some guidline.

PetrGlad:
Yes, I'd like to always have at least one other person to always look at the change.

dvdsk:
We both forgot it needs a changelog entry, I'll add one and push it to master.

Issue #660, opened 2024-12-13
rename frame -> span


dvdsk:
See discussion in https://github.com/RustAudio/rodio/issues/652

Should be renamed everywhere including in docs. Should also be put in a porting guide

Issue #652, opened 2024-12-04
Term list /glossary


dvdsk:
Rodio has more contributors :tada: and the maintenance team is growing stronger :partying_face:. Communication and documentation efforts are harder then they have to be because we do not have words for every concept in rodio. 

My goal is to build a glossary (specialized term list) and make that part of the documentation. We could make the docs link to that glossary each time one of its specialized terms is used.

It seems most urgent to get some words around the audio... pipelines?:
```
let thing = Decoder::new("path").amplify().pausable()
```
what do we call each of these steps and how they relate to each-other?

Lets give the steps names so we can discuss this more easily:
```
let decoder = Decoder::new();
let amplified = decoder.amplify();
let pausable = amplified.pausable();
```

This is an audio pipeline, but it could also be an audio tree (if we introduce mix). I think we need a words for 
- 'the thing that generates the samples on one end of the pipeline/tree`. Maybe it should be called the generator?
- `referring to an earlier or later step in the pipeline/tree`. Concretely is pausable, which is applied after amplify, the parent of amplify or its child? Maybe some other word works? And what is called the other way round?
- When this pipeline is consumed by something, what do we call it? *consumer*? It could also be called a *Sink* as that word will be freed up once we deprecate the `Sink` struct.

dvdsk:
I notice myself referring to _the later step in a pipeline_ as 'the source above it'. So that might work too

iluvcapra:
Something I found confusing when I was first working with the types was the use of the word "frame" in functions like `Source::current_frame_len`. I'm accustomed to a frame being a contiguous group of _simultaneous samples_, like a stereo 48k WAV file has 48000 _audio frames_ per second, and each frame has two samples. This is the terminology used in CoreAudio, the python wave library, it seems pretty common. 

rodio's sense of the word "frame" is more like what CoreAudio would call a "buffer" or ADM would call a "block", a contiguous array of frames from one read operation, or transaction, or whatever. It might help to clarify that.

PetrGlad:
Yes, I thought of renaming "frame" methods into "chunk", "run" or "span". "chunk" usually refers to some small contained piece, and in rodio's use it can be even the whole track. And "run" is too overloaded. So maybe it can be a "span".

iluvcapra:
[...]

PR #663, opened 2024-12-22
Feature/rename frame to span

wysiwys:
This PR renames frame -> span in the code and documentation, where applicable (see issue #660 and discussion in #652). 

One question that is open is about some differing terminology in the minimp3 decoder. Specifically, the mp3  `Frame`s from the minimp3 crate are interpreted as spans in the new rodio terminology. Therefore, when renaming values used by this decoder, should the term "frame" or "span" be preferred where possible? 

This code snippet from `src/decoder/mp3.rs`  illustrates this question:

```rust
pub struct Mp3Decoder<R>
where
    R: Read + Seek,
{
    // decoder: SeekDecoder<R>,
    decoder: Decoder<R>,
    // minimp3 Frames are interpreted as rodio spans
    current_span: Frame,
    current_span_offset: usize,
}
```
I went with this approach for now, because it seems quite explicit about the interpretation of `Frame`s, but I can also see the argument for matching the struct member name to its type.

Thanks

### DIFF
--- a/src/decoder/mp3.rs
+++ b/src/decoder/mp3.rs
@@ -14,7 +14,7 @@ where
 {
     // decoder: SeekDecoder<R>,
     decoder: Decoder<R>,
-    // minimp3 Frames are mapped onto rodio spans
+    // minimp3 Frames are interpreted as rodio spans
     current_span: Frame,
     current_span_offset: usize,
 }
--- a/src/decoder/mp3.rs
+++ b/src/decoder/mp3.rs
@@ -14,8 +14,8 @@ where
 {
     // decoder: SeekDecoder<R>,
     decoder: Decoder<R>,
-    // minimp3 Frames are interpreted as rodio spans
-    current_span: Frame,
+    // what minimp3 calls frames rodio calls spans
+    current_span: minimp3::Frame,
     current_span_offset: usize,
 }
 
--- a/src/decoder/mp3.rs
+++ b/src/decoder/mp3.rs
@@ -14,8 +14,9 @@ where
 {
     // decoder: SeekDecoder<R>,
     decoder: Decoder<R>,
-    current_frame: Frame,
-    current_frame_offset: usize,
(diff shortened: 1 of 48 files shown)


src/decoder/mp3.rs
```diff
@@ -14,8 +14,9 @@ where
 {
     // decoder: SeekDecoder<R>,
     decoder: Decoder<R>,
-    current_frame: Frame,
-    current_frame_offset: usize,
+    // minimp3 Frames are interpreted as rodio spans
```
dvdsk, line 17: Maybe: `// What minimp3 calls frames rodio calls spans`

and then fully qualify the path to `Frame` (something like `minimp3::Frame`)?

dvdsk (changes requested):
Thanks a lot, must have been quite the chore to do all this!

Good catch on the minim3 Frame, rodio span. I agree a comment is the best solution there. I added a suggestion on how we can improve that. Can you also add an entry in the changelog? I think the only public API change is the name of the trait method. 

Thats all for the rest it looks perfect!

wysiwys:
No problem! I have made those changes and added a changelog entry for the public API change. 

Do these updates to the changelog look good? I added one other entry about the renaming of 'frame' to 'span' in the crate and documentation. However, I can also combine both new entries into one, if that would be better.

dvdsk:
> I added one other entry about the renaming of 'frame' to 'span' in the crate and documentation.

good idea! Only thing you missed is the `-` character for starting a new item :)  But I'll fix that post merge :+1:. LGTM :tada: 

@PetrGlad @iluvcapra you might run into some merge conflicts because of this change. It should be pretty straight forward to fix.

dvdsk:
> > 
> 
> good idea! Only thing you missed is the `-` character for starting a new item :) But I'll fix that post merge 👍. LGTM 🎉

nvmd everything was perfect already. I think the github diff webui ate the `-`

PR #664, opened 2024-12-22
adds upgrade guide draft (frame->span + stream open)

dvdsk:
We will point users who upgrade to the next rodio release to this document. It will help them upgrade (and save us a lot of confused users and issues).

I do not think I have missed anything. I suggest we merge this and then every PR that introduces a breaking change updates this document. I have already done that for still open PR #663.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -6,4 +6,8 @@ This guide will help you update your code when upgrading from rodio 0.20 or earl
 ## OutputStream
 - The outputstream is now more configurable. Where you used `OutputStream::try_default()` you have a choice:
     - *(recommended)* Get an error when the default stream could not be opened: `OutputStreamBuilder::open_default_stream()?`
-    - Keep the old behavior using: `OutputStreamBuilder::open_stream_or_fallback()`, which tries to open the default (audio) stream trying others when the default stream could not be opened.
+    - Stay close to the old behavior using:
+      `OutputStreamBuilder::open_stream_or_fallback()`, which tries to open the
+      default (audio) stream. If that fails it tries all other combinations of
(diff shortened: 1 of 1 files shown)


UPGRADE.md
```diff
@@ -0,0 +1,9 @@
+This guide will help you update your code when upgrading from rodio 0.20 or earlier to the current version.
+
+## Source implementations
+- Source had a required method `current_frame_len`. In the latest version of rodio *frame* has been renamed to *span*. You will need to change every occurrence of `current_frame_len` to `current_span_len`.
+
+## OutputStream
+- The outputstream is now more configurable. Where you used `OutputStream::try_default()` you have a choice:
+    - *(recommended)* Get an error when the default stream could not be opened: `OutputStreamBuilder::open_default_stream()?`
+    - Keep the old behavior using: `OutputStreamBuilder::open_stream_or_fallback()`, which tries to open the default (audio) stream trying others when the default stream could not be opened.
```
PetrGlad, line 9: I am always forgetting to mention this, but there is also a change in the fallback logic. Previous version either probed default settings of all devices or all settings of the default device. Current implementation tries all the combinations. I hope that does not break anything.

dvdsk: I'll update the guide :+1:

dvdsk: take a look, if this is not right please let me know what you would write instead.

PetrGlad:
This will not be the last update that needs some review or changes for users.
I would add a top level heading that states the versions like "Upgrade from versions before 0.20.1 to 0.20.1"

dvdsk:
done

PetrGlad (approved):

PR #666, opened 2024-12-23
adds draft for maintainer guide

dvdsk:
Let me know what you think :)

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -1,8 +1,0 @@
-Guidelines for maintainers (those with write access) of rodio. Note these are
-guidelines not rules. Adhere to them as long as makes sense. This is a work in
-progress document, more might follow as we get to know what works. 
-
-Please feel free to open an issue and discuss these if you have a suggestion.
-
-1. Do not merge your own code to main, unless of course its a trivial change.
-   For example spelling/grammar or fixing up a PR by someone else.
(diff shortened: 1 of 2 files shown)


MAINTAINERS.md
```diff
@@ -0,0 +1,8 @@
+Guidelines for maintainers (those with write access) of rodio. Note these are
+guidelines not rules. Adhere to them as long as makes sense. This is a work in
+progress document, more might follow as we get to know what works. 
+
+Please feel free to open an issue and discuss these if you have a suggestion.
+
+1. Do not merge your own code to main, unless of course its a trivial change.
```
est31, line 7: IMO such a rule is good if there is a large enough maintainer community behind it. In smaller projects, I think it is less viable. But if there is at least two maintainers where both are well available, why not.

PetrGlad:
I would move these instructions into a contributor guide unless there is a lot of these. Also maybe we need release rules (maybe just a checklist).

dvdsk:
> I would move these instructions into a contributor guide unless there is a lot of these. Also maybe we need release rules (maybe just a checklist).

oh yeah that makes way more sense. Ill go do that

dvdsk:
I moved it to the end but I do not know if I like this. It kinda comes out of nowhere. You got any ideas?

PetrGlad:
You mean it is not clear what is the purpose of  this section?
"Instructions for maintainers", maybe?

dvdsk:
should be good to merge now

PR #667, opened 2024-12-25
Make CPAL depenrency optional

### LINKED ISSUE #349: Allow to disable cpal playback
I use Rodio in my project to open and parse audio files to check if files are broken.
Due using cpal, my app crashes on OS which doesn't have installed audio packages and shows error 
```
error while loading shared libraries: libasound.so.2: cannot open shared object file: No such file or directory
```
I don't need to play audio files, so it would be great to be able to disable cpal playback

PetrGlad:
Fixes #349.

There are many dependencies on output stream in rustdocs, `tests/*`, and `benches/*`. It would require a lot of work to make those playing nicely with the new configuration. So the new version only checks that it is compilable and adds an example that shows it is still functional. All of the unit tests that do not depend on the output stream should be the same so CI only checks compilation.
I have considered to make a no-op implementation of output stream builder just to make everyone happy without cpal, but that seems to be substantial piece of code that is rarely needed. We may add it if someone really wants it.

Note that in no-default-features configuration `tests/seek.rs` test fails to compile because it expects at least one seekable decoder to be available. So the CI adds one just to make the checks pass. The test is also too verbose to my taste but that is a unrelated to the PR.

The example in README shows how to compile the minimal version.
The sample conversion traits are now used directly from dasp_sample instead of ones re-exported by cpal.
Replacing channel count and sample rate with named types, could be a separate PR...


dvdsk (approved):
read through it all, seems good in its current state (I see your still working on it). Let me know when you want the final review.

dvdsk:
if your happy with #664 you can merge that and then add an entry for this change

PetrGlad:
I think this change does not add any incompatibilities unless no-default-features is used, if I did not miss anything. Type aliases do not introduce new types, and no APIs (should) have changed. A mention in the change log is indeed needed.

examples/into_file.rs
```diff
@@ -0,0 +1,17 @@
...
+use std::error::Error;
+use std::io::BufReader;
+
+/// Converts mp3 file to a wav file.
+/// This example does not use any audio devices
+/// and can be used in build configurations without `cpal` feature enabled.
+fn main() -> Result<(), Box<dyn Error>> {
+    let file = std::fs::File::open("assets/music.mp3")?;
+    let mut audio = rodio::Decoder::new(BufReader::new(file))?;
+
```
dvdsk, line 11: we could show off some of rodio's audio pipeline here, maybe put the file through automatic gain control?

Cargo.toml
```diff
@@ -29,6 +29,7 @@ num-rational = "0.4.2"
 default = ["flac", "vorbis", "wav", "mp3", "cpal"]
```
dvdsk, line 29: we got a ton of features, I was wondering about renaming `cpal` to `playback-cpal`. That makes clear what its for and if we add another playback backend in the future we can name that `playback-other`.

I would also be okay with plane `playback` I do not think the end user needs to know what we are using.

dvdsk (changes requested):
see comments

PetrGlad:
@dvdsk  Merging to master should not automatically trigger publish. Every time I merge something to master, CI tries to publish package, and only fails because the version already exists. I see that 0.20.1 is already on https://crates.io/crates/rodio.

dvdsk:
> Every time I merge something to master, CI tries to publish package

It only succeeds if you increment the version that way incrementing the version in Cargo.toml triggers a release. So making a release is trivial for us.

How would you have it work instead?

PetrGlad:
@dvdsk It looks like there is some confusion. Upgrade guide describes migration to 0.20.1, but it is already on crates.io, and is the latest released in the change log.

Maybe we could have some kind of "still in development" or "unreleased" format of version tags that does not trigger publishing.

PetrGlad:
@dvdsk I'd prefer to have a description of available features in README. To make the feature names consistent, maybe e.g. decoder versions should be expanded like `decoder-wav-hound` and `decoder-wav-symphonia`.  
However `ouput-cpal` is acceptable to me if you think it is a better choice.

dvdsk:
> @dvdsk It looks like there is some confusion. Upgrade guide describes migration to 0.20.1

thats a mistake :sweat_smile:. Should be from 0.20.1 or earlier. I'll change it now

> Maybe we could have some kind of "still in development" or "unreleased" format of version tags that does not trigger publishing.

If you want to set that up I go ahead :+1:

Cargo.toml
PetrGlad: @dvdsk renamed to `playback`. I'd rather add feature descriptions to `README.md`.

dvdsk (approved):

Issue #651, opened 2024-12-03
Callback on source end example


yara-blue:
Would have helped with: https://github.com/RustAudio/rodio/pull/600 (see last few messages there)

geeseofbeverlyroad:
Hi! Just came here to ask about something I assume might be similar. I am looking for a way to assign a function (function pointer or I guess more flexibly a closure) when the current Source has reached its end. For pausable / seekable sources, just "sleep until end" isn't enough, as it's useful to have the thread to be responsive to controls. 

Is there a way to do this already, or is it a planned feature?

geeseofbeverlyroad:
OK seems like this works:
```
let actual_source = /* your actual source */;
sink.append(actual_source);
let callback_source = EmptyCallback::<f32>::new(Box::new(|| println!("actual source has ended!")));
sink.append(callback_source);
```

It's quite a strange approach to me, since statistics like number of sources added to the sink's queue, are then "wrong", let's say I added 10 actual audio sources, but getting `sound_count` would return 20. 

It's possible to live with that and manage the queue separately somewhere of course. But wondering about the reasons you chose to go for this API specifically. Seems more "direct" to have the Sink itself notify of it progressing through its sources, in the "my current source changed (ended, skipped, etc), here's next source if any".

ch3pjw:
> there is, we really need an example of that (which is what this issue is for)

That would be super useful - I hit exactly the same question.

aribradshaw:
This appears to be implemented already. examples/callback_on_end.rs demonstrates appending EmptyCallback, moving shared state into the callback, and waiting for the source queue to finish. Git history shows it was added in commit ed005d4 (dds example for emptycallback closing #651) on January 2, 2025, and the example is registered in Cargo.toml. Unless there is additional behavior the example still needs to cover, this issue can likely be closed.

yara-blue:
you're right, thanks for letting us know!

PR #668, opened 2025-01-02
adds example for emptycallback closing #651

dvdsk:
Has an explicit comment instructing the user in the use of `move` since thats something people struggle with. I could have made the example nicer and more functional. It could be refactored with a function that generates the callback. Then we would have something like:
```rust
fn new_playlist_pos_callback() -> Box<impl Fn() + Send> {
    ...
}

sink.append(song);
sink.append(new_playlist_pos_callback());
sink.append(song);
sink.append(new_playlist_pos_callback());
sink.append(song);
sink.append(new_playlist_pos_callback());
```

I chose instead to keep it as short and simple as possible, even though the example code on its own is not as useful now. Functions returning a generic boxed closure are hard to understand and that distracts from what we are trying to show here (how to use the emptycallback source).

### DIFF
--- a/examples/callback_on_end.rs
+++ b/examples/callback_on_end.rs
@@ -1,0 +1,41 @@
+use std::error::Error;
+use std::io::BufReader;
+use std::sync::atomic::{AtomicU32, Ordering};
+use std::sync::Arc;
+
+fn main() -> Result<(), Box<dyn Error>> {
+    let stream_handle = rodio::OutputStreamBuilder::open_default_stream()?;
+    let sink = rodio::Sink::connect_new(&stream_handle.mixer());
+
+    let file = std::fs::File::open("assets/music.wav")?;
+    sink.append(rodio::Decoder::new(BufReader::new(file))?);
+
+    // lets increment a number after `music.wav` has played. We are going to use atomics
+    // however you could also use a `Mutex` or send a message through a `std::sync::mpsc`.
+    let playlist_pos = Arc::new(AtomicU32::new(0));
+
+    // The closure needs to own everything it uses. We move a clone of
+    // playlist_pos into the closure. That way we can still access playlist_pos
+    // after appending the EmptyCallback.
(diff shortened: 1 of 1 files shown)


examples/callback_on_end.rs
```diff
@@ -0,0 +1,41 @@
...
+    let sink = rodio::Sink::connect_new(&stream_handle.mixer());
+
+    let file = std::fs::File::open("assets/music.wav")?;
+    sink.append(rodio::Decoder::new(BufReader::new(file))?);
+
+    // lets increment a number after `music.wav` has played. We are going to use atomics
+    // however you could also use a `Mutex` or send a message through a `std::sync::mpsc`.
+    let playlist_pos = Arc::new(AtomicU32::new(0));
+
+    // The closure needs to own everything it uses. We move a clone of
```
PetrGlad, line 17: The closure has to be `move` because it is executed in the CPAL's thread. Maybe we should mention that.

PetrGlad (approved):

examples/callback_on_end.rs
dvdsk: Would that not be explaining too much of the internals of rodio. I chose 'own' since that is a term even rust beginners are familiar.

A middle ground would be mentioning the closure has to be Send and that is why it needs to own the data?

PetrGlad: Yes, it is always a problem in documentation because to decide on which details to omit you have to assume the level of the readers. I believe that the closure do not have to be Send if playback is not used (e.g. when writing to a file). Maybe adding "which is necessary for playback" would be enough.

dvdsk: > I believe that the closure do not have to be Send if playback is not used

Its part of the EmptyCallback callback type https://docs.rs/rodio/0.20.1/rodio/source/struct.EmptyCallback.html#structfield.callback

PetrGlad: OK, let's keep it this way.

Issue #364, opened 2021-03-30
Why the Decoder yields `i16`?


TonalidadeHidrica:
Currently, `rodio::decoder::Decoder` yields `i16` as an item.  Therefore, there are some loss when playing music with higher bit depth, such as 24- or 32-bit signed WAV.  This may make some audio manias sad.  Is there a specific reason `i16` was chosen?  Why not, for example, `f32` or `f64`?

PR #669, opened 2025-01-03
Decoder overhaul (16 to f32 samples + other enhancements)

roderickvd:
This PR improves audio quality in two ways:
1. Having decoders output `f32` samples directly instead of converting to `i16`
2. Properly handling high bit-depth FLAC and WAV files (20/24/32-bit) without bit depth reduction

Benefits include:
- Eliminates unnecessary quantization error from `i16` conversion
- Better matches the Rodio pipeline which already uses `f32` internally
- Preserves greater-than-`i16` precision from formats like WAV and FLAC
- More natural for decoders like Vorbis/MP3 that operate in floating point

Breaking changes:
- Decoder output type changes from `i16` to `f32`

Changes:
- Updated decoders to output `f32` samples and preserve full bit depth
- Updated tests and benchmarks to work with `f32` samples
- Simplified benchmark code by removing redundant `.to_f32s()` calls

Testing:
- Verified tests, benchmarks and examples run successfully
- Tested integration with [pleezer](https://github.com/roderickvd/pleezer/)

Fixes RustAudio/rodio#364

### DIFF
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -32,7 +32,7 @@ mod vorbis;
 mod wav;
 
 /// Output format of the decoders.
-pub type DecoderFormat = f32;
+pub type DecoderSample = f32;
 
 /// Source of audio samples from decoding a file.
 ///
@@ -71,7 +71,7 @@ where
 
 impl<R: Read + Seek> DecoderImpl<R> {
     #[inline]
-    fn next(&mut self) -> Option<DecoderFormat> {
+    fn next(&mut self) -> Option<DecoderSample> {
         match self {
             #[cfg(all(feature = "wav", not(feature = "symphonia-wav")))]
             DecoderImpl::Wav(source) => source.next(),
@@ -399,7 +399,7 @@ impl<R> Iterator for Decoder<R>
 where
     R: Read + Seek,
 {
-    type Item = DecoderFormat;
+    type Item = DecoderSample;
 
     #[inline]
     fn next(&mut self) -> Option<Self::Item> {
@@ -445,7 +445,7 @@ impl<R> Iterator for LoopedDecoder<R>
 where
     R: Read + Seek,
 {
-    type Item = DecoderFormat;
+    type Item = DecoderSample;
 
     #[inline]
     fn next(&mut self) -> Option<Self::Item> {
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -31,10 +31,10 @@ mod vorbis;
 #[cfg(all(feature = "wav", not(feature = "symphonia-wav")))]
 mod wav;
 
-#[cfg(feature = "integer-samples")]
+#[cfg(feature = "integer-decoder")]
 /// Output format of the decoders.
 pub type DecoderSample = i16;
-#[cfg(not(feature = "integer-samples"))]
+#[cfg(not(feature = "integer-decoder"))]
 /// Output format of the decoders.
 pub type DecoderSample = f32;
 
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -31,6 +31,9 @@ mod vorbis;
 #[cfg(all(feature = "wav", not(feature = "symphonia-wav")))]
 mod wav;
 
+/// Output format of the decoders.
+pub type DecoderFormat = f32;
+
 /// Source of audio samples from decoding a file.
 ///
 /// Supports MP3, WAV, Vorbis and Flac.
@@ -68,7 +71,7 @@ where
 
 impl<R: Read + Seek> DecoderImpl<R> {
     #[inline]
-    fn next(&mut self) -> Option<i16> {
+    fn next(&mut self) -> Option<DecoderFormat> {
         match self {
             #[cfg(all(feature = "wav", not(feature = "symphonia-wav")))]
             DecoderImpl::Wav(source) => source.next(),
@@ -396,10 +399,10 @@ impl<R> Iterator for Decoder<R>
 where
     R: Read + Seek,
 {
-    type Item = i16;
+    type Item = DecoderFormat;
 
     #[inline]
-    fn next(&mut self) -> Option<i16> {
+    fn next(&mut self) -> Option<Self::Item> {
         self.0.next()
     }
 
@@ -442,10 +445,10 @@ impl<R> Iterator for LoopedDecoder<R>
 where
     R: Read + Seek,
 {
-    type Item = i16;
+    type Item = DecoderFormat;
 
     #[inline]
-    fn next(&mut self) -> Option<i16> {
+    fn next(&mut self) -> Option<Self::Item> {
         if let Some(sample) = self.0.next() {
             Some(sample)
         } else {
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -31,6 +31,10 @@ mod vorbis;
 #[cfg(all(feature = "wav", not(feature = "symphonia-wav")))]
 mod wav;
 
+#[cfg(feature = "integer-samples")]
+/// Output format of the decoders.
+pub type DecoderSample = i16;
+#[cfg(not(feature = "integer-samples"))]
 /// Output format of the decoders.
 pub type DecoderSample = f32;
 
--- a/src/decoder/vorbis.rs
+++ b/src/decoder/vorbis.rs
@@ -6,6 +6,9 @@ use crate::Source;
 
 use crate::common::{ChannelCount, SampleRate};
 use lewton::inside_ogg::OggStreamReader;
+use lewton::samples::InterleavedSamples;
+
+use super::DecoderFormat;
 
 /// Decoder for an OGG file that contains Vorbis sound format.
 pub struct VorbisDecoder<R>
@@ -13,7 +16,7 @@ where
     R: Read + Seek,
 {
     stream_reader: OggStreamReader<R>,
-    current_data: Vec<i16>,
+    current_data: Vec<DecoderFormat>,
     next: usize,
 }
 
@@ -31,15 +34,18 @@ where
         Ok(Self::from_stream_reader(stream_reader))
     }
     pub fn from_stream_reader(mut stream_reader: OggStreamReader<R>) -> Self {
-        let mut data = match stream_reader.read_dec_packet_itl() {
-            Ok(Some(d)) => d,
-            _ => Vec::new(),
-        };
+        let mut data =
+            match stream_reader.read_dec_packet_generic::<InterleavedSamples<DecoderFormat>>() {
+                Ok(Some(d)) => d.samples,
+                _ => Vec::new(),
+            };
 
         // The first packet is always empty, therefore
         // we need to read the second frame to get some data
-        if let Ok(Some(mut d)) = stream_reader.read_dec_packet_itl() {
-            data.append(&mut d);
+        if let Ok(Some(mut d)) =
+            stream_reader.read_dec_packet_generic::<InterleavedSamples<DecoderFormat>>()
+        {
+            data.append(&mut d.samples);
         }
 
         VorbisDecoder {
@@ -98,22 +104,28 @@ impl<R> Iterator for VorbisDecoder<R>
 where
     R: Read + Seek,
 {
-    type Item = i16;
+    type Item = DecoderFormat;
 
     #[inline]
-    fn next(&mut self) -> Option<i16> {
+    fn next(&mut self) -> Option<Self::Item> {
         if let Some(sample) = self.current_data.get(self.next).copied() {
             self.next += 1;
             if self.current_data.is_empty() {
-                if let Ok(Some(data)) = self.stream_reader.read_dec_packet_itl() {
-                    self.current_data = data;
+                if let Ok(Some(data)) = self
+                    .stream_reader
+                    .read_dec_packet_generic::<InterleavedSamples<DecoderFormat>>()
+                {
+                    self.current_data = data.samples;
                     self.next = 0;
                 }
             }
             Some(sample)
         } else {
-            if let Ok(Some(data)) = self.stream_reader.read_dec_packet_itl() {
-                self.current_data = data;
+            if let Ok(Some(data)) = self
+                .stream_reader
+                .read_dec_packet_generic::<InterleavedSamples<DecoderFormat>>()
+            {
+                self.current_data = data.samples;
                 self.next = 0;
             }
             let sample = self.current_data.get(self.next).copied();
--- a/src/decoder/vorbis.rs
+++ b/src/decoder/vorbis.rs
@@ -54,6 +54,8 @@ where
             next: 0,
         }
     }
+
+    #[inline]
     pub fn into_inner(self) -> OggStreamReader<R> {
         self.stream_reader
     }
--- a/src/decoder/vorbis.rs
+++ b/src/decoder/vorbis.rs
@@ -8,7 +8,7 @@ use crate::common::{ChannelCount, SampleRate};
 use lewton::inside_ogg::OggStreamReader;
 use lewton::samples::InterleavedSamples;
 
-use super::DecoderFormat;
+use super::DecoderSample;
 
 /// Decoder for an OGG file that contains Vorbis sound format.
 pub struct VorbisDecoder<R>
@@ -16,7 +16,7 @@ where
     R: Read + Seek,
 {
     stream_reader: OggStreamReader<R>,
-    current_data: Vec<DecoderFormat>,
+    current_data: Vec<DecoderSample>,
     next: usize,
 }
 
@@ -35,7 +35,7 @@ where
     }
     pub fn from_stream_reader(mut stream_reader: OggStreamReader<R>) -> Self {
         let mut data =
-            match stream_reader.read_dec_packet_generic::<InterleavedSamples<DecoderFormat>>() {
+            match stream_reader.read_dec_packet_generic::<InterleavedSamples<DecoderSample>>() {
                 Ok(Some(d)) => d.samples,
                 _ => Vec::new(),
             };
@@ -43,7 +43,7 @@ where
         // The first packet is always empty, therefore
         // we need to read the second frame to get some data
         if let Ok(Some(mut d)) =
-            stream_reader.read_dec_packet_generic::<InterleavedSamples<DecoderFormat>>()
+            stream_reader.read_dec_packet_generic::<InterleavedSamples<DecoderSample>>()
         {
             data.append(&mut d.samples);
         }
@@ -104,7 +104,7 @@ impl<R> Iterator for VorbisDecoder<R>
 where
     R: Read + Seek,
 {
-    type Item = DecoderFormat;
+    type Item = DecoderSample;
 
     #[inline]
     fn next(&mut self) -> Option<Self::Item> {
@@ -113,7 +113,7 @@ where
             if self.current_data.is_empty() {
                 if let Ok(Some(data)) = self
                     .stream_reader
-                    .read_dec_packet_generic::<InterleavedSamples<DecoderFormat>>()
+                    .read_dec_packet_generic::<InterleavedSamples<DecoderSample>>()
                 {
                     self.current_data = data.samples;
                     self.next = 0;
@@ -123,7 +123,7 @@ where
         } else {
             if let Ok(Some(data)) = self
                 .stream_reader
-                .read_dec_packet_generic::<InterleavedSamples<DecoderFormat>>()
+                .read_dec_packet_generic::<InterleavedSamples<DecoderSample>>()
             {
                 self.current_data = data.samples;
                 self.next = 0;
--- a/src/decoder/vorbis.rs
+++ b/src/decoder/vorbis.rs
@@ -30,7 +30,7 @@ where
             return Err(data);
         }
 
-        let stream_reader = OggStreamReader::new(data).unwrap();
+        let stream_reader = OggStreamReader::new(data).expect("should still be vorbis");
         Ok(Self::from_stream_reader(stream_reader))
     }
     pub fn from_stream_reader(mut stream_reader: OggStreamReader<R>) -> Self {
@@ -145,13 +145,8 @@ fn is_vorbis<R>(mut data: R) -> bool
 where
     R: Read + Seek,
 {
-    let stream_pos = data.stream_position().unwrap();
-
-    if OggStreamReader::new(data.by_ref()).is_err() {
-        data.seek(SeekFrom::Start(stream_pos)).unwrap();
-        return false;
-    }
-
-    data.seek(SeekFrom::Start(stream_pos)).unwrap();
-    true
+    let stream_pos = data.stream_position().unwrap_or_default();
+    let result = OggStreamReader::new(data.by_ref()).is_ok();
+    let _ = data.seek(SeekFrom::Start(stream_pos));
+    result
 }
--- a/src/decoder/wav.rs
+++ b/src/decoder/wav.rs
@@ -86,7 +86,13 @@ where
                 (SampleFormat::Float, bits) => {
                     if bits == 32 {
                         let next_f32: Option<Result<f32, _>> = self.reader.samples().next();
-                        next_f32.and_then(|value| value.ok().map(|value| value.to_sample()))
+                        next_f32.and_then(|value| {
+                            value.ok().map(|value| {
+                                #[cfg(feature = "integer-decoder")] // perf
+                                let value = value.to_sample();
+                                value
+                            })
+                        })
                     } else {
                         // > 32 bits we cannot handle, so we'll just return equilibrium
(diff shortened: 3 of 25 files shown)


dvdsk:
@roderickvd first of all thank you for contributing to rodio :tada:. This is quite a big change before we go on we have to discuss if and how this best fits. This might take a while, that does not mean we do not appreciate all your work!

@iluvcapra @PetrGlad if you have time I would like to hear your opinions on this.

roderickvd:
Sure thing, I'll be here to address any feedback.

src/decoder/wav.rs
```diff
@@ -65,32 +69,38 @@ impl<R> Iterator for SamplesIterator<R>
...
-            }),
-            (SampleFormat::Int, 32) => self.reader.samples().next().map(|value| {
-                self.samples_read += 1;
-                i32_to_i16(value.unwrap_or(0))
-            }),
+            (SampleFormat::Float, 32) => self
+                .reader
+                .samples()
+                .next()
+                .map(|value| value.unwrap_or(0.0)),
```
PetrGlad, line 83: // I think we can use dasp_sample::Sample::EQUILIBRIUM as default.

PetrGlad:
// Down-sampling usually refers to time of spatial sampling frequency not samples' precision.
So far one of the goals of `rodio` was to allow running it in low-power (low performance) devices. I admit that just defaulting to `f32` would simplify the implementation, but `f32` is not always supported. Maybe we can have a feature switch to choose between, say, `u16` as default and `f32`. Ideally users should be able to choose any supported sample format but I do not see yet how exactly such API would look like. There is a discussion on how to avoid unnecessary resampling here #646. I believe sample format is relatively cheap, but maybe it can also be considered there.
So maybe 2 options higher-performance/higher-quality would be enough.

PetrGlad:
I now see that wav decoder can be cleaned up a bit, The format switch can be in its `new()` and there are some `unwrap`s  that can return an error instead.

src/decoder/mod.rs
```diff
@@ -31,6 +31,9 @@ mod vorbis;
 #[cfg(all(feature = "wav", not(feature = "symphonia-wav")))]
 mod wav;
 
+/// Output format of the decoders.
+pub type DecoderFormat = f32;
```
PetrGlad, line 35: "DecoderSampleFormat" or even "DecoderSample", maybe. Just "format" is ambiguous, since this may also refer to a stream format (channels layout, or sample rate). Or someone unfamiliar with the code may imagine that this is format of a decoder itself, somehow.

PetrGlad (commented):
I like that sample format is now a type alias, see also my comment to the PR.
`//` s  can be done separately or here, at your discretion...

src/decoder/wav.rs
```diff
@@ -65,32 +69,38 @@ impl<R> Iterator for SamplesIterator<R>
...
+            (SampleFormat::Int, 16) => self
+                .reader
+                .samples()
+                .next()
+                .map(|value| (value.unwrap_or(0) as i16).to_sample::<Self::Item>()),
+            (SampleFormat::Int, 24) => self
+                .reader
+                .samples()
+                .next()
+                .map(|value| (value.unwrap_or(0) << 8).to_sample::<Self::Item>()),
```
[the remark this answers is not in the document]
dvdsk: agreed 24 bit will just fit without lose of precision so that's not a problem. I see no reason to ever support more then 24 bit in rodio. (I am doubting the need for more then 16 tbh)

dvdsk:
> Having decoders output f32 samples directly instead of converting to i16

what is the upside to this? perf?

> Properly handling high bit-depth FLAC and WAV files (20/24/32-bit) without downsampling

can you expand a little on the use case for higher bit depth audio? As far as I know CD quality is fully transparent and CD uses 16 bit precision. Is the additional precision needed for effects? I can think of is software volume control using the extra precision but that needs to be paired with a higher bit output. Are those exposed by cpal?

> Better matches the Rodio pipeline which already uses f32 internally

Parts of rodio use f32 we go through great pain (all those generics :cry:) to allow for any sample type in most of the pipeline. Though I have never tested what the difference in performance is, I expect an integer based pipeline to be faster then a floating point one. We should test that.

If there is no noticeable decrease in performance this change would be very nice, it does away a ton of complex code. Writing and maintaining that code was painful however so I want us to be completely sure we do not need it before doing away with it.

roderickvd:
Thanks for your swift feedback.

@PetrGlad:

> // Down-sampling usually refers to time of spatial sampling frequency not samples' precision.

Yes, I've updated the PR and description. Will do the same when I push a new commit.

> So far one of the goals of `rodio` was to allow running it in low-power (low performance) devices. I admit that just defaulting to `f32` would simplify the implementation, but `f32` is not always supported. 

I understand that goal, and can only agree to support as much low-spec hardware as reasonable. Some thoughts:

* Rodio _already_ uses `f32` operations in many places. Simple ones like [`Sample::amplify`](https://github.com/RustAudio/rodio/blob/ea45c17db104f27645972e0f68bfa0836cf9d04e/src/conversions/sample.rs#L137) that cast then multiply as `f32`, or complex ones like [`AutomaticGainControl::process_sample`](https://github.com/RustAudio/rodio/blob/ea45c17db104f27645972e0f68bfa0836cf9d04e/src/source/agc.rs#L358) that normalize to `f32` for processing.

* `f32` is supported on all Rust targets, but a [limited set](https://github.com/search?q=repo%3Arust-lang%2Frust+%2Bsoft-float+path%3A%2F%5Ecompiler%5C%2Frustc_target%5C%2Fsrc%5C%2Fspec%5C%2Ftargets%5C%2F%2F&type=code) rely on software floating point, mostly exotic or by now quite old.

* Anecdotally, when we [switched from `i16` to `f32` in `librespot`](https://github.com/librespot-org/librespot/pull/660) some years ago, we tested on a Raspberry Pi Zero: it was fine.

> Maybe we can have a feature switch to choose between, say, `u16` as default and `f32`. Ideally users should be able to choose any supported sample format but I do not see yet how exactly such API would look like.

Here some ideas for probably a separate PR:

CamillaDSP uses a [type alias](https://github.com/HEnquist/camilladsp/blob/76aa55dddc237daa0477bfea45df8bfba6e22cc1/src/lib.rs#L55) for it. You have a choice between `f32` and `f64` by using a Cargo feature.

Or go the generic route with [`Sample`](https://docs.rs/dasp_sample/latest/dasp_sample/trait.Sample.html) everywhere.

> There is a discussion on how to avoid unnecessary resampling here #646. I believe sample format is relatively cheap, but maybe it can also be considered there. So maybe 2 options higher-performance/higher-quality would be enough.

High-quality resampling will invariably require floating point processing. I see that you're toying with Rubato which indeed [works with `f32` or `f64` samples](https://docs.rs/rubato/latest/rubato/#input-and-output-data-format).

> I now see that wav decoder can be cleaned up a bit, The format switch can be in its new() and there are some unwraps that can return an error instead.

Yes. I can clean it up if you want. So far I left it untouched, in order to leave the PR scoped as small as possible.

@dvdsk:

> what is the upside to this? perf?

I tried describing the benefits in the PR description. It's more about quality than about performance.

> can you expand a little on the use case for higher bit depth audio? As far as I know CD quality is fully transparent and CD uses 16 bit precision. Is the additional precision needed for effects? I can think of is software volume control using the extra precision but that needs to be paired with a higher bit output. Are those exposed by cpal?

I have the desire nor anything new to add to discussions about the merits of high resolution audio. I'll just say that it exists 😄 and so, how to support it in the best possible way?

Indeed, what would benefit from this is:
- digital signal processing; chaining DSPs in integer will increasingly introduce quantization errors
- software volume control; attenuation in `i16` will decrease dynamic range
- bouncing; risk of clipping when done in integer

`cpal` indeed exposes higher bit output [all the way up to `f64`](https://docs.rs/cpal/latest/cpal/trait.Sample.html#foreign-impls).

With regards output to platforms, all of which supported by `cpal`:
- Linux: ALSA supports [up to `FLOAT64`](https://docs.rs/alsa-sys/latest/alsa_sys/constant.SND_PCM_FORMAT_FLOAT64.html) but realistically will be used up to `S32` on USB.
- macOS: CoreAudio always uses [`Float32`](https://developer.apple.com/library/archive/documentation/MusicAudio/Conceptual/CoreAudioOverview/CoreAudioEssentials/CoreAudioEssentials.html#//apple_ref/doc/uid/TP40003577-CH10-SW16).
- Windows: WASAPI always converts to [`float`](https://learn.microsoft.com/en-us/windows/win32/coreaudio/device-formats) internally. 
- JACK Audio always uses [`float`](https://jackaudio.org/api/).

> If there is no noticeable decrease in performance this change would be very nice, it does away a ton of complex code. Writing and maintaining that code was painful however so I want us to be completely sure we do not need it before doing away with it.

If there is a performance difference, from experience with `librespot`, it is negligible on all but exotic and by now very old platforms (even the venerable Raspberry Pi Zero did fine).

Above, @PetrGlad hinted that there may be an interest to have a user-selectable audio pipeline. If so, you may want to keep the generics - they are not dynamically dispatched so are just as fast as settling on a single type of pipeline. Their very existence made this PR quite easy to make.

Personally, seeing from the overview audio host platforms, and DSP libraries like CamillaDSP and Rubato, I wonder if it's worth the trouble to indeed make it user-selectable between integer and floating point. When selecting `i16` would you also change all the current `f32` casts to integer arithmetic? You could, but it becomes a lot and for which platforms?

src/decoder/wav.rs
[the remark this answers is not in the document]
roderickvd: > Unless this solves particular problem, I'd rely on `to_sample` to do the right thing. Also `<< 8` is `* 256`, I'd expect it both loses precision and won't fit into f32.

You need this because `value` is actually an `i32` and `i32::to_sample()` expects its range to be `i32::MIN..=i32::MAX`. But it's not, because it's 24 bits (`-8_388_608..=8.388.607`) stored in a 32 bit container. So this operation scales the inner "`i24`" to a `i32` for `to_sample` to work correctly.

> Is it signed, `i24`? `f32::MANTISSA_DIGITS == 24` so I think it could fit as is into `f32`.

`f32` samples are normalized between `-1.0..1.0` so it's more about significant `f32::DIGITS`.

But! When it's OK to import `dasp_sample` directly then we can use the `dasp_sample::I24` type and forget about it 😄

src/decoder/wav.rs
roderickvd: Will do. Note that there are more places in Rodio that could use this, in particular all the implementations of [`Sample::zero_value`](https://github.com/RustAudio/rodio/blob/ea45c17db104f27645972e0f68bfa0836cf9d04e/src/conversions/sample.rs#L93). I can add that with this PR.

dvdsk:
@roderickvd first of all thank you very much for that very clear and thorough response! 

I got a pi zero and a pinephone I want to check performance on, those both have fpu's but those might be slower then integer operations (I doubt it matters here). I'll try if I can check that tomorrow, cant promise anything unfortunately.

@PetrGlad If we adopt this we should seriously consider making `Source`'s item f32. That would dramatically simplify mix, uniformsource, the new in progress resampler trait and probably more. To me that is the biggest appeal of this change. 

@est31 if you got time, whats your opinion on such a change?

src/decoder/wav.rs
```diff
@@ -5,8 +5,12 @@ use crate::source::SeekError;
 use crate::Source;
 
 use crate::common::{ChannelCount, SampleRate};
+
+use cpal::Sample;
```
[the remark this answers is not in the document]
roderickvd: Done in [a1cf45a](https://github.com/RustAudio/rodio/pull/669/commits/a1cf45a02c723043523a017a55386779023e2c70).

src/decoder/wav.rs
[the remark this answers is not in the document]
roderickvd: Done in [a1cf45a](https://github.com/RustAudio/rodio/pull/669/commits/a1cf45a02c723043523a017a55386779023e2c70).

src/decoder/mod.rs
roderickvd: Done in [a1cf45a](https://github.com/RustAudio/rodio/pull/669/commits/a1cf45a02c723043523a017a55386779023e2c70).

src/decoder/wav.rs
roderickvd: Mostly done in [a1cf45a](https://github.com/RustAudio/rodio/pull/669/commits/a1cf45a02c723043523a017a55386779023e2c70). I am sure there is more to clean up, like `Sample::saturating_add`, but I need to get to bed :wink: so I'll get to that later.

roderickvd:
[a1cf45a](https://github.com/RustAudio/rodio/pull/669/commits/a1cf45a02c723043523a017a55386779023e2c70) addresses most of the review comments.

I see in that in all decoders, there's a lot of unwrapping going on, as well as `is_wav`, `is_flac`, etc. checks. Why are those necessary? The inner decoder should fail opening them otherwise anyway.

I'll have another stab at some refactoring of e.g. `Sample::saturating_add` that we can now depend on `dasp_sample` for, but wanted to share this progress with you regardless.

roderickvd:
> @roderickvd first of all thank you very much for that very clear and thorough response!
> 
> I got a pi zero and a pinephone I want to check performance on, those both have fpu's but those might be slower then integer operations (I doubt it matters here). I'll try if I can check that tomorrow, cant promise anything unfortunately.

Great, let me know!
 
> @PetrGlad If we adopt this we should seriously consider making `Source`'s item f32. That would dramatically simplify mix, uniformsource, the new in progress resampler trait and probably more. To me that is the biggest appeal of this change.

Isn't it already generic over `u16`, `i16` and `f32` by the `Sample` trait bound?

dvdsk:
> Isn't it already generic over `u16`, `i16` and `f32` by the `Sample` trait bound?

yes however writing interfaces that cover both generic code and non-generic is rather ugly. See (wip): https://github.com/RustAudio/rodio/blob/resamplers/src/conversions/sample_rate.rs & first bullet point here: https://github.com/RustAudio/rodio/pull/670#issuecomment-2571269740. I needed to combine sample type conversion and sample_rate conversion since rubato only returns f32. It works perfectly however the generics are quite brittle and the compiler messages not that helpful. Brittle code makes refactoring hard which in turn leads to bad maintainability.

roderickvd:
https://github.com/RustAudio/rodio/pull/669/commits/01dc93931dfad7889c2c3cbb3324b819a0d9a47a adds a `integer-samples` feature that switches the decoder output from `f32` to `i16`.

On my Apple M3 - so: fast and with hard float - I timed the benchmark suite. The following values are averaged over three runs, after having performed two "warm-up" runs for cache locality whose timings were ignored.

Nothing scientific and WAV only:

```
# baseline: i16 default decoders
# time cargo bench --features integer-samples
28.52 real        28.25 user         0.24 sys

# i16 Symphonia
# time cargo bench --features "symphonia-all,integer-samples"
25.91 real        25.62 user         0.24 sys

# f32 default decoders
# time cargo bench
29.33 real        29.04 user         0.24 sys

# f32 Symphonia
# time cargo bench --features symphonia-all
25.83 real        25.56 user         0.25 sys
```

PetrGlad:
This is somewhat unrelated, but dasp_sample is more flexible than Rodio's since it implements Sample trait for plain scalar values. I wanted to use Duplex<f32> trait from it, but such a requirement should then be propagated to other sources/filters that are used in Source. We do not even have `Debug` for `Sample`s, which is rather cumbersome while debugging.

PetrGlad:
@roderickvd No, there is no need to do extra work here, I just try to keep the ideas somewhere.

dvdsk:
keep calm :) Love the enthusiasm however should we not first decide if we are moving to ungeneric f32? I'd hate to have to remove hard work. 

I wanted to get some benchmarks today but Pop-os multiarch repo's are broken and podman wont work on zfs it seems..... Tomorrow ill spin up a VM to crosscompile in, bit of a hassle :sweat:. Soon as I got that running I'll have some benchmark results.

roderickvd:
I have [none of those systems](https://github.com/search?q=repo%3Arust-lang%2Frust+%2Bsoft-float+path%3A%2F%5Ecompiler%5C%2Frustc_target%5C%2Fsrc%5C%2Fspec%5C%2Ftargets%5C%2F%2F&type=code). Even the original Raspberry Pi 1 Model A from 2012 has hardware floating point support.

dvdsk:
There is a rodio fork that specializes for embedded systems (so not cpal). I can not seem to find it unfortunately. Rodio is currently not suitable for embedded due to us allocating memory here and there. Its not one of the goals for us, since those are the only targets without hardware float we can probably ignore that category. 

I just want to see what weaker hardware float systems do with an only f32 pipeline. I can imagine it might mess up cache locality or the float ops take more power/cycles. Maybe tomorrow we will know.

roderickvd:
You mean [awedio](https://github.com/10buttons/awedio)?

dvdsk:
> You mean [awedio](https://github.com/10buttons/awedio)?

Yes! :tada: Its a great crate and has some features we are still implementing/discussing.

roderickvd:
I feel like I'm getting to a state where I'd be happy with. Things to do:

- [ ] Cleaning up the individual decoders like pointed out in the earlier review comments - FLAC, WAV and (mini) MP3 done, the test to go. 
- [ ] Renaming the `integer-samples` feature to `integer-decoder` - after all there is `f32` processing in the pipeline regardless
- [ ] Some micro-optimizations to bypass `to_sample` depending on the feature gate

Just a hobby 😁 don't worry to instruct me to take things out.

roderickvd:
OK - I think I've got it as far as I wanted to. Ready for further review.

src/decoder/vorbis.rs
```diff
@@ -133,13 +147,8 @@ fn is_vorbis<R>(mut data: R) -> bool
...
-    if OggStreamReader::new(data.by_ref()).is_err() {
-        data.seek(SeekFrom::Start(stream_pos)).unwrap();
-        return false;
-    }
-
-    data.seek(SeekFrom::Start(stream_pos)).unwrap();
-    true
+    let stream_pos = data.stream_position().unwrap_or_default();
+    let result = OggStreamReader::new(data.by_ref()).is_ok();
+    let _ = data.seek(SeekFrom::Start(stream_pos));
```
PetrGlad, line 152: Why are such errors ignored now?

PetrGlad (approved):

PetrGlad:
@dvdsk I am inclined to accept this, users who need integer samples can still have them. Or did I miss something?

src/decoder/vorbis.rs
roderickvd: The intent is to decrease the probability of panicking. 

When the stream position could not be gotten or restored, then old behavior was to panic.

New behavior is to still probe whether a demuxer/decoder can be initialized (whether we "are Vorbis, MP3, ...") and continue. This may still trigger the old behavior, because if the stream cannot be rewound, then in the `new` method that calls this `is_vorbis` it is likely that re-initializing the decoder will fail on this advance position.

dvdsk:
> @dvdsk I am inclined to accept this, users who need integer samples can still have them. Or did I miss something?

I'm okay with most of this, I just want to have a perf check done but I have been in crosscompile &cpal bug hell :sob: . But I might be able to simplify is by using the new rodio without cpal feature you introduced. Results will be a bit less reliable as we will no longer measuring the full pipeline including playback but it should still tell us enough.

I expect nothing bad to come out of the check but I would like to have it done before merging this. I think I'll be able to get the perf check done tomorrow. That is if my test targets do not magically explode.

And if the check comes out alright we should probably change the whole pipeline to f32's, but that is a discussion for a later PR.

dvdsk:
im gonna review this pr as is now and then it can be merged

roderickvd:
Thank you for putting together those tests 🎉 happy to address or answer any points you have.

I have some further improvements in the works to improve functionality & seek performance of the Symphonia decoder, which I can rebase on this one and submit afterwards.

src/decoder/wav.rs
```diff
@@ -65,36 +76,70 @@ impl<R> Iterator for SamplesIterator<R>
...
+                    if bits == 32 {
+                        let next_f32: Option<Result<f32, _>> = self.reader.samples().next();
+                        next_f32.and_then(|value| {
+                            let value = value.ok();
+                            #[cfg(feature = "integer-decoder")] // perf
+                            let value = value.map(|value| value.to_sample());
+                            value
+                        })
+                    } else {
+                        // > 32 bits we cannot handle, so we'll just return equilibrium
```
dvdsk, line 97: do we want to log or print here? Returning equilibrium is a silent error and will be very hard to debug. I would be okay with panicking too.

src/decoder/wav.rs
```diff
@@ -65,36 +76,70 @@ impl<R> Iterator for SamplesIterator<R>
...
+                    // Unofficial WAV integer bit depth, try to handle it anyway
+                    let next_i32: Option<Result<i32, _>> = self.reader.samples().next();
+                    next_i32.and_then(|value| {
+                        value.ok().map(|value| {
+                            if bits <= 32 {
+                                (value << (32 - bits)).to_sample()
+                            } else {
+                                // > 32 bits we cannot handle, so we'll just return
+                                // equilibrium and let the iterator continue
+                                Self::Item::EQUILIBRIUM
```
dvdsk, line 136: same as above

dvdsk (changes requested):
Overall very nice, like the refactoring here and there, good changes. Just one issue for me in WAV decoding. 

This pr switches the behavior on error from panic to returning silence. In my opinion that is a mistake. While I agree that panicking is sub-optimal not giving any indication that something went wrong is worse in my opinion. Maybe logging can be an alternative?

dvdsk:
> which I can rebase on this one and submit afterwards.

good idea, lets get this landed :+1:

roderickvd:
> Overall very nice, like the refactoring here and there, good changes. Just one issue for me in WAV decoding.
> 
> This pr switches the behavior on error from panic to returning silence. In my opinion that is a mistake. While I agree that panicking is sub-optimal not giving any indication that something went wrong is worse in my opinion. Maybe logging can be an alternative?

That is a good point. 6498406 should do it.
It also contains a final touch to more robustly compare floats for equality.

src/decoder/wav.rs
```diff
@@ -94,9 +94,11 @@ where
                             value
                         })
                     } else {
-                        // > 32 bits we cannot handle, so we'll just return equilibrium
-                        // and let the iterator continue
-                        Some(Self::Item::EQUILIBRIUM)
+                        #[cfg(feature = "tracing")]
+                        tracing::error!("Unsupported WAV float bit depth: {}", bits);
+                        #[cfg(not(feature = "tracing"))]
+                        println!("Unsupported WAV float bit depth: {}", bits);
```
dvdsk, line 100: we can do eprintln then it will print to stderr

dvdsk:
the tests are failing due to a cargo clippy warning in unrelated code. Lets ignore that, ill fix the issue on main.

src/decoder/wav.rs
PetrGlad: If we're settled on `tracing`, I would use only that. A simple `tracing` listener can print to stdout or stderr if necessary.
We can have an example on how to do that.

dvdsk: I do not expect most users of rodio to know or use tracing (rodio is used by a lot of rust novices it seems). Since `tracing::error!` calls do nothing without tracing set up in the final binary cut out the (e)println would mean silent errors a lot of users.

We do want the tracing support since that provides a way of redirecting such errors (essential in a TUI app)

roderickvd: What do you guys prefer? Let me know and I can wrap it up.

dvdsk: In my opinion its ready as is, this behavior matches what we do when we run into cpal stream errors. If we refine that later we can take a look at this again. But that is a separate concern for a separate PR/Issue. @PetrGlad would you agree?

roderickvd: Except I should probably still change to `eprintln`?

dvdsk: oh yes that would be nice

PetrGlad: Yes we can keep the error reporting as is for now, since this is the existing pattern. 

I do not think users should know how to use tracing. They only need to know how to enable debug output from rodio. I will have an example for this.

roderickvd: > oh yes that would be nice

Changed to `eprintln` in [9861056](https://github.com/RustAudio/rodio/pull/669/commits/9861056120d6b7ecc9826585d6a040802106a9fe).

roderickvd:
> the tests are failing due to a cargo clippy warning in unrelated code. Lets ignore that, ill fix the issue on main.
> 
> (fix pushed to master)

Rebased on it.

dvdsk (approved):

dvdsk:
lets get this merged :tada:, thank you for your valuable input, hard work and patience Roderick!

roderickvd:
👍  on to the next!

PR #675, opened 2025-01-10
Adds split_once to Source

dvdsk:
split_once can be used chop a source into two sources without duplicating the first. Those sources can then not be played simultaneously. This is usefull for adding a fadeout at the end of a song for example. Though we could later specialize fadeout at end.

The implementation uses a `SplitAt` struct, `split_once` returns an array of two of those. In the future we can introduce methods like `split` that return a vec of SplitAt's allowing users to chop a source into multiple sections

### DIFF
--- a/examples/fadeout_end.rs
+++ b/examples/fadeout_end.rs
@@ -1,0 +1,25 @@
+use std::error::Error;
+use std::io::BufReader;
+use std::time::Duration;
+
+use rodio::Source;
+
+fn main() -> Result<(), Box<dyn Error>> {
+    let stream_handle = rodio::OutputStreamBuilder::open_default_stream()?;
+    let sink = rodio::Sink::connect_new(&stream_handle.mixer());
+
+    let file = std::fs::File::open("assets/music.wav")?;
+    let music = rodio::Decoder::new(BufReader::new(file))?;
+    let [start, end] = music.split_once(Duration::from_secs(3));
+    let end_duration = end
+        .total_duration()
+        .expect("can only fade out at the end if we know the total duration");
+    let end = end.fade_out(end_duration);
+
+    sink.append(start);
+    sink.append(end);
+
+    sink.sleep_until_end();
+
+    Ok(())
+}
--- a/examples/fadeout_end.rs
+++ b/examples/fadeout_end.rs
@@ -1,16 +1,16 @@
 use std::error::Error;
-use std::io::BufReader;
 use std::time::Duration;
 
 use rodio::Source;
 
 fn main() -> Result<(), Box<dyn Error>> {
     let stream_handle = rodio::OutputStreamBuilder::open_default_stream()?;
-    let sink = rodio::Sink::connect_new(&stream_handle.mixer());
-
+    let sink = rodio::Sink::connect_new(stream_handle.mixer());
+    //
     let file = std::fs::File::open("assets/music.wav")?;
-    let music = rodio::Decoder::new(BufReader::new(file))?;
-    let [start, end] = music.split_once(Duration::from_secs(3));
+    let source = rodio::Decoder::try_from(file)?;
+
+    let [start, end] = source.split_once(Duration::from_secs(3));
     let end_duration = end
         .total_duration()
         .expect("can only fade out at the end if we know the total duration");
--- a/src/source/mod.rs
+++ b/src/source/mod.rs
@@ -7,6 +7,7 @@ use crate::common::{ChannelCount, SampleRate};
 use crate::Sample;
 use amplify::to_linear;
 use dasp_sample::FromSample;
+use split_at::SplitAt;
 
 pub use self::agc::AutomaticGainControl;
 pub use self::amplify::Amplify;
@@ -73,6 +74,7 @@ mod skip;
 mod skippable;
 mod spatial;
 mod speed;
+mod split_at;
 mod square;
 mod stoppable;
 mod take;
@@ -518,6 +520,15 @@ pub trait Source: Iterator<Item = Sample> {
         skippable::skippable(self)
     }
 
+    /// returns two sources, the second source is inactive and will return
+    /// `None` until the first has passed the split_point.
+    fn split_once(self, at: Duration) -> [SplitAt<Self>; 2]
+    where
+        Self: Sized,
+    {
+        SplitAt::new(self, at)
+    }
+
     /// Start tracking the elapsed duration since the start of the underlying
     /// source.
     ///
@@ -635,6 +646,8 @@ pub enum SeekError {
     // own `try_seek` implementations.
     /// Any other error probably in a custom Source
     Other(Box<dyn std::error::Error + Send>),
+    /// Can not seek, this part of the split is not active
+    SplitNotActive,
 }
 impl fmt::Display for SeekError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
@@ -651,6 +664,9 @@ impl fmt::Display for SeekError {
             #[cfg(feature = "wav")]
             SeekError::HoundDecoder(err) => write!(f, "Error seeking in wav source: {}", err),
             SeekError::Other(_) => write!(f, "An error occurred"),
+            SeekError::SplitNotActive => {
+                write!(f, "Can not seek, this part of the split is still active")
+            }
         }
     }
 }
@@ -663,6 +679,7 @@ impl std::error::Error for SeekError {
             #[cfg(feature = "wav")]
             SeekError::HoundDecoder(err) => Some(err),
             SeekError::Other(err) => Some(err.as_ref()),
+            SeekError::SplitNotActive => None,
         }
     }
 }
@@ -685,6 +702,7 @@ impl SeekError {
             #[cfg(feature = "wav")]
             SeekError::HoundDecoder(_) => false,
             SeekError::Other(_) => false,
+            SeekError::SplitNotActive => true,
         }
     }
 }
--- a/src/source/mod.rs
+++ b/src/source/mod.rs
@@ -520,8 +520,15 @@ pub trait Source: Iterator<Item = Sample> {
         skippable::skippable(self)
     }
 
-    /// returns two sources, the second source is inactive and will return
-    /// `None` until the first has passed the split_point.
+    /// returns two new sources that are continues segments of `self`.
+    /// The second segment is inactive and will return None until the
+    /// first segment has passed the provided split_point.
+    ///
+    /// # Seeking
+    /// If you seek outside the range of a segment the segment will
+    /// deactivate itself such that the other segment can play. This
+    /// works well when searching forward. Seeking back can require you
+    /// to play the previous segment again.
     fn split_once(self, at: Duration) -> [SplitAt<Self>; 2]
     where
         Self: Sized,
--- a/src/source/mod.rs
+++ b/src/source/mod.rs
@@ -7,7 +7,7 @@ use crate::common::{ChannelCount, SampleRate};
 use crate::Sample;
 use amplify::to_linear;
 use dasp_sample::FromSample;
-use split_at::SplitAt;
+use split_at::Segment;
 
 pub use self::agc::AutomaticGainControl;
 pub use self::amplify::Amplify;
@@ -529,11 +529,11 @@ pub trait Source: Iterator<Item = Sample> {
     /// deactivate itself such that the other segment can play. This
     /// works well when searching forward. Seeking back can require you
     /// to play the previous segment again.
-    fn split_once(self, at: Duration) -> [SplitAt<Self>; 2]
+    fn split_once(self, at: Duration) -> [Segment<Self>; 2]
     where
         Self: Sized,
     {
-        SplitAt::new(self, at)
+        Segment::new(self, at)
     }
 
     /// Start tracking the elapsed duration since the start of the underlying
@@ -654,7 +654,7 @@ pub enum SeekError {
     /// Any other error probably in a custom Source
     Other(Box<dyn std::error::Error + Send>),
     /// Can not seek, this part of the split is not active
-    SplitNotActive,
+    SegmentNotActive,
 }
 impl fmt::Display for SeekError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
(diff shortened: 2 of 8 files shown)


src/source/split_at.rs
```diff
@@ -0,0 +1,123 @@
...
+        let first_span_channel_count = input.channels();
+        let total_duration = input.total_duration();
+        [
+            Self {
+                shared_source: shared_source.clone(),
+                active: Some(input.track_position()),
+                split_duration: Some(split_point),
+                split_point,
+                first_span_sample_rate,
+                first_span_channel_count,
```
PetrGlad, line 39: Maybe the second source can be of different type that just awaits until the values are available? These fields are only useful there.
Maybe we should introduce some kind of `UniformSource` that does not have spans, or bite the bullet and document that in some cases we only expect `current_span_len() == None`. The latter is messier but can be done with existing API.

src/source/split_at.rs
```diff
@@ -0,0 +1,123 @@
...
+            self.active = shared;
+            self.active.as_mut()?
+        };
+
+        // There is some optimization potential here we are not using currently.
+        // Calling get_pos once per span should be enough
+        if input.get_pos() < self.split_point {
+            input.next()
+        } else {
+            let source = self.active.take();
```
PetrGlad, line 79: I would skip this if  `self.active.is_none()`. If someone would try to call next second time after getting None this may reset the shared source to None before second source sees it. Yes, we documented that completed iterators should not be tried, but this may let someone down...

PetrGlad (approved):
Looks good. I do not like that `first_span_` stuff but this can be cleaned up later.

Alternatively, if the motivation is to support for fade-out, we can modify fade-out to start at some later moment (it would pass-through until it's time to ramp down).  Maybe we can even have arbitrary "automation" lane, with piece-wise gain function (can be even smoothed with splines :)

src/source/split_at.rs
```diff
@@ -0,0 +1,123 @@
...
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let input = if let Some(active) = self.active.as_mut() {
+            active
+        } else {
+            // did they other stop?
+            let shared = self
+                .shared_source
+                .lock()
+                .expect("audio thread should not panic")
```
PetrGlad, line 68: For expect is recommended to state what is expected normally, I would replace this with "lock shared source in second fragment". This would give better diagnostics should this happen.

src/source/split_at.rs
```diff
@@ -0,0 +1,123 @@
...
+        // There is some optimization potential here we are not using currently.
+        // Calling get_pos once per span should be enough
+        if input.get_pos() < self.split_point {
+            input.next()
+        } else {
+            let source = self.active.take();
+            *self
+                .shared_source
+                .lock()
+                .expect("audio thread should not panic") = source;
```
PetrGlad, line 83: Similarly this could be "lock shared source at the end of first fragment". It could be shorter, but it is better to distinguish this from the expect message in `next`.

PetrGlad (changes requested):

src/source/mod.rs
```diff
@@ -498,6 +500,15 @@ where
         skippable::skippable(self)
     }
 
+    /// returns two sources, the second source is inactive and will return
+    /// `None` until the first has passed the split_point.
+    fn split_once(self, at: Duration) -> [SplitAt<Self>; 2]
```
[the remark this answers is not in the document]
dvdsk: named after: https://doc.rust-lang.org/std/primitive.str.html#method.split_once, we might later want a split method that takes a vec of durations and returns a vec of `SplitAt`s

We can do `split_once_at`?

src/source/mod.rs
```diff
@@ -622,6 +635,9 @@ impl fmt::Display for SeekError {
             #[cfg(feature = "wav")]
             SeekError::HoundDecoder(err) => write!(f, "Error seeking in wav source: {}", err),
             SeekError::Other(_) => write!(f, "An error occurred"),
+            SeekError::SplitNotActive => {
```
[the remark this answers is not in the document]
dvdsk: By using the word split I hope to have users associate the error with the `split_once` function and `SplitAt` struct. Fragment is another term so that association is gone. 

What is the problem with `SplitNotActive`? I agree its not a very clear term.

src/source/split_at.rs
dvdsk: > Maybe the second source can be of different type that just awaits until the values are available? These fields are only useful there.

I thought about that however this way we can reuse `SplitAt` for a future function that splits into more then two splits.

> Maybe we should introduce some kind of UniformSource that does not have spans, or bite the bullet and document that in some cases we only expect current_span_len() == None. The latter is messier but can be done with existing API.

`current_span_len().is_none()` already has the meaning that the span will never change. At least thats what the trait doc says I think....

src/source/split_at.rs
dvdsk: good catch! ill do that tomorrow!

src/source/split_at.rs
dvdsk: The docs disagree: https://doc.rust-lang.org/std/option/enum.Option.html#recommended-message-style & https://doc.rust-lang.org/std/result/enum.Result.html#recommended-message-style

dvdsk: oh wait, you mean the message should be: "audio thread does not panic"?

src/source/split_at.rs
dvdsk: Its a blocking lock so it can not fail because its already locked, since then it just blocks..

I find: `The audio thread cant panic while first fragment is storing source in lock` clearer, though pretty long.

dvdsk:
> Alternatively, if the motivation is to support for fade-out, we can modify fade-out to start at some later moment (it would pass-through until it's time to ramp down).

That was the one of the motivations for creating this. It seemed valuable to have a method of splitting a source into two "mutable" sections for other effects. I thought about a fadeout_end source but it would only be marginally more efficient. It is easier to use/find but I think Sink/ the new Player should help out with that. The Player should get a fadeout_end setting.

dvdsk:
> Maybe we can even have arbitrary "automation" lane, with piece-wise gain function (can be even smoothed with splines :)

This sounds interesting though I do not fully get how it should work yet.

dvdsk:
ill fixup your suggestions later, gotta get to bed now :moon:

src/source/mod.rs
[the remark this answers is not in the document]
PetrGlad: `split_once` in std uses _content_ of the sequence as delimiter. So, number of pieces in the result is unpredictable, the delimiter can occur more than once. Therefore, in that case it is necessary to distinguish the method that uses only first delimiter from one that returns all the fragments. In this case of the `Source` splitter, the delimiter is time which is unique.

src/source/split_at.rs
PetrGlad: My primary complaint here is that if this panic happens, the message will not be useful. I'd want to have something that helps to identify where this happened (in release mode stack may not be helpful).

src/source/split_at.rs
PetrGlad: Yes, it is. Maybe you'd have a better idea. I just want the message to be more informative.

src/source/mod.rs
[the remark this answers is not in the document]
PetrGlad: Yes, maybe. I thought that "fragment" maybe useful in other places where partial source fragments will be used...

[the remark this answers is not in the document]
PetrGlad: English is not my native language but to me it feels like "split" normally refers to the location or cause of divide, not the parts it separates.

src/source/mod.rs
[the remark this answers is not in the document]
dvdsk: In this case returning the results on the stack (via the array) allows unpacking & increasing performance, for example: 
```rust
let [start, end] = some_source.split_once(..);
```
is easier and safer (hidden panic in [0] & [1] that can not be checked compile time) then:
```rust
let split = some_source.split(..);
let start = split[0];
let end = split[1];
```

src/source/split_at.rs
dvdsk: I got it: we will make the `current_span_len` report `Some(1)` while the segment is inactive. That should trigger a re-query of sample-rate and channel count every time. 

`Some(0)` is documented to mean no more data so that would not work.

src/source/mod.rs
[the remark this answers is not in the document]
dvdsk: I've renamed it to Segment.

dvdsk:
ready for review again

yara-blue:
I forgot about this, would be nice to have still.

examples/fadeout_end.rs
```diff
@@ -0,0 +1,25 @@
...
+use std::time::Duration;
+
+use rodio::Source;
+
+fn main() -> Result<(), Box<dyn Error>> {
+    let stream_handle = rodio::OutputStreamBuilder::open_default_stream()?;
+    let sink = rodio::Sink::connect_new(&stream_handle.mixer());
+
+    let file = std::fs::File::open("assets/music.wav")?;
+    let music = rodio::Decoder::new(BufReader::new(file))?;
```
roderickvd: Update to use `try_new`.

src/source/mod.rs
```diff
@@ -651,6 +671,9 @@ impl fmt::Display for SeekError {
             #[cfg(feature = "wav")]
             SeekError::HoundDecoder(err) => write!(f, "Error seeking in wav source: {}", err),
             SeekError::Other(_) => write!(f, "An error occurred"),
+            SeekError::SegmentNotActive => {
+                write!(f, "Can not seek, this segement is still active")
```
roderickvd, line 675: Typo: segement

src/source/split_at.rs
```diff
@@ -0,0 +1,137 @@
...
+    S: Source,
+    S::Item: crate::Sample,
+{
+    type Item = <S as Iterator>::Item;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let input = if let Some(active) = self.active.as_mut() {
+            active
+        } else {
+            // did they other stop and is it in our segment?
```
roderickvd: Typo: they?

src/source/split_at.rs
```diff
@@ -0,0 +1,137 @@
...
+    S::Item: crate::Sample,
+{
+    fn current_span_len(&self) -> Option<usize> {
+        if let Some(input) = self.active.as_ref() {
+            input.current_span_len()
+        } else {
+            // We do not know the channel count nor sample rate if the source
+            // is inactive. We will provide dummy values. This ensures the
+            // caller will recheck when we become active
+            Some(1)
```
roderickvd, line 92: Hmm. Seems kind of hackish. I understand why it would work from the comments, but that seems more like a side-effect? (spans booo)

src/source/split_at.rs
```diff
@@ -0,0 +1,137 @@
...
+        }
+    }
+
+    fn channels(&self) -> ChannelCount {
+        self.active
+            .as_ref()
+            .map(Source::channels)
+            .unwrap_or_default()
```
roderickvd: Won't a ChannelCount or SampleRate of 0 cause panics later?

tests/split.rs
```diff
@@ -0,0 +1,25 @@
...
+    let source = SamplesBuffer::new(1, 1, input);
+
+    let [start, end] = source.split_once(Duration::from_secs(3));
+
+    let played: Vec<_> = start.chain(end).collect();
+    assert_eq!(input.as_slice(), played.as_slice());
+}
+
+#[test]
+fn seek_over_segment_boundry() {
```
roderickvd: Typo: boundry

src/source/split_at.rs
```diff
@@ -0,0 +1,137 @@
...
+}
+
+impl<S> Segment<S>
+where
+    S: Source,
+    <S as Iterator>::Item: crate::Sample,
+{
+    /// see docs at [Source::split_once];
+    pub(crate) fn new(input: S, split_point: Duration) -> [Self; 2] {
+        let shared_source = Arc::new(Mutex::new(None));
```
roderickvd, line 22: Why keep a shared source, instead of truly returning two separate sources?

tests/split.rs
```diff
@@ -0,0 +1,25 @@
...
+    let input = [0, 1, 2, 3, 4, 5, 6, 7].map(|s| s as f32);
+    let source = SamplesBuffer::new(1, 1, input);
+
+    let [mut start, mut end] = source.split_once(Duration::from_secs(3));
+    assert_eq!(start.next(), Some(0.0));
```
roderickvd, line 21: I suggest the documentation and this example to be clearer and more exhaustive. As I understand, effectively you get two sources that each have a range of `0..total_duration` independent from the original `total_duration` unless you merge them again. So you cut into two "clips" - that's pretty clear.

Then seeking is seeking within that clip, and no longer related to the original source.

roderickvd (commented):
Nice idea but some questions marks lingering.

src/source/split_at.rs
yara-blue: I'm open to suggestions, this was the best I could come up with.

src/source/split_at.rs
yara-blue: yes! good point. I think there are a few places we used to treat zero as placeholder/unknown. I'll add a reasonable default here now (1 channel, sample_rate 44100). Its only for one frame.

The only "perfect" solution I can think off is seeking to the split then back on source creation. Though that introduces a lot of latency starting the source, and requires back seeking to be supported which is not always the case....

Honestly the whole span system just fundamentally breaks this and many other parts of Rodio. We keep hacking around it but maybe we should stop that. There is a strong argument for redoing the entire audio engine and not adding more stuff until that is done. At that point we should wait for the audio ecosystem stuff. 

Maybe I should concentrate my time on that instead....

src/source/split_at.rs
yara-blue: There is no way to clone each source, decoder for Example is not clone.

tests/split.rs
yara-blue: agreed

yara-blue:
Okay so looking at it again, this is just not sound. The channel count after the split point can be queried before we are there. This needs span less rodio to work. Closing

Issue #680, opened 2025-01-14
ChannelVolume usage causes crackling/distortion


will3942:
We are using ChannelVolume to output Sources in mono on multichannel devices.

We have noticed that when using ChannelVolume the audio is distorted and crackly versus when not using ChannelVolume.

Code sample for good audio:

```
use std::io::BufReader;

use rodio::source::ChannelVolume;

fn main() {
    let (_stream, handle) = rodio::OutputStream::try_default().unwrap();
    let sink = rodio::Sink::try_new(&handle).unwrap();

    let file = std::fs::File::open("assets/sample.mp3").unwrap();
    let decoder = rodio::Decoder::new(BufReader::new(file)).unwrap();

    sink.append(decoder);

    sink.sleep_until_end();
}
```

Code sample for distortion:

```
use std::io::BufReader;

use rodio::source::ChannelVolume;

fn main() {
    let (_stream, handle) = rodio::OutputStream::try_default().unwrap();
    let sink = rodio::Sink::try_new(&handle).unwrap();

    let file = std::fs::File::open("assets/sample.mp3").unwrap();
    let decoder = rodio::Decoder::new(BufReader::new(file)).unwrap();

    let channel_volumes = ChannelVolume::new(decoder, [1.0, 1.0].to_vec());

    sink.append(channel_volumes);

    sink.sleep_until_end();
}
```

Sample mp3: https://github.com/will3942/test-rodio/raw/refs/heads/main/assets/sample.mp3

will3942:
This occurs in both release / non-release mode (although more evident in release).

Particularly evident during 13 to 20s section.

dvdsk:
can you try out PR https://github.com/RustAudio/rodio/pull/669? That might fix it.

will3942:
> can you try out PR #669? That might fix it.

As far as my ears can hear it that seems to fix it or at least come very close to fixing it.

will3942:
@dvdsk Thanks for the input - what's your thinking on the timelines for the next release of rodio? We're planning on moving to production next week with our app and so will be running off a Rodio fork with a few of these changes (ChannelVolume temporary fix or ChannelRouter and this f32 change).

Happy to understand how we can support at all on the development - we're really interested in the changes to the resampler (e.g. Rubato) and ChannelRouter.

will3942:
[...]

PR #681, opened 2025-01-17
Improve ChannelVolume's Mono conversion algorithm to use the average of samples

will3942:
Fixes https://github.com/RustAudio/rodio/issues/680

This avoids any potential clipping/overflow.

As ChannelVolume deals with Samples in the same format they come in as (e.g. i16), previously just using `.saturating_add()` to sum all the samples would mean that the sum would be clipped/overflow if the signals were large enough. By averaging the Samples we avoid this overflow.

### DIFF
--- a/src/source/channel_volume.rs
+++ b/src/source/channel_volume.rs
@@ -34,12 +34,14 @@ where
         I::Item: Sample,
     {
         let mut sample = None;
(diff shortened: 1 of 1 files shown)


dvdsk:
thank you very much, elegant solution using amplify there :+1:

PR #682, opened 2025-01-19
Trigger GitHub release/publish explicitly, scripts cleanup

PetrGlad:
This avoid irrelevant errors during build as release/publish is now triggered manually.  This helps to make it intentional, now it is not obvious that change in Cargo.toml may trigger publication, so some versions may end up on Crates.io accidentally and then be depended upon.
The tag is created only after successful publication, so we know which version was released.
Also some scripts were unnecessarily complex, I tried to streamline that.

One drawback is that one have to wait for tests on master to complete successfully. It is possible to invoke the tests from the release workflow. This is more reliable, but will do some work twice.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,14 +28,11 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: install linux deps
-        run: |
-          sudo apt update
-          sudo apt install -y --no-install-recommends libasound2-dev pkg-config
+      - name: Install linux build requirements
+        run: sudo apt install --yes --no-install-recommends libasound2-dev pkg-config
         if: contains(matrix.os, 'ubuntu')
 
       - name: install ${{ matrix.toolchain }} toolchain
-        id: install_toolchain
         run: rustup toolchain install ${{ matrix.toolchain }} && rustup default ${{ matrix.toolchain }}
 
       - run: cargo clippy -- -D warnings
@@ -55,33 +52,3 @@ jobs:
       # Check minimal build. `tests/seek.rs` fails if there are no decoders at all,
       # adding one to make the tests check pass.
       - run: cargo check --tests --lib --no-default-features --features mp3
-
-  create-git-tag:
-    if: github.event_name == 'push' && github.ref == 'refs/heads/master'
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout the code.
-        uses: actions/checkout@v4
-
-      - name: Fetch tags.
-        #### FIXME Test this (it should fetch all tags without getting the history).
-        run: git fetch --prune --unshallow --tags
-
-      - name: Create a tag for current version if it does not exist yet.
-        id: create_version_tag
-        run: |
-          echo "Current git commit is $(git rev-list -n 1 HEAD)."
-          VERSION="$(yq '.package.version' Cargo.toml)"
-          echo "Project version from Cargo.toml: $VERSION"
-          VERSION_TAG="v$VERSION"
-          if git tag | grep "$VERSION_TAG"; then
-            echo "Tag $VERSION_TAG already exists at $(git rev-list -n 1 $VERSION_TAG), leaving it unchanged."
-          else
-            echo "Creating new tag $VERSION_TAG ..."
-            ###### FIXME Review this using https://github.com/actions/checkout
-            ######        User name and email may need to be updated.
-            git config user.name 'github-actions'
-            git config user.email 'github-actions@github.com'
-            git tag --annotate "$VERSION_TAG" --message "Release version $VERSION_TAG"
-            git push origin "$VERSION_TAG"
-          endif
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,7 @@ name: CI
 on:
   pull_request:
   push:
-    branches: [main, master]
+    branches: [ main, master ]
 
 env:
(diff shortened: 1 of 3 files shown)


.github/workflows/publish.yaml
```diff
@@ -0,0 +1,40 @@
...
+  cargo-publish:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.version_tag }}
+
+      - name: Fetch tags
+        run: git fetch --prune --unshallow --tags
```
PetrGlad, line 15: Default checkout is shallow last-version-only. This also fetches all the tags. Previous version tried to clone the whole repo every time.

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+      - name: Install linux build requirements
+        run: sudo apt install --yes --no-install-recommends libasound2-dev pkg-config
+
+      - name: Publish and tag
+        run: |
+          echo "Current git commit is $(git rev-list -n 1 HEAD)."
+          VERSION="$(yq '.package.version' Cargo.toml)"
```
PetrGlad, line 23: `yq` is installed by defatult in Ubuntu. Unfortunately it cannot write TOML yet. If it could we could also automatically update the cargo version to something like`major.minor.(patch+1)-dev`.

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,40 @@
...
+            exit 1
+          fi
+
+          cargo publish --token "${{ secrets.CRATESIO_TOKEN }}"
+
+          echo "Tagging current version with $VERSION_TAG ..."
+          # The bot name and email is taken from here https://github.com/actions/checkout/pull/1707
+          # see also https://api.github.com/users/github-actions%5Bbot%5D
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
PetrGlad, line 38: This seems to be the actual GitHub user that executes these scripts, and it is recommended in the checkout action documentation.

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,40 @@
+name: Publish
+
+on: workflow_dispatch
```
PetrGlad, line 3: This makes this workflow available for manual triggering on the project's Action tab.

.github/workflows/ci.yml
```diff
@@ -28,14 +28,11 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: install linux deps
-        run: |
-          sudo apt update
```
PetrGlad, line 33: The action's Ubuntu image comes with the package list already. No need to update unless we need a package that was first published several days ago.

.github/workflows/ci.yml
```diff
@@ -28,14 +28,11 @@ jobs:
...
-          sudo apt install -y --no-install-recommends libasound2-dev pkg-config
+      - name: Install linux build requirements
+        run: sudo apt install --yes --no-install-recommends libasound2-dev pkg-config
         if: contains(matrix.os, 'ubuntu')
 
       - name: install ${{ matrix.toolchain }} toolchain
-        id: install_toolchain
```
PetrGlad, line 38: Ids are only needed to refer to this step somewhere else, no need in that right now.

.github/workflows/publish.yaml
dvdsk: lots cleaner then the grep | cut etc hack we had here, nice!

.github/workflows/publish.yaml
dvdsk: good call, we do not release that often so making it a conscious choice seems the right thing to do

.github/workflows/ci.yml
```diff
@@ -55,72 +52,3 @@ jobs:
...
-        version=${{ steps.extract_version.outputs.version }}
-        version_name="v$version"
-        if git rev-parse "refs/tags/$version_name" >/dev/null 2>&1; then
-          echo "Tag $version_name already exists"
-          echo "tag_exists=true" >> $GITHUB_OUTPUT
-        else
-          echo "Tag $version_name does not exist"
-          echo "tag_exists=false" >> $GITHUB_OUTPUT
-        fi
-    - name: Create and push tag
```
PetrGlad, line 118: Splitting these steps into separate runs, in turn required to pass things between them, which is too complex. 
Also sh/bash convention is to use uppercase for variable names.
(If really necessary .inputs and .outputs seems to be [the way to do it](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#example-defining-outputs-for-a-job)).

dvdsk (approved):
Thanks for going through the pain of fixing this. Lets hope it works :)

Good to merge

.github/workflows/ci.yml
dvdsk: it is such pain to pass things between steps..... Honestly for my other projects the github CI just calls a bash file. Bash is just as cursed as these workflows but at least I know bash and I can test it locally.

PetrGlad:
Will try to merge another day, when I have time to debug it. Chances are this won't work on the first try.

dvdsk:
> Will try to merge another day, when I have time to debug it. Chances are this won't work on the first try.

if it does not, just so you know, its not a problem. Worst case we revert the mergen when we want to release again (which will be in many weeks anyway)

Issue #686, opened 2025-01-23
Thread panic: Invalid range for Time fractional part.


will3942:
I have received this panic a few times after running on Rodio's master commit d93b1f332e5758c5cdda41926aa6bcb05cdcf69d

From my tracing I believe this may be related to seeking (https://github.com/RustAudio/rodio/blob/master/src/decoder/symphonia.rs#L180-L212)

```
thread '<unnamed>' panicked at /Users/willevans/.cargo/registry/src/index.crates.io-6f17d22bba15001f/symphonia-core-0.5.4/src/units.rs:197:9:
Invalid range for Time fractional part.
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread '<unnamed>' panicked at core/src/panicking.rs:221:5:
panic in a function that cannot unwind
stack backtrace:
   0:        0x105c8d6e0 - std::backtrace_rs::backtrace::libunwind::trace::h6285d60e8a2c022a
                               at /rustc/90b35a6239c3d8bdabc530a6a0816f7ff89a0aaf/library/std/src/../../backtrace/src/backtrace/libunwind.rs:116:5
   1:        0x105c8d6e0 - std::backtrace_rs::backtrace::trace_unsynchronized::h21fc036d4b8dd7fe
                               at /rustc/90b35a6239c3d8bdabc530a6a0816f7ff89a0aaf/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:        0x105c8d6e0 - std::sys::backtrace::_print_fmt::h69654d1f9cd065f2
                               at /rustc/90b35a6239c3d8bdabc530a6a0816f7ff89a0aaf/library/std/src/sys/backtrace.rs:66:9
   3:        0x105c8d6e0 - <std::sys::backtrace::BacktraceLock::print::DisplayBacktrace as core::fmt::Display>::fmt::h7adefaad4a31afc0
                               at /rustc/90b35a6239c3d8bdabc530a6a0816f7ff89a0aaf/library/std/src/sys/backtrace.rs:39:26
   4:        0x105cae87c - core::fmt::rt::Argument::fmt::h42151196bb7af094
                               at /rustc/90b35a6239c3d8bdabc530a6a0816f7ff89a0aaf/library/core/src/fmt/rt.rs:177:76
   5:        0x105cae87c - core::fmt::write::h381c0b0ce6ab972a
[...]

PR #687, opened 2025-01-23
Fix Symphonia seek for a duration with a fractional element of 0

will3942:
Fixes: https://github.com/RustAudio/rodio/issues/686


will3942:
@dvdsk Ready for review - it was harder to get a 0ms test file than it should have been...

src/decoder/symphonia.rs
```diff
@@ -345,3 +348,22 @@ impl Iterator for SymphoniaDecoder {
         Some(sample)
     }
 }
+
+#[cfg(feature = "symphonia-mp3")]
+#[cfg(test)]
+mod tests {
```
dvdsk, line 354: we can make this an integration test right? It only uses public parts of rodio. It would still need the compile time feature but that could test both the minimp3 and symphonia decoder.

dvdsk (approved):
Ready to merge as is, could make the unit test an integration test. Let me know what you choose and ill merge.

edit: oh and good find! thanks for fixing my bug (this one is my fault) :tada:

will3942:
> Ready to merge as is, could make the unit test an integration test. Let me know what you choose and ill merge.
> 
> edit: oh and good find! thanks for fixing my bug (this one is my fault) 🎉

No worries - have moved this to be an integration test :)

src/decoder/symphonia.rs
```diff
@@ -301,10 +301,13 @@ fn skip_back_a_tiny_bit(
     }: Time,
 ) -> Time {
     frac -= 0.0001;
```
PetrGlad, line 303: Maybe
```rust
if frac < 0.0001 {
} else {
}
```?

Comment (dvdsk) [reply src/decoder/symphonia.rs:303]: its never called at the beginning of the source. This handles seeking beyond a source.

Comment (dvdsk) [reply src/decoder/symphonia.rs:303]: so this is good to merge right?

Comment (PetrGlad) [reply src/decoder/symphonia.rs:303]: If it fixes the problem - yes, we can merge. Since it is already an improvement.
We can handle other cases later if necessary.
The source itself can be less than a second long. The check compares with < 1ms overlap, but tick-back is 0.1ms. what about overlaps in range 0.1...1.0 ms?  I wonder if there are any symphonia functions to work with their timestamps or ways to convert those from/to Duration.

Comment (dvdsk) [reply src/decoder/symphonia.rs:303]: > The source itself can be less than a second long

ahh damn, I did not think of that. Thanks for clarifying. No I agree now, lets fix this properly only then merge. This will need a comment in the code explaining this case or a future refactor might throw the code out.

Comment (roderickvd) [reply src/decoder/symphonia.rs:303]: For which codecs is it necessary to `skip_back_a_teeny_bit`? Code comment says that some codecs need it, but I can’t remember running into it before.

Comment (dvdsk) [reply src/decoder/symphonia.rs:303]: I placed this comment in the code: 
```rust
        let time = if seek_beyond_end {
            let time = self.total_duration.expect("if guarantees this is Some");
            skip_back_a_tiny_bit(time) // some decoders can only seek to just before the end
        } else {
            pos.as_secs_f64().into()
        };
```

As I remember symphonia does not seek if the time is beyond the length of the track. In rodio we have defined seek to saturate at the track length. If symphonia now saturates we can remove this.

roderickvd: When you seek to or beyond the end, you get an `UnexpectedEof` which is anyway what you get when the stream ends.

roderickvd: What do you mean with saturating? I assume the behavior should be that the iterator returns `None`. Is that also what you mean?

I would make (and hav made) the Symphonia decoder such that it handles _this specific end-of-stream_ `UnexpectedEof` that way.

dvdsk: > What do you mean with saturating?

see the docs: https://docs.rs/rodio/latest/rodio/source/trait.Source.html#method.try_seek

*As long as the duration of the source is known seek is guaranteed to saturate at the end of the source. For example given a source that reports a total duration of 42 seconds calling try_seek() with 60 seconds as argument will seek to 42 seconds.*

> I would make (and have made) the Symphonia decoder such that it handles this specific end-of-stream UnexpectedEof that way.

Yeah that's way simpler, go ahead and do that :+1: Maybe I was worried about skipping back? but that's not a feature nor can we make it one.

dvdsk:
the original code, by me, was more fundamentally broken then I and this fix in this PR assumed. In #697 the underlying issue has been addressed. Still this was a very useful investigation enabling that fundamental fix

Issue #696, opened 2025-02-03
total_duration not working with symphonia decoders


dvdsk:
Moving from: https://github.com/RustAudio/rodio/issues/190#issuecomment-2629571291

summary: 
@gulis1 offerd to implement total_duration, I noted it should already be there for symphonia. Then they noted that is not working which I verified locally. Finally I invited them to have a go at fixing it if they would like to and pointed to roderick planning improvements to that piece of code.

roderickvd:
Copying from there:

Yes. Beyond the feature flags, the audio specifications can be None for some codecs until the first packet is decoded. So that needs fixing. I’ll take it along with my refactoring. 

Could you contribute that test? Long time since I did some TDD 😉

dvdsk:
Made a proper integration test, see the branch `tests/total_duration`: https://github.com/RustAudio/rodio/tree/tests/total_duration.rs If whomever picks this up branches from that we will eventually get the test in main. (I've not merged it to main since then CI breaks for all PR's or we have to mark the test as ignored)

roderickvd:
You are right in your analysis. `ReadSeekSource` definitely needs to have a way to have `byte_len` set.

Beyond files there are different ways of supplying a source stream with a known length (in particular, a HTTP stream). And then there are more Symphonia knobs that users might like to turn, or rely on defaults. Therefore I was thinking about making it a builder pattern.

dvdsk:
> Beyond files there are different ways of supplying a source stream with a known length (in particular, a HTTP stream). And then there are more Symphonia knobs that users might like to turn, or rely on defaults. Therefore I was thinking about making it a builder pattern.

Idea, we could add `with_total_duration` to source that modifies any duration from the stream/file. Could be useful for webstreams that have a known `total_duration` but are still in progress.

I agree that a builder could be a good here.

roderickvd:
First proposal for a builder is over at #697.

It doesn't have a `with_total_duration` yet, because I thought to get it from the decoder, but it could be added. It does have the `byte_len` that should let Symphonia determine the `total_duration` for Vorbis files too.

roderickvd:
> Made a proper integration test, see the branch `tests/total_duration`: https://github.com/RustAudio/rodio/tree/tests/total_duration.rs If whomever picks this up branches from that we will eventually get the test in main. (I've not merged it to main since then CI breaks for all PR's or we have to mark the test as ignored)

I merged in that test branch over at #697.

Issue #612, opened 2024-09-16
Refactor: Remove Mp4Type


dvdsk:
Mp4 extensions do not have any real value: https://en.wikipedia.org/wiki/MP4_file_format#Filename_extensions

Therefore the mp4 constructor can just pass mp4. https://github.com/RustAudio/rodio/blob/70c236c7ac495b9ccb82716ffba06cb3fe5fa44c/src/decoder/mod.rs#L314

This makes the function more in line with the rest (no need for a hint) and removes confusion from users.

Issue #527, opened 2023-12-17
even with ```features = ["symphonia-isomp4"]``` mp4 files just give Unrecognized format


Per6:
it is nowhere explained, how to use mp4 files with rodio, just, that it is diabeled by default. I assumed, it would work by enabeling the "symphonia-isomp4" feature however it does not work and returns an error of unrecognized format.
```
let file = File::open(path).expect("unable to open file");
let file = BufReader::new(file);
// mp4 crashes in let source = ...
let source = Decoder::new(file).expect("unable to convert file to a 
music file");
```
i've also tried to use new_mp4 but it doesn't work with any of the Mp4Types and returns the same error just for mp4
```
let source = if path
		.file_name()
		.unwrap()
		.to_string_lossy()
		.ends_with(".mp4")
	{
		Decoder::new_mp4(file, rodio::decoder::Mp4Type::Mp4)
			.expect("unable to convert mp4 file to a music file")
	} else {
		Decoder::new(file).expect("unable to convert file to a music file")
	};

Per6:
I don't know much about codecs, but even with:
rodio = { version = "0.17", features = ["symphonia-isomp4", "symphonia-all"] }
symphonia = { version = "0.5", features = ["isomp4", "all", "all-codecs"] }
it still gives the same error of: unable to convert mp4 file to a music file: UnrecognizedFormat

roderickvd:
You also need `symphonia-aac` (assuming that it's AAC within the MP4 container). Done by default in Rodio 0.21.

PR #697, opened 2025-02-04
Decoder builder with additional settings

roderickvd:
This PR started with the intention to make the Symphonia decoder more robust and expose more of its functionality to users. The changes are implemented through a builder pattern while maintaining compatibility with the existing API.

Along the way, opportunities are taken to refactor the existing codebase.

Commits so far focus on:
- Adding a builder pattern for configuring decoders
- Improving `ReadSeekSource` to better handle seeking and duration calculations
- Making seeking behavior configurable (coarse vs accurate seeking)
- Exposing gapless playback configuration
- Adding comprehensive documentation

Fixes:
- #527 
- #577
- #612
- #696

Supersedes:
- #687

### DIFF
--- a/src/decoder/builder.rs
+++ b/src/decoder/builder.rs
@@ -10,16 +10,20 @@
 //! use std::fs::File;
 //! use rodio::Decoder;
 //!
-//! let file = File::open("audio.mp3").unwrap();
-//! let len = file.metadata().unwrap().len();
+//! fn main() -> Result<(), Box<dyn std::error::Error>> {
+//!     let file = File::open("audio.mp3")?;
+//!     let len = file.metadata()?.len();
 //!
-//! let decoder = Decoder::builder()
-//!     .with_data(file)
-//!     .with_byte_len(len)      // Enable seeking and duration calculation
-//!     .with_hint("mp3")        // Optional format hint
-//!     .with_gapless(true)      // Enable gapless playback
-//!     .build()
-//!     .unwrap();
+//!     Decoder::builder()
+//!         .with_data(file)
+//!         .with_byte_len(len)      // Enable seeking and duration calculation
+//!         .with_hint("mp3")        // Optional format hint
+//!         .with_gapless(true)      // Enable gapless playback
+//!         .build()?;
+//!
+//!     // Use the decoder...
+//!     Ok(())
+//! }
 //! ```
 //!
 //! # Settings
@@ -55,9 +59,7 @@ pub struct Settings {
     /// Can be obtained from file metadata or by seeking to the end of the stream.
     pub(crate) byte_len: Option<u64>,
 
-    /// Whether to use coarse seeking.
-    /// Coarse seeking is faster but less accurate: it may seek to a position slightly before or
-    /// after the requested one, especially when the bitrate is variable.
+    /// Whether to use coarse seeking, or sample-accurate seeking instead.
     pub(crate) coarse_seek: bool,
 
     /// Whether to trim frames for gapless playback.
@@ -101,13 +103,17 @@ impl Default for Settings {
 /// use std::fs::File;
 /// use rodio::decoder::DecoderBuilder;
 ///
-/// let file = File::open("audio.mp3").unwrap();
-/// let decoder = DecoderBuilder::new()
-///     .with_data(file)
-///     .with_hint("mp3")
-///     .with_gapless(true)
-///     .build()
-///     .unwrap();
+/// fn main() -> Result<(), Box<dyn std::error::Error>> {
+///     let file = File::open("audio.mp3")?;
+///     let decoder = DecoderBuilder::new()
+///         .with_data(file)
+///         .with_hint("mp3")
+///         .with_gapless(true)
+///         .build()?;
+///
+///     // Use the decoder...
+///     Ok(())
+/// }
 /// ```
 #[derive(Clone, Debug)]
 pub struct DecoderBuilder<R> {
@@ -134,11 +140,15 @@ impl<R: Read + Seek + Send + Sync + 'static> DecoderBuilder<R> {
     /// use std::fs::File;
     /// use rodio::decoder::DecoderBuilder;
     ///
-    /// let file = File::open("audio.mp3").unwrap();
-    /// let decoder = DecoderBuilder::new()
-    ///     .with_data(file)
-    ///     .build()
-    ///     .unwrap();
+    /// fn main() -> Result<(), Box<dyn std::error::Error>> {
+    ///     let file = File::open("audio.mp3")?;
+    ///     let decoder = DecoderBuilder::new()
+    ///         .with_data(file)
+    ///         .build()?;
+    ///
+    ///     // Use the decoder...
+    ///     Ok(())
+    /// }
     /// ```
     pub fn new() -> Self {
         Self::default()
@@ -160,16 +170,23 @@ impl<R: Read + Seek + Send + Sync + 'static> DecoderBuilder<R> {
     /// use std::fs::File;
     /// use rodio::Decoder;
     ///
-    /// let file = File::open("audio.mp3").unwrap();
-    /// let len = file.metadata().unwrap().len();
-    /// let decoder = Decoder::builder()
-    ///     .with_data(file)
-    ///     .with_byte_len(len)
-    ///     .build()
-    ///     .unwrap();
+    /// fn main() -> Result<(), Box<dyn std::error::Error>> {
+    ///     let file = File::open("audio.mp3")?;
+    ///     let len = file.metadata()?.len();
+    ///     let decoder = Decoder::builder()
+    ///         .with_data(file)
+    ///         .with_byte_len(len)
+    ///         .build()?;
+    ///
+    ///     // Use the decoder...
+    ///     Ok(())
+    /// }
     /// ```
     ///
     /// Alternatively, it can be obtained by seeking to the end of the stream.
+    ///
+    /// An incorrect byte length can lead to unexpected behavior, including but not limited to
+    /// incorrect duration calculations and seeking errors.
     pub fn with_byte_len(mut self, byte_len: u64) -> Self {
         self.settings.byte_len = Some(byte_len);
         self
@@ -177,7 +194,7 @@ impl<R: Read + Seek + Send + Sync + 'static> DecoderBuilder<R> {
 
     /// Enables or disables coarse seeking.
     ///
-    /// This needs byte_len to be set. Coarse seeking is faster but less accurate:
+    /// This needs `byte_len` to be set. Coarse seeking is faster but less accurate:
     /// it may seek to a position slightly before or after the requested one,
     /// especially when the bitrate is variable.
     pub fn with_coarse_seek(mut self, coarse_seek: bool) -> Self {
@@ -195,7 +212,7 @@ impl<R: Read + Seek + Send + Sync + 'static> DecoderBuilder<R> {
 
     /// Sets a format hint for the decoder.
     ///
-    /// When known, this can help the decoder to select the correct codec.
+    /// When known, this can help the decoder to select the correct codec faster.
     /// Common values are "mp3", "wav", "flac", "ogg", etc.
     pub fn with_hint(mut self, hint: &str) -> Self {
         self.settings.hint = Some(hint.to_string());
@@ -204,7 +221,7 @@ impl<R: Read + Seek + Send + Sync + 'static> DecoderBuilder<R> {
 
     /// Sets a mime type hint for the decoder.
     ///
-    /// When known, this can help the decoder to select the correct demuxer.
+    /// When known, this can help the decoder to select the correct demuxer faster.
     /// Common values are "audio/mpeg", "audio/vnd.wav", "audio/flac", "audio/ogg", etc.
     pub fn with_mime_type(mut self, mime_type: &str) -> Self {
         self.settings.mime_type = Some(mime_type.to_string());
@@ -214,7 +231,7 @@ impl<R: Read + Seek + Send + Sync + 'static> DecoderBuilder<R> {
     /// Configure whether the decoder should report as seekable.
     ///
     /// For reliable seeking behavior, `byte_len` should be set either from file metadata
-    /// or by seeking to the end of the stream. While seeking may work without byte_len
+    /// or by seeking to the end of the stream. While seeking may work without `byte_len`
     /// for some formats, it is not guaranteed.
     ///
     /// # Examples
@@ -222,16 +239,20 @@ impl<R: Read + Seek + Send + Sync + 'static> DecoderBuilder<R> {
     /// use std::fs::File;
     /// use rodio::Decoder;
     ///
-    /// let file = File::open("audio.mp3").unwrap();
-    /// let len = file.metadata().unwrap().len();
+    /// fn main() -> Result<(), Box<dyn std::error::Error>> {
+    ///     let file = File::open("audio.mp3")?;
+    ///     let len = file.metadata()?.len();
     ///
-    /// // Recommended: Set both byte_len and seekable
-    /// let decoder = Decoder::builder()
-    ///     .with_data(file)
-    ///     .with_byte_len(len)
-    ///     .with_seekable(true)
-    ///     .build()
-    ///     .unwrap();
+    ///     // Recommended: Set both byte_len and seekable
+    ///     let decoder = Decoder::builder()
+    ///         .with_data(file)
+    ///         .with_byte_len(len)
+    ///         .with_seekable(true)
+    ///         .build()?;
+    ///
+    ///     // Use the decoder...
+    ///     Ok(())
+    /// }
     /// ```
     pub fn with_seekable(mut self, is_seekable: bool) -> Self {
         self.settings.is_seekable = is_seekable;
--- a/src/decoder/builder.rs
+++ b/src/decoder/builder.rs
@@ -1,0 +1,307 @@
+//! Builder pattern for configuring and constructing decoders.
+//!
+//! This module provides a flexible builder API for creating decoders with custom settings.
+//! The builder allows configuring format hints, seeking behavior, byte length and other
+//! parameters that affect decoder behavior.
+//!
+//! # Examples
+//!
+//! ```no_run
+//! use std::fs::File;
+//! use rodio::Decoder;
+//!
+//! let file = File::open("audio.mp3").unwrap();
+//! let len = file.metadata().unwrap().len();
+//!
+//! let decoder = Decoder::builder()
+//!     .with_data(file)
+//!     .with_byte_len(len)      // Enable seeking and duration calculation
+//!     .with_hint("mp3")        // Optional format hint
+//!     .with_gapless(true)      // Enable gapless playback
+//!     .build()
+//!     .unwrap();
+//! ```
+//!
+//! # Settings
+//!
+//! The following settings can be configured:
+//!
+//! - `byte_len` - Total length of the input data in bytes
+//! - `hint` - Format hint like "mp3", "wav", etc
+//! - `mime_type` - MIME type hint for container formats
+//! - `seekable` - Whether seeking operations are enabled
+//! - `gapless` - Enable gapless playback
+//! - `coarse_seek` - Use faster but less precise seeking
+
+use std::io::{Read, Seek};
+
+#[cfg(feature = "symphonia")]
+use self::read_seek_source::ReadSeekSource;
+#[cfg(feature = "symphonia")]
+use ::symphonia::core::io::{MediaSource, MediaSourceStream};
+
+use super::*;
+
+/// Audio decoder configuration settings.
+/// Support for these settings depends on the underlying decoder implementation.
+/// Currently, settings are only used by the Symphonia decoder.
+#[derive(Clone, Debug)]
+pub struct Settings {
+    /// The length of the stream in bytes.
+    /// This is required for:
+    /// - Reliable seeking operations
+    /// - Duration calculations in formats that lack timing information (e.g. MP3, Vorbis)
+    ///
+    /// Can be obtained from file metadata or by seeking to the end of the stream.
+    pub(crate) byte_len: Option<u64>,
+
+    /// Whether to use coarse seeking.
+    /// Coarse seeking is faster but less accurate: it may seek to a position slightly before or
+    /// after the requested one, especially when the bitrate is variable.
+    pub(crate) coarse_seek: bool,
+
+    /// Whether to trim frames for gapless playback.
+    /// Note: Disabling this may affect duration calculations for some formats
+    /// as padding frames will be included.
+    pub(crate) gapless: bool,
+
+    /// An extension hint for the decoder about the format of the stream.
+    /// When known, this can help the decoder to select the correct codec.
+    pub(crate) hint: Option<String>,
+
+    /// An MIME type hint for the decoder about the format of the stream.
+    /// When known, this can help the decoder to select the correct demuxer.
+    pub(crate) mime_type: Option<String>,
+
+    /// Whether the decoder should report as seekable.
+    pub(crate) is_seekable: bool,
+}
+
+impl Default for Settings {
+    fn default() -> Self {
+        Self {
+            byte_len: None,
+            coarse_seek: false,
+            gapless: true,
+            hint: None,
+            mime_type: None,
+            is_seekable: false,
+        }
+    }
+}
+
+/// Builder for configuring and creating a decoder.
+///
+/// This provides a flexible way to configure decoder settings before creating
+/// the actual decoder instance.
+///
+/// # Examples
+///
+/// ```no_run
+/// use std::fs::File;
+/// use rodio::decoder::DecoderBuilder;
+///
+/// let file = File::open("audio.mp3").unwrap();
+/// let decoder = DecoderBuilder::new()
+///     .with_data(file)
+///     .with_hint("mp3")
+///     .with_gapless(true)
+///     .build()
+///     .unwrap();
+/// ```
+#[derive(Clone, Debug)]
+pub struct DecoderBuilder<R> {
+    /// The input data source to decode.
+    data: Option<R>,
+    /// Configuration settings for the decoder.
+    settings: Settings,
+}
+
+impl<R> Default for DecoderBuilder<R> {
+    fn default() -> Self {
+        Self {
+            data: None,
+            settings: Settings::default(),
+        }
+    }
+}
+
+impl<R: Read + Seek + Send + Sync + 'static> DecoderBuilder<R> {
+    /// Creates a new decoder builder with default settings.
+    ///
+    /// # Examples
+    /// ```no_run
+    /// use std::fs::File;
+    /// use rodio::decoder::DecoderBuilder;
+    ///
+    /// let file = File::open("audio.mp3").unwrap();
+    /// let decoder = DecoderBuilder::new()
+    ///     .with_data(file)
+    ///     .build()
+    ///     .unwrap();
+    /// ```
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Sets the input data source to decode.
+    pub fn with_data(mut self, data: R) -> Self {
+        self.data = Some(data);
+        self
+    }
+
+    /// Sets the byte length of the stream.
+    /// This is required for:
+    /// - Reliable seeking operations
+    /// - Duration calculations in formats that lack timing information (e.g. MP3, Vorbis)
+    ///
+    /// The byte length should typically be obtained from file metadata:
+    /// ```no_run
+    /// use std::fs::File;
+    /// use rodio::Decoder;
+    ///
+    /// let file = File::open("audio.mp3").unwrap();
+    /// let len = file.metadata().unwrap().len();
+    /// let decoder = Decoder::builder()
+    ///     .with_data(file)
+    ///     .with_byte_len(len)
+    ///     .build()
+    ///     .unwrap();
+    /// ```
+    ///
+    /// Alternatively, it can be obtained by seeking to the end of the stream.
+    pub fn with_byte_len(mut self, byte_len: u64) -> Self {
+        self.settings.byte_len = Some(byte_len);
+        self
+    }
+
+    /// Enables or disables coarse seeking.
+    ///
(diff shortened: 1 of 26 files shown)


gulis1:
The builder looks great to me, thanks for the work!

Maybe we could add a new `with_file` or `from_file` method? I think playing audio from files is one of the most common things new users will use Rodio for, so having a method that extracts the `byte_len` from the file's metadata (specially for ogg vorbis files)`would be helpful.

roderickvd:
> Maybe we could add a new `with_file` or `from_file` method? I think playing audio from files is one of the most common things new users will use Rodio for, so having a method that extracts the `byte_len` from the file's metadata (specially for ogg vorbis files)`would be helpful.

Yes, I've got a `with_file` cooked up that I'll push hopefully tomorrow.

roderickvd:
Probably should re-add `Decoder::builder` as a shorthand to `DecoderBuilder::new`.

src/decoder/mod.rs
```diff
@@ -494,26 +593,38 @@ impl<R: Read + Seek> DecoderImpl<R> {
...
+/// # Examples
+/// ```no_run
+/// use std::fs::File;
+/// use std::convert::TryFrom;
+/// use rodio::Decoder;
+///
+/// let file = File::open("audio.mp3").unwrap();
+/// let decoder = Decoder::try_from(file).unwrap();
+/// ```
+impl TryFrom<std::fs::File> for Decoder<BufReader<std::fs::File>> {
```
dvdsk, line 599: is the idea to have multiple `TryFrom` implementations? (so one for BufReader for example). Otherwise making a normal member function might make more sense.

src/decoder/mod.rs
```diff
@@ -494,26 +593,38 @@ impl<R: Read + Seek> DecoderImpl<R> {
...
+/// Converts a `File` into a `Decoder` with automatic optimizations.
+///
+/// This implementation:
+/// - Wraps the file in a `BufReader` for better performance
+/// - Sets `byte_len` from file metadata when available, which can improve seeking operations
+///
+/// # Examples
+/// ```no_run
+/// use std::fs::File;
+/// use std::convert::TryFrom;
```
dvdsk, line 593: is actually in the prelude since rust 2021

src/decoder/mod.rs
```diff
@@ -69,8 +114,357 @@ where
...
+///     .unwrap();
+/// ```
+#[derive(Clone, Debug)]
+pub struct DecoderBuilder<R> {
+    /// The input data source to decode.
+    data: Option<R>,
+    /// Configuration settings for the decoder.
+    settings: Settings,
+    /// Whether the decoder should loop playback.
+    looped: bool,
```
dvdsk, line 182: instead of this flag we can have `build` & `build_looped`

src/decoder/mod.rs
dvdsk, line 1: this is getting a bit long maybe? Might be a good idea to split the loopeddecoder and builder each to its own file

src/decoder/mod.rs
```diff
@@ -188,217 +576,255 @@ impl<R: Read + Seek> DecoderImpl<R> {
...
 
-        #[cfg(feature = "symphonia")]
-        {
-            let mss = MediaSourceStream::new(
-                Box::new(ReadSeekSource::new(data)) as Box<dyn MediaSource>,
-                Default::default(),
-            );
+    fn try_from(file: std::fs::File) -> Result<Self, Self::Error> {
+        let mut builder = DecoderBuilder::new();
+        if let Some(len) = file.metadata().ok().map(|m| m.len()) {
```
dvdsk, line 604: We could return an error if the metadata is not available. If we can play the file we should be able to get the metadata for it. In (I would say extreme) edge cases where there is no metadata and for some reason the file is playable users can now use the builder :tada: .

src/decoder/mod.rs
```diff
@@ -188,217 +576,255 @@ impl<R: Read + Seek> DecoderImpl<R> {
...
-            Ok(decoder) => Ok(Decoder(DecoderImpl::Wav(decoder))),
+    ///
+    /// # Errors
+    ///
+    /// Returns `DecoderError::UnrecognizedFormat` if the audio format could not be determined
+    /// or is not supported.
+    pub fn new(data: R) -> Result<Self, DecoderError> {
+        match DecoderBuilder::new().with_data(data).build()? {
+            DecoderOutput::Normal(decoder) => Ok(decoder),
+            DecoderOutput::Looped(_) => unreachable!("Builder defaults to non-looped"),
```
dvdsk, line 627: splitting into `build` and `build_looped` would also take care of this.

src/decoder/mod.rs
```diff
@@ -494,56 +927,110 @@ where
...
+    ///
+    /// Even for looped playback, seeking past the end of the stream will not automatically
+    /// wrap around to the beginning - it will return an error just like a normal decoder.
+    /// Looping only occurs when reaching the end through normal playback.
     fn try_seek(&mut self, pos: Duration) -> Result<(), SeekError> {
-        self.0.try_seek(pos)
+        match &mut self.inner {
+            Some(inner) => inner.try_seek(pos),
+            None => Err(SeekError::NotSupported {
+                underlying_source: "No decoder available",
```
dvdsk, line 1027: this can only be None if the "reset" seek failed. 

Maybe we can do 
```rust
struct LoopBackFailure;

impl Error for LoopBackFailure {
  .. some msg like "Looped source ended when it failed to loop back"
}

SeekError::Other(Box::new(LoopBackFailure))
```

dvdsk:
What should we do about other decoders? I think its out of the scope of this PR to check what features they could support and implement them. We should still communicate to users that the chose feature set is not supported. 

Maybe decoder creation should return a Error feature not supported? (and we add a bold warning to the builder options that could cause that). The defaults should probably be picked such that they do not produce that error.

src/decoder/symphonia.rs
```diff
@@ -197,7 +202,7 @@ impl Source for SymphoniaDecoder {
         let seek_res = self
             .format
             .seek(
-                SeekMode::Accurate,
```
dvdsk, line 200: GitHub wont let me leave a comment where I want to.. So this is about `skip_back_a_tiny_bit`.

Like you said in #687 we should make it so the source returns None next if the seek is beyond the end. That might still fit in this PR.

dvdsk (commented):
..

src/decoder/mod.rs
roderickvd: I added implementations for `BufReader` and `Cursor<T: AsRef[u8]>`.

src/decoder/mod.rs
roderickvd: That's a great idea. Thanks for pulling me out of that tunnel.

src/decoder/mod.rs
roderickvd: Removed & done.

src/decoder/mod.rs
roderickvd: Yes, will do. Not done yet.

src/decoder/mod.rs
roderickvd: I agree. Done.

src/decoder/mod.rs
roderickvd: Done.

src/decoder/mod.rs
roderickvd: I now re-used the existing `DecoderError::IoError`. Seems good enough without introducing another error type, do you agree?

src/decoder/symphonia.rs
roderickvd: Removed that `skip_back_a_tiny_bit` and threw in another small optimization with getting `total_duration`.

I was thinking of doing quite some more refactoring of the Symphonia decoder, but now think it's better to get this "builder" PR settled. To do meaningful refactoring there, I need to get this span length thing down first anyway.

roderickvd:
> What should we do about other decoders? I think its out of the scope of this PR to check what features they could support and implement them. We should still communicate to users that the chose feature set is not supported.
>
> Maybe decoder creation should return a Error feature not supported? (and we add a bold warning to the builder options that could cause that). The defaults should probably be picked such that they do not produce that error.

Today, without this PR, the Symphonia decoder is gapless (trims) and has seeking functionality in a lot of cases.
The other decoders do not without much of a hint to the user. (yeah, `Err` when attempting to seek)

`Settings::default` in this PR shares with current behavior that gapless is enabled by default, but differs in that `is_seekable` is `false` by default. That's in order to deal with the upcoming fix in [Rodio#340](https://github.com/pdeljanov/Symphonia/pull/340): `byte_len` needs to be set for correct seeking behavior, either from metadata or by manually doing `Seek::seek(SeekFrom::End(0))`.

At one point I thought about having builder methods feature-gated, but that won't work because of https://github.com/RustAudio/rodio/issues/539.

So that leaves us with:
1. soft-failure like now, with documented behavior
2. printing / logging a message
3. throwing an error somewhere

Numbers 2 and 3 could only be done in the `build` / `build_looped` phase, because only at that point will we know whether, say, an Ogg Vorbis file will be decoded by `lewton` or Symphonia.

I'm kind of partial to number 1: settings only do something with Symphonia and that's that.
I'll do a check to see what settings the other decoders could take, but I don't think it's much (or anything at all).

Besides that check, and moving stuff out into manageable-sized files, I'm having a bit of a fight with the `seek.rs` integration test for `RL.ogg` and `RL.m4a`. They were disabled before, but should be working now. Any ideas?

> reviewed up to [a4db631](https://github.com/RustAudio/rodio/commit/a4db631a6aa2996eccb1ebe2a08c4ed0824f3dba)

🙏

src/decoder/mod.rs
dvdsk: for context this is with IoError
```rust
            None => Err(SeekError::Other(Box::new(DecoderError::IoError(
                "Looped source ended when it failed to loop back".to_string(),
            )))),
```

and that looks great to me :+1:

src/decoder/symphonia.rs
dvdsk: agreed, lets do that later :+1:

dvdsk:
> I'm having a bit of a fight with the `seek.rs` integration test for `RL.ogg` and `RL.m4a`. They were disabled before, but should be working now. Any ideas?

I see, its the `seek_does_not_break_channel_order` test that is failing. That is strange given it works with the other symphonia decoders. Its not a difference in test files, I've taken a quick look in audacity: `RL.wav` & `RL.ogg` look identical.

Might be time to test by hand, just to be sure the test is correct. Try making an example that plays the left right and check if the beep switches speaker. The `wav` and `ogg` should have the same switching. (That is what the test attempts to check).

roderickvd:
> > I'm having a bit of a fight with the `seek.rs` integration test for `RL.ogg` and `RL.m4a`. They were disabled before, but should be working now. Any ideas?
> 
> I see, its the `seek_does_not_break_channel_order` test that is failing. That is strange given it works with the other symphonia decoders. Its not a difference in test files, I've taken a quick look in audacity: `RL.wav` & `RL.ogg` look identical.
> 
> Might be time to test by hand, just to be sure the test is correct. Try making an example that plays the left right and check if the beep switches speaker. The `wav` and `ogg` should have the same switching. (That is what the test attempts to check).

I've found that I can get the test to pass by increasing the sample window in `tests/seek.rs` from 100 to ~1000 samples. When seeking these Ogg Vorbis and MP4/AAC files, it turns out that the actual position is slightly under 100 time units off from the target position.

It doesn't sit very well with me, and I'm finding `SymphoniaDecoder::refine_position` a bit suspect. I'm struggling to understand it. I assume its purpose is to do a linear search forward to the target position. Where I'm struggling is that to do that, it immediately goes for the `next_packet` and decode that.

What if the target position is actually in the current buffer, before the next packet? At least those should be deducted of the `samples_to_pass` is what I'm thinking. But like I said, I'm struggling, so do not hesitate to let me know what I'm missing.

> reviewed up to [a4db631](https://github.com/RustAudio/rodio/commit/a4db631a6aa2996eccb1ebe2a08c4ed0824f3dba)

I rebased to pull in #699. For the record your review was now up to [fb1b777](https://github.com/RustAudio/rodio/pull/697/commits/fb1b7774c251630cba4d6b4ada58be396a26f4ef).

roderickvd:
Also this:

```rust
let mut samples_to_pass = seek_res.required_ts - seek_res.actual_ts;
// ...
self.current_span_offset = samples_to_pass as usize * self.channels() as usize;
```

The unit of `required_ts` and `actual_ts` is timestamp in `TimeBase` units, not number of samples.
So it's more like a `time_to_pass` from which we would have to calculate the number of samples.

dvdsk:
> Playing around with this but it doesn't yet pass the tests:

is it just the ogg thats failing or the rest too?

roderickvd:
> thanks! & `decoder.reset()` seems to also be missing.

Yes, added that in my code snippet [above](https://github.com/RustAudio/rodio/pull/697#issuecomment-2652082528) that I'm trying to get to work appropriately. Maybe I need to start looking at the angle of the tests too, see whether they are correct in full.

> is it just the ogg thats failing or the rest too? To my eye it _should_ work, especially since the previous code did (on all but ogg).

Currently failing all five cases of `seek_does_not_break_channel_order`. So something is really off.

By the way, merely putting `self.decoder.reset()` in the current master also causes breakage.
Will investigate further, but have a bit less time on my hands the coming days.

roderickvd:
I think I'm onto it: due to the truncating behavior of calculating the number of samples from the duration, the code above does not guarantee it is aligned with the correct channel. Working on it to make it so.

roderickvd:
😅 that should do it. Next to my hunch in https://github.com/RustAudio/rodio/pull/697#issuecomment-2661056356, there was also the case for Ogg & MP4 that rewinding a stream would successfully decode a packet with zero samples and only metadata, which caused the iterator to end.

Even so, I disabled one specific MP4 test (`seek_does_not_break_channel_order`) because there are several outstanding issues with Symphonia that make position calculation unreliable. There's multiple issues and PRs handling that, so we should re-check when Symphonia 0.5.5 or greater is released.

Increasing the test sample window from 100 to 1000 makes the test pass, but hides the underlying problem.

Props on having all these tests; they caught a lot of stuff.

While I still have more than a few ideas to refactor the Symphonia decoder, I am going to move this out of draft now. Appreciate your review to get this settled first.

src/decoder/mod.rs
roderickvd: Still have to take this one on.

roderickvd: Done.

src/decoder/symphonia.rs
```diff
@@ -174,39 +182,43 @@ impl Source for SymphoniaDecoder {
...
+        // so we need to reset the decoder to make sure it's in sync and prevent
+        // audio glitches.
+        self.decoder.reset();
+
+        // Force the iterator to decode the next packet.
+        self.current_span_offset = usize::MAX;
+
+        if matches!(self.seek_mode, SeekMode::Accurate) {
+            self.refine_position(seek_res)?;
+            self.current_span_offset += active_channel;
```
roderickvd, line 220: This needs to apply to coarse seeking as well.

src/decoder/builder.rs
```diff
@@ -0,0 +1,307 @@
...
+}
+
+/// Builder for configuring and creating a decoder.
+///
+/// This provides a flexible way to configure decoder settings before creating
+/// the actual decoder instance.
+///
+/// # Examples
+///
+/// ```no_run
```
dvdsk, line 100: rust documentation guidelines recommend to not use unwrap in example (users tend to copy them verbatim and keep them in). We copied that into our own [guidlines](https://github.com/RustAudio/rodio/blob/master/CONTRIBUTING.md#documentation)

In practice that means replace the `.unwrap` with `?` and start the example with:
`# fn main() -> Result<(), Box<dyn std::error::Error>>`
while ending with
`# Ok(())`

src/decoder/builder.rs
```diff
@@ -0,0 +1,307 @@
...
+    /// let len = file.metadata().unwrap().len();
+    /// let decoder = Decoder::builder()
+    ///     .with_data(file)
+    ///     .with_byte_len(len)
+    ///     .build()
+    ///     .unwrap();
+    /// ```
+    ///
+    /// Alternatively, it can be obtained by seeking to the end of the stream.
+    pub fn with_byte_len(mut self, byte_len: u64) -> Self {
```
dvdsk, line 173: might be worth to mention what happens if I set an incorrect byte_len. What does byte_len smaller then actual len do? Same for the inverse.

src/decoder/builder.rs
```diff
@@ -0,0 +1,307 @@
...
+    pub fn with_hint(mut self, hint: &str) -> Self {
+        self.settings.hint = Some(hint.to_string());
+        self
+    }
+
+    /// Sets a mime type hint for the decoder.
+    ///
+    /// When known, this can help the decoder to select the correct demuxer.
+    /// Common values are "audio/mpeg", "audio/vnd.wav", "audio/flac", "audio/ogg", etc.
+    pub fn with_mime_type(mut self, mime_type: &str) -> Self {
```
dvdsk, line 209: why is there both a mime type and a hint? (To be clear, this is just me not understanding). Its all audio right?

dvdsk (commented):
Not yet done. Like what I've seen so far. Huge increase in code quality and documentation compared to what we had before! :tada: :balloon: :partying_face:

src/decoder/symphonia.rs
```diff
@@ -181,37 +182,43 @@ impl Source for SymphoniaDecoder {
...
-        self.current_span_offset += to_skip;
+        // Seeking is a demuxer operation without the decoder knowing about it,
+        // so we need to reset the decoder to make sure it's in sync and prevent
+        // audio glitches.
+        self.decoder.reset();
+
+        // Force the iterator to decode the next packet.
+        self.current_span_offset = usize::MAX;
+
+        if matches!(self.seek_mode, SeekMode::Accurate) {
```
dvdsk, line 218: `if let SeekMode::Accurate = self.seek_mode {` _might_ be neater (your call).

src/decoder/symphonia.rs
```diff
@@ -273,72 +282,60 @@ impl std::error::Error for SeekError {
...
-        for _ in 0..MAX_DECODE_RETRIES {
-            if decoded.is_err() {
-                let packet = self.format.next_packet().map_err(SeekError::Retrying)?;
-                decoded = self.decoder.decode(&packet);
-            }
+        // Calculate the number of samples to skip.
+        let mut samples_to_skip = (Duration::from(
+            self.time_base
+                .ok_or(source::SeekError::SymphoniaDecoder(SeekError::Refining(
+                    "time base not available",
```
dvdsk, line 289: what does it mean for the time base to be not available? Is that a bug in symphonia or a property of the file being decoded? Specifically is there something a user can do if they encounter this error?

src/decoder/symphonia.rs
```diff
@@ -273,72 +282,60 @@ impl std::error::Error for SeekError {
...
 
-    #[inline]
     fn next(&mut self) -> Option<Self::Item> {
         if self.current_span_offset >= self.buffer.len() {
-            let packet = self.format.next_packet().ok()?;
-            let mut decoded = self.decoder.decode(&packet);
-            for _ in 0..MAX_DECODE_RETRIES {
+            let mut decode_errors = 0;
+            let decoded = loop {
+                let packet = self.format.next_packet().ok()?;
```
dvdsk, line 317: just to make sure this is on your radar (for the future, not this pr). `next_packet` could return `ResetRequired`: 

> then the track list must be re-examined and all `Decoder`s re-created.

src/decoder/symphonia.rs
```diff
@@ -273,72 +282,60 @@ impl std::error::Error for SeekError {
...
                 }
-            }
-            let decoded = decoded.ok()?;
+
+                // Loop until we get a packet with audio frames. This is necessary because some
+                // formats can have packets with only metadata, particularly when rewinding, in
+                // which case the iterator would otherwise end with `None`.
+                // Note: checking `decoded.frames()` is more reliable than `packet.dur()`, which
+                // can resturn non-zero durations for packets without audio frames.
+                let decoded = decoded.ok()?;
```
dvdsk, line 333: this `.ok()?` should be an `.expect("if above handles Err variant")` or the `if decoded.is_err` could become a match returning the `Ok` value.

src/decoder/symphonia.rs
```diff
@@ -273,72 +282,60 @@ impl std::error::Error for SeekError {
...
+                if decoded.frames() > 0 {
+                    break decoded;
+                }
+            };
+
```
dvdsk, line 338: (comment regarding existing code not in the scope of this PR)

not sure about `self.current_span_offset = 0;`. Symphonia does not specify that a packet ends at a frame boundary. I think it does that in practice, but its not promised and might change in the future. Maybe we should handle that here.

src/decoder/builder.rs
roderickvd: I'll add something and keep it short. Basically it's just wrong and I propose we shy away from making any promises about what happens then. Currently, Symphonia is likely to report incorrect `total_duration` and seek sub-optimally or to unexpected positions.

src/decoder/builder.rs
roderickvd: Can't speak for Symphonia but I would explain as follows:
- for some sources (like HTTP streams) you may know the MIME content type but not the extension
- for other sources (like files) it may be the other way around

Further: as per #612 the `.m4x` extensions are quite meaningless, but at least the MIME type is correct. Symphonia [understand both](https://github.com/pdeljanov/Symphonia/blob/ef9bbd8dd147b05cc911dafe0ae3663ae81b692d/symphonia-format-isomp4/src/demuxer.rs#L313) but setting the MIME type seems a tad bit more accurate here to me.

src/decoder/symphonia.rs
roderickvd: I can think of two cases where this might happen:
1. the source is malformed and misses the time base metadata
2. the source has time base metadata in an unsupported format

As a case in point for point 2, I think Symphonia only recently added support for "MOOV" atoms in MP4 although it could play the AAC fine.

Currently, it will mean that accurate seeking will bail out. If we want something else, I see a couple of alternatives:
1. We could fall back to coarse seeking. Would we want to do that silently?
2. We could assume that the time base is one sample per Hz per channel. Could be right in a lot of cases, in other cases it would _seem_ to be seeking accurately but in the end not so.
3. We could offer a builder option to set the time base. Should it override a "decoded" time base or serve as backup?
4. Or maybe simply improve the error message with "time base not available; accurate seeking not supported".

I'm not so much a fan of option 3, because:
a. I think it's overly advanced and out of balance with the other options
b. it requires in-depth user knowledge of the source which seems unlikely

src/decoder/symphonia.rs
roderickvd: Yes, on the radar. Currently only occurs on chained Ogg streams. Relevant test files are available at https://people.xiph.org/~xiphmont/test-vectors/vorbis/ - could we include them? Doesn't state the license and then you could debate whether they are defined as "software" under the [Xiph License](https://xiph.org/flac/license.html) or not.

src/decoder/symphonia.rs
roderickvd: Good catch.

src/decoder/builder.rs
dvdsk: > I propose we shy away from making any promises about what happens then.

agreed. In that case we should note that

src/decoder/symphonia.rs
roderickvd: This is the problem I have tried to express in several other issues and PRs: making guarantees about span length or offset is really hard, if not impossible, with decoded sources.

src/decoder/builder.rs
dvdsk: and the hints speed up the choice of the correct decoder? Or is it possible to pick a wrong format when doing so automatically (like `build_impl` does)?

roderickvd: Yes, it speeds up probing by prioritizing the hinted demuxer/decoder. The probe continue with other demuxers/decoders if the provided hints did not yield the expected result.

src/decoder/symphonia.rs
dvdsk: Very clear. I like option 4 the best. If the time base is only needed for accurate seeking would it make sense to make the error something like `SeekError::AccurateSeekNotSupported`? Users can then match on the SeekError and handle it by seeking coarsely.

src/decoder/builder.rs
dvdsk: In that case maybe we could add the word faster to the docs, concretely: `this can help the decoder to select the correct demuxer faster`. Since now a user might think the mime/hint is needed for correct functionality in some edge case.

src/decoder/symphonia.rs
dvdsk: go ahead and add them, I'll shoot them a mail. If they tell me its not okay we can remove them.

src/decoder/symphonia.rs
dvdsk: > making guarantees about span length or offset is really hard

agreed on span length, it is neigh on impossible given seeks may happen at any time. The only workaround is injecting silence but that is:
 - way to complex to do for every source
 - a horrible user experience
So lets just ignore span length here, we are doing away with that anyway and focus on offset.

It should be possible to keep the offset correct by tracking the number of samples emitted by the decoder since its start. Though that could get tricky if the channel count changes across a seek or packet boundary. In that case repeating the last sample might be the best solution with the least distortion. On the other hand such a channel count seems exceedingly rare to me, we could just ignore that and handle it once someone runs into it and opens an issue?

dvdsk: > This is the problem I have tried to express in several other issues and PRs

If you feel like I'm simply not getting you we could clear things up on a call. That's might be faster/easier then having to type everything out all the time. You can always mail me at opensource at the domain on my Github page. 

On a related note, I'll be attending [Rust-Week](https://rustweek.org/) this may if anyone wants to meet up.

roderickvd: > It should be possible to keep the offset correct by tracking the number of samples emitted by the decoder since its start.

Yes, assuming packets end at a frame boundary (wrapping to the start of this discussion).

> Though that could get tricky if the channel count changes across a seek or packet boundary.

Honestly don't know: do sources exist where, _within the same track_, channel count or sample rate changes? It's possible for _different tracks_ within a container, yes. For seeks that will be fine as long as we seek within the same track (today it always assumes the "default" track)*. For `next_packet` we'll be able to handle it (in a future PR) with `ResetRequired `.

\* need to fix that too: because `next_packet` will continue iterating over all tracks in a container, seeking will be broken once you move past the first track (which is typically the "default" track).

> In that case repeating the last sample might be the best solution with the least distortion.

Zero-order hold also introduces other undesirable effects, like DC offset and harmonic distortion.

> On the other hand such a channel count seems exceedingly rare to me, we could just ignore that and handle it once someone runs into it and opens an issue?

I agree: it's an edge case that's a _goal_ to cover (and on the radar indeed) but not _required_.

> If you feel like I'm simply not getting you we could clear things up on a call. That's might be faster/easier then having to type everything out all the time. You can always mail me at opensource at the domain on my Github page.

Not that, was saying so to try and get convergence on the different (many) issues and PRs about spans. In my dream world we'd dispense with them entirely and find some other way for filters that depend on them today.

Since we're both in the NL though, I did think about proposing a beer at some point if you (and/or others) would be up to that. Off to bed now but will write you later; open to receiving e-mail from you also.

dvdsk: > Yes, assuming packets end at a frame boundary (wrapping to the start of this discussion).

that is something we could track right? A `u64` can function as a running sample counter. It would hold the number of samples ever emitted by the source. We then check that. 

It might get complex if we start handling changes in channel count. Still I think even that can be handled. At that point we can reset the counter to zero and do a zero-order hold/repeat the samples until we have a frame boundary again. If a user decides to seek right in the middle of that 'frame boundary fixing' we will have to handle that.

But lets not do any of this, lets either hope or ask if packets end at frame boundaries.

> Not that, was saying so to try and get convergence on the different (many) issues and PRs about spans.

I'll @ you in the span issue to get that done.

> Since we're both in the NL though, I did think about proposing a beer at some point if you (and/or others) would be up to that. Off to bed now but will write you later; open to receiving e-mail from you also.

Lets do that. I am located in Leiden btw.

dvdsk:
this will also need some minor rebasing/merging/fixing because the `change generic sample to f32` change landed :tada:. It should be relatively easy though. Let me know if you need help.

roderickvd:
I've addressed all relevant review comments in [dc324f9](https://github.com/RustAudio/rodio/pull/697/commits/dc324f9bbc76461de48c7710dd350ec7482dbd35). For ease of reviewing, I propose I rebase _after_ the review is otherwise accepted.

src/decoder/builder.rs
```diff
@@ -55,9 +59,7 @@ pub struct Settings {
     /// Can be obtained from file metadata or by seeking to the end of the stream.
     pub(crate) byte_len: Option<u64>,
 
-    /// Whether to use coarse seeking.
-    /// Coarse seeking is faster but less accurate: it may seek to a position slightly before or
```
dvdsk: > /// Coarse seeking is faster but less accurate: it may seek to a position slightly before or after the requested one, especially when the bitrate is variable.

was this not true? I like to have it documented here if it is true

dvdsk (approved):
splendid

You removed the explanation of what coarse seek, is it not defined by symphonia?

It does not matter, this PR is done and a wonderful improvement. And thanks for waiting with the rebase, it was very easy to review like this..

src/decoder/builder.rs
roderickvd: It is but it's duplicate with the documentation of `fn with_coarse_seek`. I chose to remove the duplicate parts, and instead mention here that Rodio's accurate seeking means "sample-accurate".

dvdsk: oh its in `fn with_coarse_seek`, that a better place for it :+1:

roderickvd:
Rebasing done, CI passes, good to merge.

nc7s:
Thanks for implementing this! I can now save adding symphonia and doing the calculations myself. Would you mind kindly releasing it?

roderickvd:
Yes, see #721

PR #699, opened 2025-02-10
Fix WavDecoder::total_duration.

IohannRabeson:
It was ignoring the channels count.

### DIFF
--- a/src/decoder/wav.rs
+++ b/src/decoder/wav.rs
@@ -44,9 +44,9 @@ where
         let channels = spec.channels;
 
         let total_duration = {
-            let sample_rate = sample_rate as u64;
-            let secs = len / sample_rate;
-            let nanos = ((len % sample_rate) * 1_000_000_000) / sample_rate;
+            let data_rate = sample_rate as u64 * channels as u64;
+            let secs = len / data_rate;
+            let nanos = ((len % data_rate) * 1_000_000_000) / data_rate;
             Duration::new(secs, nanos as u32)
         };
 


IohannRabeson:
Ok, I'm checking and I will fix it.

src/decoder/wav.rs
```diff
@@ -44,7 +44,7 @@ where
         let channels = spec.channels;
 
         let total_duration = {
-            let sample_rate = sample_rate as u64;
```
dvdsk, line 47: good catch!

It might not be best to have `sample_rate = sample_rate *channels` since that is not really the sample rate anymore. Maybe we can call it data_rate? 

Alternatively I kinda like  `len / sample_rate / channels` instead of `len / sample_rate` though the nano's line would get more complex then. We could rewrite that as: `nanos = (len - secs) * 1_000_000_000 / sample_rate / channels`.

I'm leaning to just renaming it to data_rate.

dvdsk:
> D'oh! Pointy hat to me, good one. I probably copied the same mistake in the `lewton` FLAC decoder too:

you've got this (very nice) comment there that points to the opposite:
```rust
        // `samples` in FLAC means "inter-channel samples" aka frames
        // so we do not divide by `self.channels` here.
```

src/decoder/wav.rs
roderickvd: > I'm leaning to just renaming it to data_rate.

Me too. It saves the hassle of preventing overflowing `nanos` again.

roderickvd:
> > D'oh! Pointy hat to me, good one. I probably copied the same mistake in the `lewton` FLAC decoder too:
> 
> you've got this (very nice) comment there that points to the opposite:
> 
> ```rust
>         // `samples` in FLAC means "inter-channel samples" aka frames
>         // so we do not divide by `self.channels` here.
> ```

🤦‍♂️ I will go to bed now 😄

dvdsk:
I'm gonna go and add some integration tests to prevent this from ever regressing. Looks good to go for me

dvdsk:
> 🤦‍♂️ I will go to bed now 😄

I should follow your example, I just spend some time writing total duration tests before realizing I already did that... three days ago (to be merged as part of #697).

PR #700, opened 2025-02-11
Use f32 samples

PetrGlad:
WIP. Will work on it again on weekend, so keeping it here as draft for now.

Remaining part is to think through how sample formats should be converted. 
There is a sample converter that works on any iterator. Maybe we can have a legacy source trait that will be parametrized and with span length getter. This can be used to ease migration to the new API. Do we need sample format adapters for decoders? 
To see how the API should look like some use cases for sample format conversion are welcome. 

Output callback converts samples anyway so there is no need to adapt that explicitly. Output stream sample format only determines the samples that are sent to ALSA. Likely the output sample format should default to f32 in the output stream builder.

Our default resampler can now be simplified (see `lerp` function).

Some benches on my notebook degraded ~5%, others stayed on par with older implementation.


dvdsk:
First off all, thanks for the work it is huge.

> Some benches on my notebook degraded ~5%

strange, that's not what I estimated for arm. I had a quick run on my system (ryzen 5600 (x64)). I have no idea whats happening. 

- Amplify is way slower (on the other hand 100 microseconds extra is not really an issue). Since it takes a float already that makes no sense to me.

data:
```
effects            fastest       │ slowest       │ median        │ mean          │ samples │ iters
├─ amplify curr    380.1 µs      │ 430.7 µs      │ 382.8 µs      │ 389.4 µs      │ 100     │ 100
├─ amplify pr      569.9 µs      │ 1.248 ms      │ 572.6 µs      │ 578.5 µs      │ 100     │ 100

```

- The other effects stay the same or very slightly speed up. This makes sense they need floats so a conversion from floats is now gone.

- Resampler speeds up dramatically for large sample_rates, I guess it might autovectorise more? I have not looked at the generated assembly 

data (left out some to ease compare):
```
resampler         fastest       │ slowest       │ median        │ mean          │ samples │ iters
╰─ current                      │               │               │               │         │
   ├─ 8000        2.39 ms       │ 3.729 ms      │ 2.42 ms       │ 2.447 ms      │ 100     │ 100
   ├─ 22050       4.05 ms       │ 4.162 ms      │ 4.056 ms      │ 4.06 ms       │ 100     │ 100
   ├─ 44100       2.666 ms      │ 2.738 ms      │ 2.669 ms      │ 2.67 ms       │ 100     │ 100
   ├─ 96000       12.39 ms      │ 13.4 ms       │ 12.64 ms      │ 12.65 ms      │ 100     │ 100
   ├─ 192000      23.77 ms      │ 26.06 ms      │ 24 ms         │ 24.03 ms      │ 100     │ 100
   ╰─ 384000      47.67 ms      │ 58.18 ms      │ 48.13 ms      │ 48.25 ms      │ 100     │ 100
╰─ new                          │               │               │               │         │
   ├─ 8000        2.424 ms      │ 2.732 ms      │ 2.438 ms      │ 2.443 ms      │ 100     │ 100
   ├─ 22050       4.04 ms       │ 4.379 ms      │ 4.068 ms      │ 4.072 ms      │ 100     │ 100
   ├─ 44100       2.767 ms      │ 2.882 ms      │ 2.796 ms      │ 2.796 ms      │ 100     │ 100
   ├─ 96000       11.75 ms      │ 12.18 ms      │ 11.81 ms      │ 11.81 ms      │ 100     │ 100
   ├─ 192000      22.19 ms      │ 22.44 ms      │ 22.22 ms      │ 22.23 ms      │ 100     │ 100
   ╰─ 384000      44.75 ms      │ 46.16 ms      │ 44.79 ms      │ 44.81 ms      │ 100     │ 100
```

I would like to know why amplify is slowing down. Overall performance is good enough for me.

dvdsk:
extra context: this is what amplify is now (inlined a dasp_call):

```rust
    /// Multiplies the value of this sample by the given amount.
    #[inline]
    fn amplify(self, value: f32) -> Self {
        {
            let amp = value.to_sample();
            let self_f = self.to_float_sample();
            (self_f * amp).to_sample()
        }
    }

```

PetrGlad:
`dasp`'s amplify is not necessary anymore. I have to see where that is used. If we have `f32` every time, then just `*` would do instead.

PetrGlad:
@dvdsk How many time have you ran the benches? Microbenchmarks are always noisy. Was it in release mode?

The new `amplify` cannot be simpler:
```rust
#[inline]
        self.input.next().map(|value| value * self.factor)
    }
```
If this one is indeed slower , only reason I could think of is less efficient cache usage since samples now take more space.

dvdsk:
> @dvdsk How many times have you ran the benches? Microbenchmarks are always noisy. Was it in release mode?

we run them with `cargo bench`, that should always build a release build. Divan runs each test 100 times. I have warmed up the cache by running the entire suite a few times alternating between this pr and main branch. I observed some fluctuation in those runs, about a percent. But nothing on the order of the *amplify* change.

dvdsk:
could you try it on your machine? (just clone the repo somewhere else so you can run them easily one after another). Just to make sure nothing really wierd is going on.

PetrGlad:
BTW, I normally only look at the slowest and average since slowest is important for real-time and average estimates sustained rate.
I ran the tests again, variation between tests on my notebook is significant, maybe something else interferes but there are 16 CPU cores that are almost all idling...

Amplify average in old and new implementation is within 1% (new one is faster). Max value may be worse but it varies too much however value ranges are about the same.
Have you tested both implementations in the same process? How did you do it? By copying other implementation?

Rate conversion is slower by about 12% on average for 384 kHz. Best old (master) vs. worst new (f32)  is about 18%. I've looked at the resampler bench, it actually uses  `UniformSourceIterator` so maybe something else is getting in the way.

roderickvd:
Just to rule it out, what if moving `lerp` into a separate module prevents certain compile-time optimizations? Does it matter whether you compile with `codegen-units = 1` or not?

src/stream.rs
```diff
@@ -50,7 +50,7 @@ impl Default for OutputStreamConfig {
             channel_count: 2,
             sample_rate: HZ_44100,
             buffer_size: BufferSize::Default,
-            sample_format: SampleFormat::I8,
```
dvdsk, line 53: why not f32? If its supported by the OS that will probably have the lowest overhead.

src/source/noise.rs
```diff
@@ -42,7 +42,7 @@ impl WhiteNoise {
     pub fn new(sample_rate: SampleRate) -> Self {
         Self {
             sample_rate,
-            rng: SmallRng::from_entropy(),
```
dvdsk, line 45: what is the difference between these? Would `seed_from_u64(0)` not be better? then the noise is always the same.

src/stream.rs
PetrGlad: Just forgot to update. But maybe I16 can be a better default since it is widely supported.

PetrGlad:
Yes, inlined `lerp` is a tad faster (by ~4%).

dvdsk:
might be out of scope of this PR.

Currently we open streams with the default output config for that device. Depending on the method used we fall back to other output configs. Should we not instead see if one of the output configs supported by the device has f32 as sample format and then pick that output format? If there are multiple we could look for the f32 sample format config that is closest to the default stream format.

PetrGlad:
After some pondering I decided to remove `SamplesConverter` existing sources can be adapted just by adding `DaspSample::from_sample(s)` or
```rust
inner.next().map(|s| DaspSample::from_sample(s))
```
to the `next()` method. So using this as an adapter will not help much.

src/decoder/mod.rs
```diff
@@ -31,12 +31,8 @@ mod vorbis;
 #[cfg(all(feature = "wav", not(feature = "symphonia-wav")))]
 mod wav;
 
-#[cfg(feature = "integer-decoder")]
 /// Output format of the decoders.
 pub type DecoderSample = Sample;
```
dvdsk, line 35: we can inline this right? No worries I'll take care of it post merge.

src/stream.rs
```diff
@@ -55,14 +55,18 @@ pub struct OutputStreamBuilder {
     config: OutputStreamConfig,
 }
 
-impl Default for OutputStreamConfig {
```
dvdsk: I do not know why the `Default` impl is gone. You have refactored that code so you know what you must know what your doing.

src/stream.rs
```diff
@@ -55,14 +55,18 @@ pub struct OutputStreamBuilder {
...
-    fn default() -> Self {
-        Self {
-            channel_count: 2,
-            sample_rate: HZ_44100,
-            buffer_size: BufferSize::Default,
-            sample_format: SampleFormat::I16,
-        }
+impl core::fmt::Debug for OutputStreamBuilder {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
```
dvdsk, line 59: I do not know why these changes are new here? they come from main. Its not a problem though I do not mind.

dvdsk (approved):
good to merge! Thank you very much for this huge PR. Rodio is 500 lines smaller now and that is a very good thing. The code and examples have also become simpler :partying_face:

src/decoder/mod.rs
PetrGlad: Yes, this can be inlined. I thought it might still be useful.

PR #703, opened 2025-02-17
Do not use cpal errors when rodio runs into an issue

dvdsk:
This prevents a search for BuildStreamError::StreamConfigNotSupported pointing to rodio code. That puts you on the wrong leg making you think the issue is in rodio where it is in Cpal. This was the only case in rodio where we issue a cpal error.

### DIFF
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -289,6 +289,9 @@ pub enum StreamError {
     SupportedStreamConfigsError(cpal::SupportedStreamConfigsError),
     /// Could not find any output device
     NoDevice,
+    /// New cpal sample format that rodio does not yet support please open
+    /// an issue if you run into this.
+    UnknownSampleFormat,
 }
 
 impl fmt::Display for StreamError {
@@ -299,6 +302,7 @@ impl fmt::Display for StreamError {
             Self::DefaultStreamConfigError(e) => e.fmt(f),
             Self::SupportedStreamConfigsError(e) => e.fmt(f),
             Self::NoDevice => write!(f, "NoDevice"),
+            Self::UnknownSampleFormat => write!(f, "UnknownSampleFormat"),
         }
     }
 }
@@ -311,6 +315,7 @@ impl error::Error for StreamError {
             Self::DefaultStreamConfigError(e) => Some(e),
             Self::SupportedStreamConfigsError(e) => Some(e),
             Self::NoDevice => None,
+            Self::UnknownSampleFormat => None,
         }
     }
 }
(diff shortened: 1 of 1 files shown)


src/stream.rs
```diff
@@ -452,8 +455,9 @@ impl OutputStream {
                 error_callback,
                 None,
             ),
-            _ => Err(cpal::BuildStreamError::StreamConfigNotSupported),
+            _ => return Err(StreamError::UnknownSampleFormat),
```
PetrGlad, line 458: This can only happen with incompatible CPAL version. I wonder why they decided to mark cpal sample format enum as `non_exhaustive`. This could probably panic instead.

PetrGlad (approved):
I should have submitted my patch yesterday :)

dvdsk:
> I should have submitted my patch yesterday :)

I thought about pushing it to main directly :sweat_smile: you should have done that :stuck_out_tongue:

src/stream.rs
```diff
@@ -299,6 +302,7 @@ impl fmt::Display for StreamError {
             Self::DefaultStreamConfigError(e) => e.fmt(f),
             Self::SupportedStreamConfigsError(e) => e.fmt(f),
             Self::NoDevice => write!(f, "NoDevice"),
+            Self::UnknownSampleFormat => write!(f, "UnknownSampleFormat"),
```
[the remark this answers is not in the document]
dvdsk: UnexpectedSampleFormat? Since its pretty strange if it happens.

src/stream.rs
dvdsk: yeah had the same thought. But decided they must have done it for a reason. One of their examples (beep) has some commented out sample formats, so they might add those in the future (24bit for example)

src/stream.rs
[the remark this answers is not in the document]
PetrGlad: It sounds like we expected something else instead... "Unknown" in this situation is only unknown to `rodio` but not `cpal` :) I like "unsupported" the best here but "unexpected" would also do.

dvdsk:
should be good, if it is please merge it :+1:

src/stream.rs
PetrGlad: The reason I do not understand this is that cargo compiles both `rodio` and `cpal` from sources every time, so it can check that all match branches are there. Maybe there are some scenarios where pre-compiled Rust libraries are used, but I have not seen such projects so far.

dvdsk: I recommend reading the RFC for `non_exhaustive`, it goes through all the reasons for using it. https://rust-lang.github.io/rfcs/2008-non-exhaustive.html

dvdsk: and cpal is adding 24bit support to this enum it seems. Its non breaking because they have non_exhaustive on the enum. https://github.com/RustAudio/cpal/pull/927

PR #704, opened 2025-02-17
Better diagnostics when CPAL reports 0 buffer size in device config.

PetrGlad:


src/stream.rs
```diff
@@ -1,14 +1,13 @@
-use std::io::{Read, Seek};
```
dvdsk, line 1: I see the imports have been sorted and newlines removed. 

I always keep imports separated into three sections each with a empty line in between. std imports at the top, then those from other libs finally crate local imports at the bottom. 

Its something I have copied from examples and other projects such as Tokio. Though the order seems to change somewhat from project to project. 

We do not have that in the rodio contributors guide atm. I would be for adding it and keeping the current import formatting. What is your preference here?

dvdsk (commented):
I'm not sure if this is something we should address in rodio. To me it seems this is a clear cpal problem, cpal is returning an error with input provided by... cpal. There is no reason why a zero sized buffer is impossible. Rodio does not need nonzero buffers. Maybe cpal changes in the future and zero sized buffers become possible, maybe that is the case now and the bug is not the zero sized buffer but cpal returning an error. 

Lets wait till we hear from the cpal devs then decide what to do.

src/stream.rs
PetrGlad: This change by `cargo fmt`.

PetrGlad:
@dvdsk That is exactly what assertions are intended for. If we expect non zero from cpal and cannot continue otherwise we should put an assertion in.

dvdsk:
> If we expect non zero from cpal and cannot continue otherwise

isn't the only reason we can not continue otherwise that cpal can not deal with non zero?

PetrGlad:
Even if cpal always can process configs it produces, stream builder gives an option for users to alter configs.

dvdsk:
> Even if cpal always can process configs it produces, stream builder gives an option for users to alter configs.

that is a really good argument

PR #708, opened 2025-02-27
Add with_error_callback to OutputStreamBuilder

### LINKED ISSUE #465: Allow user to get to cpal errors such as device unplugged.
Hi, I'm writing a tui and am using rodio for audio playback.

I'm not sure how to handle `an error occurred on output stream: The requested device is no longer available. For example, it has been unplugged`. This happens when I unplug my earphones which makes sense. I want to retry the default device but I'm not sure how to detect this error because it comes from `eprintln!`. 

Also, I'd like there to be no prints in the first place since it screws up my terminal output. I would think errors are routed through the `log` crate for library users to handle. 

What am I missing on both fronts?

will3942:
Allows a user to handle a cpal error, allowing them to optionally recreate the stream if desired or take another action (such as relaunching the app).

I wasn't able to do this without using `Box` - happy to take any suggestions if you can think of a better implementation as I only have very nascent Rust knowledge.

Closes #465


dvdsk:
> I wasn't able to do this without using `Box` - happy to take any suggestions if you can think of a better implementation as I only have very nascent Rust knowledge.

make the builder generic over the error callback (you will need to use `OutputStreamBuilder<ErrorCallback>` instead of `Self<ErrorCallback`  in `with_error_callback`). There should be no need to make OutputStream generic.

will3942:
> > I wasn't able to do this without using `Box` - happy to take any suggestions if you can think of a better implementation as I only have very nascent Rust knowledge.
> 
> 
> 
> make the builder generic over the error callback (you will need to use `OutputStreamBuilder<ErrorCallback>` instead of `Self<ErrorCallback`  in `with_error_callback`). There should be no need to make OutputStream generic. 

I tried that for a while but was unable to get it working with a fallback to the default error callback but will try again next week.

PetrGlad:
I think a dyn Fn makes sense here, the callback is not performance-critical. 

The problem with opening default stream is that it may try to open stream multiple times. At the same time the call back gets owned at each attempt. See also my `comment in open_stream_or_fallback`.

src/stream.rs
```diff
@@ -136,19 +147,30 @@ impl OutputStreamBuilder {
...
     /// supported by the device.
     /// If all attempts fail returns initial error.
-    pub fn open_stream_or_fallback(&self) -> Result<OutputStream, StreamError> {
+    pub fn open_stream_or_fallback(self) -> Result<OutputStream, StreamError> {
         let device = self.device.as_ref().expect("output device specified");
-        OutputStream::open(device, &self.config).or_else(|err| {
+
+        OutputStream::open(device, &self.config, self.error_callback).or_else(|err| {
             for supported_config in supported_output_configs(device)? {
                 if let Ok(handle) = Self::default()
```
PetrGlad, line 175: Default callback is used here if the first attempt fails... So custom callback will not be used in subsequent attempts...

dvdsk:
> I think a dyn Fn makes sense here, the callback is not performance-critical.

I worry about the underflow error (when the source can not deliver samples fast enough to the OS). The a slow error handler could make the problem worse, a single underflow error could cause many more. I do not know if the delay caused by dyn is severe enough to cause that.

will3942:
> > I think a dyn Fn makes sense here, the callback is not performance-critical.
> 
> I worry about the underflow error (when the source can not deliver samples fast enough to the OS). The a slow error handler could make the problem worse, a single underflow error could cause many more. I do not know if the delay caused by dyn is severe enough to cause that.

@dvdsk I have now implemented this using generics instead without dyn dispatch. Let me know your feedback.

src/stream.rs
will3942: Thanks - good catch, I've resolved this now!

examples/error_callback.rs
```diff
@@ -0,0 +1,34 @@
...
+
+fn main() -> Result<(), Box<dyn Error>> {
+    // You can use any other output device that can be queried from CPAL.
+    let default_device = cpal::default_host()
+        .default_output_device()
+        .ok_or("No default audio output device is found.")?;
+
+    let stream_handle = rodio::OutputStreamBuilder::from_device(default_device)?
+        .with_error_callback(|err| {
+            // Filter for where err is a DeviceNotAvailable error.
```
dvdsk, line 16: most users will want to use the callback with an mpsc::Sender right? I might make sense to have a little more complex example that illustrates that.

Maybe the callback could send `cpal::StreamError::DeviceNotAvailable` into a std::mpsc and print all the others?

will3942:
> Looks great, if you want to you could make the example a little more in-depth (see comment) but this is good to merge.

All done - feel free to merge once you're happy!

dvdsk (approved):

dvdsk:
O.o. CI is broken it seems....

well I'll see if I can fix that in a bit. PR looks great, ill try to get CI sorted and then merge this.

will3942:
> O.o. CI is broken it seems....
> 
> well I'll see if I can fix that in a bit. PR looks great, ill try to get CI sorted and then merge this.

CI is all fixed - I broke it by forgetting a `cargo fmt`. Thanks for your help on this PR!

PetrGlad:
I wonder if anyone can simulate/test underflow error. I could not. On my system no matter how slow the source is pipewire/ALSA does not complain (at least not through CPAL).

PetrGlad:
On my system (Linux+ALSA+Pipewire). When I stop audio server I get `BackendSpecificError` instead.
Generally, I do not believe there are any recoverable errors that we get in the callback.

PetrGlad:
@will3942 Are there any recoverable errors you get from this callback?

will3942:
> @will3942 Are there any recoverable errors you get from this callback?

How do you define recoverable? Recover without having to tear down and destroy the resources?

roderickvd:
> I wonder if anyone can simulate/test underflow error. I could not. On my system no matter how slow the source is pipewire/ALSA does not complain (at least not through CPAL).

One way to do it is to use an RPi with SD card and do a lot of I/O on it, writes in particular.

PetrGlad:
@will3942 Yes, by "recoverable" here I mean being able to proceed without re-initializing the stream.

PetrGlad:
@roderickvd The output stream can not know where delay comes from. So I believe that just artificially stalling the sample callback should have the same effect.

will3942:
> @will3942 Yes, by "recoverable" here I mean being able to proceed without re-initializing the stream.

Not that I'm aware of, no.

In my application code I will be using this to remove and recreate the OutputStream, checking if the device is still available. It would be nice if Rodio could optionally provide this functionality out of the box but it feels like there's quite a few toggles/options/timeouts you'd want to provide the user in determining how that logic is provided.

PR #709, opened 2025-02-28
Make ChannelCount and SampleRate NonZero

dvdsk:
I ran into a lot of bugs while adding tests that had to do with channel being set to zero somewhere. While this change makes the API slightly less easy to use it prevents very hard to debug crashes/underflows etc.

Similar to the channels situation a lot of bugs seem to be caused by zero sample_rate. Since audio can literally not play at zero speed it makes no sense to have zero sample_rate's.

Neither channel = 0 nor sample_rate = 0 are used as a sentinel value anywhere

Performance might drop in decoders, the current implementation makes the bound check every time `channels` or `sample_rate` is called which should be about once per span. This could be cached to alleviate that.

### DIFF
--- a/benches/resampler.rs
+++ b/benches/resampler.rs
@@ -4,7 +4,7 @@ use rodio::source::UniformSourceIterator;
 mod shared;
 use shared::music_wav;
 
-use rodio::Source;
+use rodio::{SampleRate, Source};
 
 fn main() {
     divan::main();
@@ -31,6 +31,7 @@ const COMMON_SAMPLE_RATES: [u32; 12] = [
 
 #[divan::bench(args = COMMON_SAMPLE_RATES)]
 fn resample_to(bencher: Bencher, target_sample_rate: u32) {
+    let target_sample_rate = SampleRate::new(target_sample_rate).unwrap();
     bencher
         .with_inputs(|| {
             let source = music_wav();
--- a/benches/resampler.rs
+++ b/benches/resampler.rs
@@ -31,7 +31,7 @@ const COMMON_SAMPLE_RATES: [u32; 12] = [
 
 #[divan::bench(args = COMMON_SAMPLE_RATES)]
 fn resample_to(bencher: Bencher, target_sample_rate: u32) {
-    let target_sample_rate = SampleRate::new(target_sample_rate).unwrap();
+    let target_sample_rate = SampleRate::new(target_sample_rate).expect("Is not zero");
     bencher
         .with_inputs(|| {
             let source = music_wav();
--- a/src/conversions/sample_rate.rs
+++ b/src/conversions/sample_rate.rs
@@ -54,7 +54,6 @@ where
         to: SampleRate,
         num_channels: ChannelCount,
     ) -> SampleRateConverter<I> {
-        assert!(num_channels >= 1);
         assert!(from >= 1);
         assert!(to >= 1);
 
@@ -64,11 +63,11 @@ where
         } else {
             let first = input
                 .by_ref()
-                .take(num_channels as usize)
+                .take(num_channels.get() as usize)
                 .collect::<Vec<_>>();
             let next = input
                 .by_ref()
-                .take(num_channels as usize)
+                .take(num_channels.get() as usize)
                 .collect::<Vec<_>>();
             (first, next)
         };
@@ -85,7 +84,7 @@ where
             next_output_span_pos_in_chunk: 0,
             current_span: first_samples,
             next_frame: next_samples,
-            output_buffer: Vec::with_capacity(num_channels as usize - 1),
+            output_buffer: Vec::with_capacity(num_channels.get() as usize - 1),
         }
     }
 
@@ -106,7 +105,7 @@ where
 
         mem::swap(&mut self.current_span, &mut self.next_frame);
         self.next_frame.clear();
-        for _ in 0..self.channels {
+        for _ in 0..self.channels.get() {
             if let Some(i) = self.input.next() {
                 self.next_frame.push(i);
             } else {
@@ -213,7 +212,7 @@ where
             // removing the samples of the current chunk that have not yet been read
             let samples_after_chunk = samples_after_chunk.saturating_sub(
                 self.from.saturating_sub(self.current_span_pos_in_chunk + 2) as usize
-                    * usize::from(self.channels),
+                    * usize::from(self.channels.get()),
             );
             // calculating the number of samples after the transformation
             // TODO: this is wrong here \|/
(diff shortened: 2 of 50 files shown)


PetrGlad:
Yes, maybe. Have you tried to bench it?

src/math.rs
```diff
@@ -10,6 +10,17 @@ pub fn lerp(first: &f32, second: &f32, numerator: u32, denominator: u32) -> f32
     first + (second - first) * numerator as f32 / denominator as f32
 }
 
+/// short macro to generate a `NonZero`. It panics during compile if the
+/// passed in literal is zero. Used for `ChannelCount` and `Samplerate`
+/// constants
+macro_rules! nz {
```
roderickvd, line 16: A similar macro `nonzero` is provided by [nonzero_ext](https://crates.io/crates/nonzero_ext). I understand it may be too heavy-weight to pull in a dependency just for that, but maybe we could use the same name for recognizability and ease of switching if we ever want to use more of `nonzero_ext` in the future.

dvdsk: I had a look its not a transitive dependencies of rodio so for now lets keep it out. (Another argument is that `nz!` is nice in short, though that might be bad taste on my end). 

If you ever need something from nonzero_ext do not hesitate to include it though!

yara-blue:
Unwrap is now const, and Nonzero:new also. That should replace this PR.

yara-blue:
Though most of this PR is still quite valuable so lets not delete the branch.

src/source/noise.rs
yara-blue, line 1: fun thing, if you checkout this PR locally this is how it should be (zero changes).... wth github?

yara-blue:
this PR is utterly borked github's diff view does not match what I have locally. I am reopening it in the hope it gets fixed.

yara-blue:
new PR did not work....

tests/seek.rs
yara-blue, line 1: another rebase issue

yara-blue:
Well that was messy :(

I would like to merge this tomorrow. I've carefully gone through the full diff and it looks good now. Performance issues (I do not expect any since branches should be on cold paths) can easily be addressed later. If we leave this though other PR's will land and we will need another rebase/merge conflict.

benches/resampler.rs
```diff
@@ -31,6 +31,7 @@ const COMMON_SAMPLE_RATES: [u32; 12] = [
 
 #[divan::bench(args = COMMON_SAMPLE_RATES)]
 fn resample_to(bencher: Bencher, target_sample_rate: u32) {
+    let target_sample_rate = SampleRate::new(target_sample_rate).unwrap();
```
roderickvd: Here an `expect` may be nice. Never know how this is going to be called.

src/conversions/sample_rate.rs
```diff
@@ -242,115 +238,110 @@ impl<I> ExactSizeIterator for SampleRateConverter<I> where I: ExactSizeIterator<
...
-                let ns = channels as usize;
+                let ns = channels.get() as usize;
                 let mut i = input;
                 i.truncate(ns * (i.len() / ns));
                 i
             };
 
             let output =
-                SampleRateConverter::new(input.clone().into_iter(), from, to, channels as ChannelCount)
+                SampleRateConverter::new(input.clone().into_iter(), SampleRate::new(from).unwrap(), to, channels)
```
roderickvd: I was seeing if this could use `expect` to, because `k` is not guaranteed to be non-zero, when my IDE tells me that this function isn't called anywhere?

src/source/channel_volume.rs
```diff
@@ -24,10 +24,10 @@ where
...
     /// channel at the volume set by the user. The volume can be changed using
     /// [`ChannelVolume::set_volume`].
-    pub fn new(input: I, channel_volumes: Vec<f32>) -> ChannelVolume<I>
-    where
-        I: Source,
-    {
+    ///
+    /// # Panics if channel_volumes is empty
+    pub fn new(input: I, channel_volumes: Vec<f32>) -> ChannelVolume<I> {
+        assert!(!channel_volumes.is_empty());
```
roderickvd: We could change the iterator to do nothing in this case? (use `if let Some(channel_volume) = self.channel_volumes.get(i)` instead of indexing)

src/source/chirp.rs
```diff
@@ -37,7 +38,7 @@ impl Chirp {
             sample_rate,
             start_frequency,
             end_frequency,
-            total_samples: (duration.as_secs_f64() * (sample_rate as f64)) as u64,
+            total_samples: (duration.as_secs_f64() * (sample_rate.get() as f64)) as u64,
```
roderickvd, line 41: I was thinking that this could use a `.ceil()` call, but looking at the rest of the code it may even be better if we stored this as a floating point in the struct.

src/source/periodic.rs
```diff
@@ -12,7 +12,8 @@ where
     // TODO: handle the fact that the samples rate can change
     // TODO: generally, just wrong
     let update_ms = period.as_secs() as u32 * 1_000 + period.subsec_millis();
-    let update_frequency = (update_ms * source.sample_rate()) / 1000 * source.channels() as u32;
+    let update_frequency =
+        (update_ms * source.sample_rate().get()) / 1000 * source.channels().get() as u32;
```
roderickvd, line 16: For frequencies like 44100 Hz, integer division by 1000 will cause rounding errors. We should probably do the calculation in floating point first.

src/conversions/sample_rate.rs
yara-blue: Github is also not showing a line number..... strange
src/conversions/sample_rate.rs+295 btw

yara-blue: I've put in an expect explaining k and to are both nonzero

src/source/channel_volume.rs
yara-blue: I rather panic, passing in zero channel volumes seems like an obvious mistake. I've removed it from this PR since it has nothing to do with nonzero.

src/source/chirp.rs
yara-blue: Is that a separate issue (from the move to nonzero)?

src/source/periodic.rs
yara-blue: Also separate issue from the move to nonzero right?

(I fully agree btw, feel free to fix this directly on main)

yara-blue:
> That must've been quite a chore. Reviewing it, I came across a couple of points that are mostly not specific to this change, but stood out.

Thanks for the fast and thorough review! I think there are two unrelated (but important) issues that should be addressed somewhere else.

> We could entertain changing SampleRate and ChannelCount upstream in cpal to be NonZero, too.

Good idea! Seems like cpal is still actively maintained (yay :)) Go for it!

benches/resampler.rs
yara-blue: done

yara-blue:
This should be perfect now so I'm going to merge this. If any of you disagree feel free to revert it.

PR #711, opened 2025-03-01
Move `Arc` into `Mixer`

casey:


dvdsk:
yeah this seems slightly easier to use. I like that the `Arc` is now out of sight for the end user. I'm surprised the changelog does not need editing. But that seems to check out.

dvdsk:
oh and thanks :+1: 

do not hesitate to let us know if you think of anything else!

casey:
You bet! Thanks for reviewing and merging!

PetrGlad:
I am not sure this is an implementation detail that had to be hidden. Now one have to explain that Mixer has to be cloned if one wants to pass it around. IMO Arc made that apparent.

dvdsk:
It seems I might have merged to soon, I did not think anyone would disagree. Apologies for not asking/waiting for your input Petr, I did not want to bother you. Now to the matter at hand:

> Now one have to explain that Mixer has to be cloned if one wants to pass it around. IMO Arc made that apparent.

It does not need to be *cloned* right? You can just move it. Yes if you need two instances you need to clone it, but that is the case for `String` too. 

Or do you mean the `Arc` signals that those two instances will point to the same thing, and now we need to describe that to the user? 

In that case I understand your point. I am not sure though if Arc is a way to signal that. There are plenty of API's in the rust standard library that behave similarly right? Like `mpsc::Sender` for example. It is clone but it will keep sending to the same receiver. That is analogous to our mixer 'handle' mixing into the same source. Maybe a rename could help here, `MixerInput` comes to mind as an option.

Issue #714, opened 2025-03-06
Better volume control


dvdsk:
I noticed volume control in decibels in bevy. I think it would be great to adopt into rodio (with proper credit). It would augment the current `amplify`.

See: https://github.com/bevyengine/bevy/blob/ed7b366b24d7106f4715369eb3516875f915f536/crates/bevy_audio/src/volume.rs#L1

roderickvd:
When at it, add a logarithmic mapping from `[0.0; 1.0]` over a range of 60 dB: https://www.dr-lex.be/info-stuff/volumecontrols.html

dvdsk:
Reading that source it seems a little more complex then a simple logarithm. But it is nicely explained so this should still be quite easy to get implemented.

dvdsk:
> No more than 10 lines: https://github.com/roderickvd/pleezer/blob/26599fcc63f97f54681e496ecb7fe8cd5dcb3214/src/player.rs#L1277

That is actually very nice! If you want to and have the time please contribute it.

dvdsk:
though I'm not sure in what form. I think we need to keep the current amplify. Maybe `amplify_log`? We should definitely point to it from the `amplify` docs since this is probably what users will want.

PetrGlad:
Maybe just add conversions for now? Using to_decibels() , to_linear(): `amplify(to_linear(-1.2))`?
By the way is 100% volume <-> 0.0dB in Bevy?

roderickvd:
> Maybe just add conversions for now? Using to_decibels() , to_linear(): `amplify(to_linear(-1.2))`?

I like it.

daalfox:
Can I give this a try?

dvdsk:
> Can I give this a try?

go ahead, have fun!

dvdsk:
implemented in #715

PR #715, opened 2025-03-14
feat: Allow logarithmic amplification (needs input on API)

daalfox:
Adds volume control in decibels discussed in #714

### DIFF
--- a/src/source/amplify.rs
+++ b/src/source/amplify.rs
@@ -13,7 +13,7 @@ where
 }
 
 /// Internal function that converts decibels to linear
-pub fn to_linear(decibels: f32) -> f32 {
+pub(super) fn to_linear(decibels: f32) -> f32 {
     f32::powf(10f32, decibels * 0.05)
 }
 
@@ -147,7 +147,7 @@ mod test {
     fn convert_decibels_to_linear() {
         for (db, linear) in DECIBELS_LINEAR_TABLE {
             dbg!(db, linear, f32::abs(to_linear(db) - linear));
-            assert!(f32::abs(to_linear(db) - linear) < EPSILON,)
+            assert!(f32::abs(to_linear(db) - linear) < EPSILON)
         }
     }
 }
--- a/src/source/amplify.rs
+++ b/src/source/amplify.rs
@@ -31,6 +31,12 @@ impl<I> Amplify<I> {
         self.factor = factor;
     }
 
+    /// Modifies the amplification factor logarithmically.
+    #[inline]
+    pub fn set_log_factor(&mut self, factor: f32) {
+        self.factor = to_linear(factor);
+    }
+
     /// Returns a reference to the inner source.
     #[inline]
     pub fn inner(&self) -> &I {
--- a/src/source/amplify.rs
+++ b/src/source/amplify.rs
@@ -12,6 +12,11 @@ where
     Amplify { input, factor }
 }
 
+/// converts decibels to linear
+pub fn to_linear(decibels: f32) -> f32 {
+    f32::powf(10f32, decibels * 0.05)
+}
+
 /// Filter that modifies each sample by a given value.
 #[derive(Clone, Debug)]
 pub struct Amplify<I> {
@@ -91,5 +96,52 @@ where
     #[inline]
     fn try_seek(&mut self, pos: Duration) -> Result<(), SeekError> {
         self.input.try_seek(pos)
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    const EPSILON: f32 = 0.3;
+    /// Based on [Wikipedia's Decibel article].
+    ///
+    /// [Wikipedia's Decibel article]: https://web.archive.org/web/20230810185300/https://en.wikipedia.org/wiki/Decibel
+    const DECIBELS_LINEAR_TABLE: [(f32, f32); 27] = [
+        (100., 100000.),
+        (90., 31623.),
+        (80., 10000.),
+        (70., 3162.),
+        (60., 1000.),
+        (50., 316.2),
+        (40., 100.),
+        (30., 31.62),
+        (20., 10.),
+        (10., 3.162),
+        (5.998, 1.995),
+        (3.003, 1.413),
+        (1.002, 1.122),
+        (0., 1.),
+        (-1.002, 0.891),
+        (-3.003, 0.708),
+        (-5.998, 0.501),
+        (-10., 0.3162),
+        (-20., 0.1),
+        (-30., 0.03162),
+        (-40., 0.01),
+        (-50., 0.003162),
+        (-60., 0.001),
+        (-70., 0.0003162),
+        (-80., 0.0001),
+        (-90., 0.00003162),
+        (-100., 0.00001),
+    ];
+
+    #[test]
+    fn convert_decibels_to_linear() {
+        for (db, linear) in DECIBELS_LINEAR_TABLE {
+            dbg!(db, linear, f32::abs(to_linear(db) - linear));
+            assert!(f32::abs(to_linear(db) - linear) < EPSILON,)
+        }
     }
 }
--- a/src/source/amplify.rs
+++ b/src/source/amplify.rs
@@ -108,8 +108,6 @@ where
 #[cfg(test)]
 mod test {
     use super::*;
-
-    const EPSILON: f32 = 0.3;
     /// Based on [Wikipedia's Decibel article].
     ///
     /// [Wikipedia's Decibel article]: https://web.archive.org/web/20230810185300/https://en.wikipedia.org/wiki/Decibel
@@ -146,8 +144,13 @@ mod test {
     #[test]
     fn convert_decibels_to_linear() {
         for (db, linear) in DECIBELS_LINEAR_TABLE {
-            dbg!(db, linear, f32::abs(to_linear(db) - linear));
-            assert!(f32::abs(to_linear(db) - linear) < EPSILON)
+            const PRECISION: f32 = 5.066e3;
+            let to_linear = to_linear(db);
+
+            assert!(
+                2.0 * (to_linear - linear).abs()
+                    < PRECISION * f32::EPSILON * (to_linear.abs() + linear.abs())
+            );
         }
     }
 }
--- a/src/source/amplify.rs
+++ b/src/source/amplify.rs
@@ -12,7 +12,7 @@ where
     Amplify { input, factor }
 }
 
-/// converts decibels to linear
+/// Internal function that converts decibels to linear
 pub fn to_linear(decibels: f32) -> f32 {
     f32::powf(10f32, decibels * 0.05)
 }
--- a/src/source/mod.rs
+++ b/src/source/mod.rs
@@ -5,10 +5,11 @@ use core::time::Duration;
 
 use crate::common::{ChannelCount, SampleRate};
 use crate::Sample;
+use amplify::to_linear;
 use dasp_sample::FromSample;
 
 pub use self::agc::AutomaticGainControl;
-pub use self::amplify::{to_linear, Amplify};
+pub use self::amplify::Amplify;
 pub use self::blt::BltFilter;
 pub use self::buffered::Buffered;
 pub use self::channel_volume::ChannelVolume;
--- a/src/source/mod.rs
+++ b/src/source/mod.rs
@@ -268,16 +268,14 @@ pub trait Source: Iterator<Item = Sample> {
         const LOG_VOLUME_GROWTH_RATE: f32 = 6.907_755_4;
         const LOG_VOLUME_SCALE_FACTOR: f32 = 1000.0;
 
-        if value > NORMALIZATION_MIN && value < NORMALIZATION_MAX {
-            let mut amplitude = f32::exp(LOG_VOLUME_GROWTH_RATE * value) / LOG_VOLUME_SCALE_FACTOR;
-            if value < 0.1 {
-                amplitude *= value * 10.0;
+        let value = value.clamp(NORMALIZATION_MIN, NORMALIZATION_MAX);
 
-                return amplify::amplify(self, amplitude);
-            }
+        let mut amplitude = f32::exp(LOG_VOLUME_GROWTH_RATE * value) / LOG_VOLUME_SCALE_FACTOR;
+        if value < 0.1 {
+            amplitude *= value * 10.0;
         }
 
-        amplify::amplify(self, value.clamp(NORMALIZATION_MIN, NORMALIZATION_MAX))
+        amplify::amplify(self, amplitude)
     }
 
     /// Applies automatic gain control to the sound.
--- a/src/source/mod.rs
+++ b/src/source/mod.rs
@@ -255,6 +255,7 @@ pub trait Source: Iterator<Item = Sample> {
     /// Normalized amplification in `[0.0, 1.0]` range. This method better matches the perceived
     /// loudness of sounds in human hearing and is recommended to use when you want to change
     /// volume in `[0.0, 1.0]` range.
+    /// based on article: https://www.dr-lex.be/info-stuff/volumecontrols.html
     ///
     /// **note: it clamps values outside this range.**
     #[inline]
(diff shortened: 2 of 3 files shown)


daalfox:
This is my first contribution on this create. Just checking in with you about the general approach for this. Is it ok?

src/source/amplify.rs
```diff
@@ -1,11 +1,55 @@
+use std::ops::Mul;
 use std::time::Duration;
 
 use super::SeekError;
 use crate::common::{ChannelCount, SampleRate};
 use crate::Source;
 
+/// Amplify factor
+#[derive(Clone, Debug)]
+pub enum AmpFactor {
```
dvdsk, line 10: `VolumeChange` sounds better then `AmpFactor`

src/source/amplify.rs
```diff
@@ -53,7 +97,9 @@ where
 
     #[inline]
     fn next(&mut self) -> Option<Self::Item> {
-        self.input.next().map(|value| value * self.factor)
+        self.input
+            .next()
+            .map(|value| value * f32::from(self.factor.clone()))
```
dvdsk, line 102: this is running `f32::from()` and `factor.clone()` about 96 thousand times a second. Instead you could keep this as a simple multiply and do the `f32::from()` in `set_factor`.

src/source/amplify.rs
```diff
@@ -16,13 +60,13 @@ where
...
     input: I,
-    factor: f32,
+    factor: AmpFactor,
 }
 
 impl<I> Amplify<I> {
     /// Modifies the amplification factor.
     #[inline]
-    pub fn set_factor(&mut self, factor: f32) {
+    pub fn set_factor(&mut self, factor: AmpFactor) {
```
dvdsk, line 69: alternative design: do not use `AmpFactor` but keep this a `f32`, then add a second method to `Amplify` for setting a decibel factor. For example: `set_log_factor`.

src/source/mod.rs
```diff
@@ -442,7 +443,7 @@ pub trait Source: Iterator<Item = Sample> {
     /// let source = source.buffered().reverb(Duration::from_millis(100), 0.7);
     /// ```
     #[inline]
-    fn reverb(self, duration: Duration, amplitude: f32) -> Mix<Self, Delay<Amplify<Self>>>
+    fn reverb(self, duration: Duration, amplitude: AmpFactor) -> Mix<Self, Delay<Amplify<Self>>>
```
dvdsk, line 446: I see why you want for an `enum`. We could still add a method `reverb_decibel`.

Alternatively maybe a builder like approach could be nice. Reverb gets to be its own type and this function no longer takes an amplitude argument. Instead we use a default value and you set the amplitude later with a setter. There are then two setters mirroring, a log/decibel one and a linear one.

dvdsk (changes requested):
I understand the choices you made however I think we can do better. Let me know what you think of the feedback.

Further more did you look at the resource Roderick linked in the issue? It features a slightly more involved calculation that leads to a nicer volume curve. We could use that instead or have both the current log and that nicer one.

daalfox:
Yeah I wanted to keep myself in the same page with you. I just skimmed through the article I will check it out. Thanks for review

roderickvd:
You can look at the code I linked to, which compresses that article into 10 lines of code. You can copy and/or modify from me under public domain.

daalfox:
> You can look at the code I linked to, which compresses that article into 10 lines of code. You can copy and/or modify those lines from me under public domain.

@roderickvd Do I add this to `Sink`, `SpacialSink` and `ChannelVolume` and call it `normalized_volume` or something?

src/source/amplify.rs
```diff
@@ -12,6 +12,11 @@ where
     Amplify { input, factor }
 }
 
+/// converts decibels to linear
+pub fn to_linear(decibels: f32) -> f32 {
```
dvdsk, line 16: why is this public? (its generally good to keep as much private as possible that way its not breaking if we remove/change it in the future)

dvdsk (changes requested):
Take a look at `source/mod.rs`. Do we want a `amplify_decibel()` in there?

dvdsk:
> @roderickvd Do I add this to `Sink`, `SpacialSink` and `ChannelVolume` and call it `set_normalized_volume` or something? Maybe I open a new PR for that?

I would also add it to amplify. But you are right, there are many places we deal with volume atm, we will probably want it there too.

daalfox:
So maybe we keep `to_linear` that is for any decibel value, an internal api and we add a `to_linear_normalized` to change a linear value from `[0.0, 1.0]` to a logarithmic one?
Not sure if I'm understanding this currectly

daalfox:
The main issue I'm having here is that I don't understand how I fit the mapping that @dvdsk suggested into the `Amplify` interface
To my little knowledge I got from this PR, this `Amplify` struct just multiplies each sample in the source with the given `factor` and we don't deal with volume here at all.

roderickvd:
Maybe we need to sketch out & agree on the API first. To me, `Sink::set_volume` and `Source::amplify` are the same in that both modify signal amplitude.

```rust
// Direct linear gain
sink.set_volume(1.5);  // amplify by 150%
source.amplify(0.5);   // attenuate to 50%

// Using decibels
sink.set_volume(db_to_gain(6.0));  // +6 dB
source.amplify(db_to_gain(-3.0));  // -3 dB

// Using normalized with logarithmic scaling
sink.set_volume(norm_to_log_gain(0.5));   // 50% perceived volume
source.amplify(norm_to_log_gain(0.75));   // 75% perceived volume
```

daalfox:
The thing I like about the `Amplify` is that it just amplifies the samples and doesn't care about the human interpretation of loudness and changing that feels wrong. I also think that just adding some public helper functions feel a lot nicer.

dvdsk:
## Terms

> Maybe we need to sketch out & agree on the API first. To me, `Sink::set_volume` and `Source::amplify` are the same in that both modify signal amplitude.

Funny, to me setting volume and amplifying are different :)

Lets see what kind of terms we got:
- Volume: absolute signal strength at output
- Gain: change to signal strength before output
- Amplification: unit of gain?

The name of `Sink::set_volume` follows that definition. Since `Sink` is connected directly into the output mixer it is the last part of a rodio pipeline so you could say amplifying their is modifying the volume.

> amplifies the samples and doesn't care about the human interpretation of loudness and changing that feels wrong

Yes I get that. That makes sense from a programming standpoint. On the other hand human hearing is logarithmic so linear amplification is generally not the best from a UI perspective and thus should not be the default.

## Helper vs UI
> adding some public helper functions feel a lot nicer

That would make it clear that decibels is just another way of expressing that amplification factor. On the other hand its not super discoverable and using a factor would still be the default.

this:
```rust
some_source
  .pausable()
  .fade_in(Duration::from_secs(60))
  .amplify(to_linear(10));

other_source
  .pausable()
  .fade_in(Duration::from_secs(60))
  .amplify(1.1);
```

is to me less clear then:
```rust
some_source
  .pausable()
  .fade_in(Duration::from_secs(60))
  .amplify_by_db(10);

some_source
  .pausable()
  .fade_in(Duration::from_secs(60))
  .amplify_by_factor(1.1);
```

Its a difficult UI to get right....

dvdsk:
>``` // Using normalized with logarithmic scaling
> sink.set_volume(norm_to_log_gain(0.5));   // 50% perceived volume
> source.amplify(norm_to_log_gain(0.75));   // 75% perceived volume
> ```

I like the use of perceived there, `maybe perceived_gain` would work somewhere. `amplify_with_perceived_increase` (using increase since is an easier word then gain).

dvdsk:
I agree with @roderickvd btw and I think we should keep going back and forth till we all agree we have the best possible API.

daalfox:
Maybe it's best to close this and revisit it at some point later wdyt? I plan to help with refactoring on this crate. Is there any particular section you want me to work on?

dvdsk:
> Maybe it's best to close this and revisit it at some point later wdyt?

I see no reason to close this, we need a little more input to figure out the best API though. I'll mark it as help wanted and hope someone else chimes in.

dvdsk:
> Feel like `amplify` can also be used in place of `set_volume`s

on sink? yes though that would be a breaking change affecting a lot of users. Since we already want to redo sink entirely I rather keep such a change till then.

Maybe @roderickvd or @PetrGlad know whats a good issue to work on while we wait for more input on this.

roderickvd:
> Maybe @roderickvd or @PetrGlad know whats a good issue to work on while we wait for more input on this.

You know, I'd actually like to get agreement on the API and have @daalfox continue work on it. It's useful, not too hard, and has already begun, so let's also complete it. We just need agreement.

> > Feel like `amplify` can also be used in place of `set_volume`s
> 
> on sink? yes though that would be a breaking change affecting a lot of users. Since we already want to redo sink entirely I rather keep such a change till then.

I also do not see the point of breaking the current API, although:
- it'd be OK to deprecate one if we'd really decide that would be the way to go (not convinced)
- I do agree that they currently do the same thing

Because w.r.t. to your @dvdsk stating:

> * Volume: absolute signal strength at output

I don't disagree with that definition, but that's currently not what it does. `set_volume` just multiplies the sample value by some number, without taking into account any gain that became before. As far as I can see, `Sink` even delegates to `Amplify` to do it.

They are the same today, and I don't recognize a need for `set_volume` to know about & compensate for any prior gain. In the future, I do think it's be great if it could (optionally?) gradually ramp volume up/down when changing volume, in order to prevent pops and clicks.

So back to the API... I'm for not breaking the current API if we don't need to, and we don't need to. So it comes down to the question: do we want helper functions (less discoverable, easy to plug as arguments to the existing functions) or do we want new functions? Let's take it from there and then we can decide on naming.

daalfox:
I have one more question, do we panic outside of [0.0, 1.0] range for normalized?

dvdsk:
> I have one more question, do we panic outside of [0.0, 1.0] range for normalized?

We could also return a Result<(), OutOfRangeError>. That might be better since it would trigger users to read the docs :). Which would note the required range and have an Errors section pointing that out.

dvdsk:
> You know, I'd actually like to get agreement on the API

agreed

> So it comes down to the question: do we want helper functions (less discoverable, easy to plug as arguments to the existing functions) or do we want new functions?

I prefer new functions

roderickvd:
> I have one more question, do we panic outside of [0.0, 1.0] range for normalized?

We’re talking about the 0-100% “volume slider” right? For ease of use we could also simply clamp and document that.

dvdsk:
> > I have one more question, do we panic outside of [0.0, 1.0] range for normalized?
> 
> We’re talking about the 0-100% “volume slider” right? For ease of use we could also simply clamp and document that.

yeah lets do that :+1:

daalfox:
Feel free to rephrase the documentations as you like.

src/source/amplify.rs
```diff
@@ -93,3 +104,50 @@ where
...
+        (-80., 0.0001),
+        (-90., 0.00003162),
+        (-100., 0.00001),
+    ];
+
+    #[test]
+    fn convert_decibels_to_linear() {
+        for (db, linear) in DECIBELS_LINEAR_TABLE {
+            dbg!(db, linear, f32::abs(to_linear(db) - linear));
+            assert!(f32::abs(to_linear(db) - linear) < EPSILON,)
```
roderickvd, line 150: There's a superfluous `,` (comma) right before the end parenthesis.

Instead of redefining `EPSILON`, can we:
1. use a different name;
2. and/or use some form of comparison with relative error:

```rust
2.0 * (x - y).abs() < z * f32::EPSILON * (x.abs() + y.abs())
```

where `z` can be changed to handle numerical stability and precision.

src/source/mod.rs
```diff
@@ -242,6 +243,42 @@ pub trait Source: Iterator<Item = Sample> {
...
+        if value > NORMALIZATION_MIN && value < NORMALIZATION_MAX {
+            let mut amplitude = f32::exp(LOG_VOLUME_GROWTH_RATE * value) / LOG_VOLUME_SCALE_FACTOR;
+            if value < 0.1 {
+                amplitude *= value * 10.0;
+
+                return amplify::amplify(self, amplitude);
+            }
+        }
+
+        amplify::amplify(self, value.clamp(NORMALIZATION_MIN, NORMALIZATION_MAX))
```
roderickvd, line 279: I think the clamping should be before. Line 270 does effectively the same.

src/source/mod.rs
```diff
@@ -242,6 +243,42 @@ pub trait Source: Iterator<Item = Sample> {
...
+    /// Amplifies the sound logarithmically by the given value.
+    #[inline]
+    fn amplify_decibel(self, value: f32) -> Amplify<Self>
+    where
+        Self: Sized,
+    {
+        amplify::amplify(self, to_linear(value))
+    }
+
+    /// Normalized amplification in `[0.0, 1.0]` range. This method better matches the perceived
```
roderickvd, line 255: For proper attribution, I suggest adding a link to the source of the algorithm: https://www.dr-lex.be/info-stuff/volumecontrols.html

roderickvd (commented):
Thanks a lot picking this up again.

src/source/mod.rs
daalfox: You mean like this?
```rust
        if !(NORMALIZATION_MIN..=NORMALIZATION_MAX).contains(&value) {
            return amplify::amplify(self, value.clamp(NORMALIZATION_MIN, NORMALIZATION_MAX));
        }

        let mut amplitude = f32::exp(LOG_VOLUME_GROWTH_RATE * value) / LOG_VOLUME_SCALE_FACTOR;
        if value < 0.1 {
            amplitude *= value * 10.0;
        }

        amplify::amplify(self, amplitude)
```

src/source/amplify.rs
daalfox: Can you help me with this one? I'm not too familiar with floating precision.

src/source/mod.rs
roderickvd: I was thinking like so:

```rust
let value = value.clamp(NORMALIZATION_MIN, NORMALIZATION_MAX);
```

Are we after the same thing? For me this would limit the "volume slider" to not go negative or above 100%.

src/source/amplify.rs
roderickvd: Sure, glad that you ask. Try this:

```rust
const PRECISION: f32 = 1.0;
let to_linear = to_linear(db);
assert!(2.0 * (to_linear - linear).abs() < PRECISION * f32::EPSILON * (to_linear.abs() + linear.abs()))
```

And try to find the smallest `PRECISION` that passes the test.
And/or manually compute the `linear` results in the array with more precision.

Let me know if this helps.

src/source/mod.rs
daalfox: I thought if I clamp before, code in the condition will always run, but the way I did, the condition doesn't run if value is out of range and we just clamp it without running computations inside if clause. Currect me if I'm wrong.

PetrGlad: I am not sure the signal should be clamped. While it is processed the samples can go well outside -1..1. It must be normalized or limited before actual output. Another problem might be if a filter expects limited signal already and may break otherwise. I would make normalization or limiting and explicit flag or even maybe a separate source/filter. 
Another consideration is that hard clamping will introduce distortions.

roderickvd: @PetrGlad agreed, though note that we are talking about limiting the volume factor, not the signal. Definitely should not limit the signal.

> I thought if I clamp before, code in the condition will always run, but the way I did, the condition doesn't run if value is out of range and we just clamp it without running computations inside if clause. Currect me if I'm wrong.

This logarithmic amplification is valid only for a volume range of `0.0..=1.0`, like you have documented, so unless *I* am missing something, I'd say just clamp right at the beginning and be done with it.

dvdsk (approved):
This looks great to me! Ended up being quite a nuanced feature. Thank you for getting this through!

daalfox:
Yeah that was something. Thank you everyone

dvdsk:
missed something (no worries its the tiniest thing). This deserves a change-log entry! :)

daalfox:
> missed something (no worries its the tiniest thing). This deserves a change-log entry! :)

Done

roderickvd:
Awesome thanks for hanging out with us 👍 

As one last improvement may I request that you reword the changelog statement of "better volume control" to be more specific like "perceptually linear"?

dvdsk:
:partying_face:

Issue #296, opened 2020-07-02
DynamicMixer stops working if it ever has zero sources


kaikalii:
Title. From what I can tell from the source code, there are a couple checks where if the `DynamicMixer` has no sources, `Iterator::next` returns `None`, which terminates the mixer `Source`. From this point, any sources added via `DynamicMixerController::add` will not be played.

The current workaround I am using for this is to immediately add a `Zero` source when creating the mixer, thus keeping it "alive". However, I don't think this should be necessary.

wkordalski:
At least some note in the documentation of dynamic mixer should exist.
This could help me when sometimes the sound was playing and sometimes not (depending what occured first: adding the first source or detaching dynamic mixer from output).

I agree, that adding `Zero` source should not be necessary.

PR #719, opened 2025-03-26
Document the behavior of empty mixer

wkordalski:
The behavior of empty mixer can be surprising.
See: https://github.com/RustAudio/rodio/issues/296

Documenting this behavior can shorten the debugging sessions when the sound sometimes is played and sometimes not.

Feel free to rephrase the docstring.


dvdsk (approved):

dvdsk:
Thanks for the addition! These kind of doc changes are quite valuable.

PR #725, opened 2025-04-03
Revert to default buffer size

### LINKED ISSUE #724: Regression: Failing audio playback on Raspberry Pi.
After updating from 0.20.1 to the current master branch, audio playback stops after a split second on a first generation Raspberry Pi. 
The error `audio stream error: A backend-specific error has occurred: `alsa::poll()` returned POLLERR` is repeatedly printed to stderr.

This happens when using rodio as a dependency, as well as when running examples directly from the rodio repository (tested with music_wav and music_mp3).

This is on fully up to date Raspberry Pi OS where I have had Rodio 0.20.1 running successfully.

roderickvd:
Current v0.20.1 uses the backend's default buffer size. `master` has changed this to prefer a period size of 1024. This cannot be expected to always work correctly, even if it's within the minimum / maximum reported supported frame count.

This reverts that back to the default, while still allowing the `with_buffer_size` builder method to override it.

Also includes some clippy fixes now that Rust 1.86 is out in a separate commit.

Fixes #724

### DIFF
--- a/benches/effects.rs
+++ b/benches/effects.rs
@@ -12,7 +12,7 @@ fn main() {
 
 #[divan::bench]
 fn reverb(bencher: Bencher) {
-    bencher.with_inputs(|| music_wav()).bench_values(|source| {
+    bencher.with_inputs(music_wav).bench_values(|source| {
         source
             .buffered()
             .reverb(Duration::from_secs_f32(0.05), 0.3)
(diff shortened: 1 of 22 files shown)


dvdsk (approved):

PetrGlad:
Default buffer size may be unexpectedly long (seconds). That is why I capped it to that value.

PR #726, opened 2025-04-04
Turn documention of with_buffer_size into guide

dvdsk:
The errors caused by too small buffers (POLLERR for example) are hard to debug. We can not rely on minimum buffer sizes provided by the OS through CPAL. This makes this parameter rather important which is why I expanded the setters docs into more of a guide.

See discussion in:
https://github.com/RustAudio/rodio/pull/725
and 
https://github.com/RustAudio/rodio/issues/724


src/stream.rs
```diff
@@ -163,8 +163,33 @@ where
...
+    /// minimum delay (latency) of `<buffer size>/<sample_rate*channel_count>`
+    /// seconds before a change made through rodio takes effect.
+    ///
+    /// ## Large vs Small buffer
+    /// - A larger buffer size results in high latency. Changes made trough
+    ///   Rodio (volume/skip/effects etc) takes longer before they can be heard.
+    /// - A small buffer might cause higher CPU usage, playback interruptions
+    ///   or result in errors like: `alsa::poll() returned POLLERR`
+    ///
+    /// ## Recommendation
```
roderickvd, line 189: Would we want to mention some typical values, like:
- low-latency (audio production, live monitoring): 512-1024
- general use (games, media playback): 1024-2048
- stability-focused (background music, non-interactive): 2048-4096

A reason *not* to do it, is that this is somewhat opinionated.
Then again, it could put novices on the right track.

roderickvd (commented):
Good idea.

PR #728, opened 2025-04-05
Use Arc in SamplesBuffer for cheap Clones

yanqui-uxo:


dvdsk:
good idea!

Can you turn it into an Arc<[sample]> though?

That should work via an `.into()` on the Vec. Example:
```let thing: Arc<[i32]> = vec![1, 2, 3].into();```

PetrGlad:
I wonder if there are real use cases where this would help?

yanqui-uxo:
I created this PR because my project involves playing premade sources on key press, which can overlap with themselves. From my understanding, this necessitates cloning.

dvdsk:
Its also useful in games for often repeated audio.

Issue #539, opened 2024-01-18
Rodio's features can be confusing


dvdsk:
#### Terminology
An audio file is a _container_ (mp4/mkv). A _container_ contains one or more streams and possibly metadata. Those streams are encoded with a _codec_. For audio those streams are usually multiple tracks. `Rodio` currently has no way to deal with multiple tracks in one file/container. 

Some containers are special and can contain only a stream encoded with one specific codec. For example _mp3_ contains only _MPEG-1 Audio Layer 3_ encoded stream. _Flac_ is another of these.

#### Features in rodio
Features in `rodio` can have two functions:

- allow users to pick one decoder over another (for example `minimp3` vs `symphonia`). Though this can only be done by disabling/enabling support by a decoder for a container.
- speed up compilation by disabling unused codecs. Important for upstream libs building on `rodio` like `bevy`

Right now there is no structure to the `symphonia` related features. There are features which 
 - enable a single _container_ 
 - a single _container_ and _codecs_ often used with it. 
 - a single _codec_.

Right now (cause of #516) we can not play _ogg_ files. While there is a feature to enable the _vorbis codec_ usually used in an _ogg_ there is no feature to enable support for the _ogg container_. This will be addressed shortly.

#### Issue
This can be pretty confusing to users not familiar with the difference between codecs and decoders (such as me this morning). The errors they get if the fail to enable a codec are puzzling to them. Without looking at the source code it is not possible to predict what a feature combination will do.

#### Proposal
1.  Document how enabled container feature will decides the decoder that will be used. Using different decoders for different streams in the same container will only be possible when specifying enabled codec and containers separately. It will be possible for example to configuring rodio such that it uses `lewton` for _ogg_ containing a _vorbis_ stream while decoding _ogg_ containing _flac_ with `symphonia`.

 2. Make the current container features (for example `symphonia-isomp4`) enable all codecs. This removes the pitfall of #516. 
 
 4. Add features that only enable the container without any codec to facilitate 1. and speeding up compilation.

_________________

Please let me know what you (the rodio user) think.

roderickvd:
Can we close this? I think the features got a good upgrade.

Issue #516, opened 2023-10-11
Reverb example rans into error with symphonia vorbis decoder 


dvdsk:
Probably belongs upstream in symphonia, ran into it testing #513. 

seems like `assets/music.ogg` breaks the decoder.

Needed it documented somewhere so here it is. Reproducible on master branch with:

```
cargo r --example reverb --features symphonia-all
```

est31:
It's not hard to test if the culprit is symphonia: symphonia probably also has examples in their own repo. And if not, there is probably some symphonia user one could try.

dvdsk:
replicated upstream, moving this issue there.

see: https://github.com/pdeljanov/Symphonia/issues/235

dvdsk:
Issue is on this side, I'm working on it. Expect a PR sometime this week.

PR #731, opened 2025-04-11
fix: minimal build with no default features

### LINKED ISSUE #730: Build failure with no default features
Latest `master` does not build with no default features or only `playback`. The decoder throws a whole bunch of errors. I'll take it on.

roderickvd:
This commit makes the decoder module and its exports conditional on the having at least one decoder format enabled. Also changes the CI to prevent regression.

Fixes #730

### DIFF
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -780,7 +780,6 @@ impl fmt::Display for DecoderError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         let text = match self {
             DecoderError::UnrecognizedFormat => "Unrecognized format",
-            #[cfg(feature = "symphonia")]
             DecoderError::IoError(msg) => &msg[..],
             #[cfg(feature = "symphonia")]
             DecoderError::DecodeError(msg) => msg,
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,7 @@ name: CI
 on:
   pull_request:
   push:
-    branches: [ main, master ]
+    branches: [main, master]
 
 env:
   RUSTFLAGS: "-C debuginfo=0 -D warnings"
@@ -18,8 +18,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ macos-latest, windows-latest, ubuntu-latest ]
-        toolchain: [ stable, beta, nightly ]
+        os: [macos-latest, windows-latest, ubuntu-latest]
+        toolchain: [stable, beta, nightly]
         include:
           - os: macos-latest
             MACOS: true
@@ -49,6 +49,5 @@ jobs:
       # `cargo test` does not check benchmarks and `cargo test --all-targets` excludes
       # documentation tests. Therefore, we need an additional docs test command here.
       - run: cargo test --doc
-      # Check minimal build. `tests/seek.rs` fails if there are no decoders at all,
-      # adding one to make the tests check pass.
-      - run: cargo check --tests --lib --no-default-features --features mp3
+      # Check minimal build.
+      - run: cargo check --no-default-features
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -165,6 +165,13 @@ mod wav_output;
 
 pub mod buffer;
 pub mod conversions;
+#[cfg(any(
+    feature = "flac",
+    feature = "mp3",
+    feature = "symphonia",
+    feature = "vorbis",
+    feature = "wav",
+))]
 pub mod decoder;
 pub mod mixer;
 pub mod queue;
@@ -172,6 +179,13 @@ pub mod source;
 pub mod static_buffer;
 
 pub use crate::common::{ChannelCount, Sample, SampleRate};
+#[cfg(any(
+    feature = "flac",
+    feature = "mp3",
+    feature = "symphonia",
+    feature = "vorbis",
+    feature = "wav",
+))]
 pub use crate::decoder::Decoder;
 pub use crate::sink::Sink;
 pub use crate::source::Source;


roderickvd:
I thought about creating a transitive "decoder" feature instead of duplicating the "symphonia, wav, flac, etc." list. Did not do it because it may confuse users.

src/decoder/mod.rs
```diff
@@ -780,7 +780,6 @@ impl fmt::Display for DecoderError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         let text = match self {
             DecoderError::UnrecognizedFormat => "Unrecognized format",
-            #[cfg(feature = "symphonia")]
```
roderickvd, line 783: It's not just Symphonia that uses this variant. I started trying to fix up the decoder to work with no formats, but there's no point, so switched to feature gating.

roderickvd:
Argh. Thought about a quick fix, but the rabbit hole goes deeper. Compiling with just `--no-default-features --features playback` also gives a whole slew of non-trivial errors. There are a lot of expectations that there is at least *one* decoder today.

I see two options:
1. *Do* create a transitive "decoder" feature, and gate everything behind it.
2. Accept that Rodio needs at least one decoder to compile correctly.

I'm for number 1 but it'll take a bit more time, which I'm unlikely to have coming few nights.

dvdsk:
I want to draw your attention to this: https://github.com/RustAudio/rodio/issues/539

the tldr: compile features are a mess ATM and offer a lot of avenues for issues.

I say if we start changing (and breaking some code that way) lets fix things properly and overhaul the entire feature set.

roderickvd:
Sure, I understand. What would you like to do for v0.21?

Remove this “decoder” feature for now, until we sort everything out, and just duplicate the `#[cfg(any(…)]` everywhere?

Or: accept that compiling without default features is broken? I’d rather not because it breaks projects that don’t use any part of the decoder.

dvdsk:
There is no rush to release v0.21 right? I can try and implement the changes proposed in #539 and then combine that with the transitive decoder feature you proposed. Then the theme for v0.21 will be: 'removing footguns'. Which I kinda like.

roderickvd:
Fine for me - I thought we were in a release train but totally fine. Just to check: so you’re taking over on that feature issue? Let me know if and what you’d like me to contribute.

dvdsk:
> Fine for me - I thought we were in a release train but totally fine.

Yes your right there. This could be an easy change however, and if we do it now we probably no longer need to make breaking changes to features in the future.

> so you’re taking over on that feature issue? Let me know if and what you’d like me to contribute.

I can allocate some time to take on #539. It seems like its own issue to me, on the other hand maybe changing the features should be one PR. I have no idea how to address it (539) yet. If you have ideas please, feel free to take on #539 and merge that with this :).

dvdsk:
Taken from 539:

>1.  Document how enabled features determines the decoder used by `rodio`. Using different decoders for different streams in the same container will only be possible when specifying enabled codec and containers separately. As an example, it will be possible to configure `rodio` such that it uses `lewton` for _ogg_ containing a _vorbis_ stream while decoding _ogg_ containing _flac_ with `symphonia`.
>
> 2. Make the current container features (for example `symphonia-isomp4`) enable all codecs. This removes the pitfall of #516. 
> 
> 3. Add features that only enable a container without enabling any codecs to facilitate 1. and allow speeding up compilation.

Looking at it now the above seems overkill to me. Maybe we should have a feature for each container and none for codecs (they are all enabled when a container that can use them is enabled).

I say we first propose what we want the features to be before implementing them.

src/stream.rs
```diff
@@ -283,6 +291,7 @@ where
     }
 }
 
+#[cfg(feature = "__decoder")]
```
PetrGlad, line 294: Without playback support this whole file should be excluded.
Does anyone actually need to exclude all features, including default ones?

roderickvd: Generally I suggest as best practice that projects compile correctly with no default features. 

I ran into this myself, as I was upgrading a project that only uses the playback feature but none of the decoders.

roderickvd:
> Looking at it now the above seems overkill to me. Maybe we should have a feature for each container and none for codecs (they are all enabled when a container that can use them is enabled).

Enabling all codecs for a container leads to the question: the Symphonia ones or the others? When both are enabled ("they are all enabled") then the implementation in v0.20 as well as v0.21 will give "the others" (hound, lewton, claxon) precedence over Symphonia.

Currently, default decoders are lewton, claxon, hound, and finally Symphonia only for MP3. Here's an opinion, I don't know how popular: how about changing the defaults all to Symphonia instead?

Pros: more functionality, more consistent without hybrid Symphonia / other decoder mix 
Cons: some users may not want to have MPL licensed decoders. To me not really an issue if we still give them choice to switch to "the others".

All these variants should compile without warning then:
```bash
# defaults: playback,flac,vorbis,wav,mp3
# all decoders Symphonia
# "vorbis" will also enable "ogg"
cargo build 

# defaults plus aac
# "aac" will also enable "mp4"
cargo build --features aac

# whoever just wants to have the DSP modules
cargo build --no-default-features

# as a playback library with own sources
cargo build --no-default-features --features playback

# like the defaults but with some different Symphonia decoders
# "aac" will also enable "mp4"
# "vorbis" will also enable "ogg"
cargo build --no-default-features --features playback,vorbis,aac,aiff

# like the defaults but with alternative decoders
cargo build --no-default-features --features playback,claxon,lewton,hound,minimp3

# advanced usage: lots of stuff could be in a container
cargo build --no-default-features --features flac,mp4
```

This will probably still require an internal "decoder" feature flag.

Feel free to improve.

dvdsk:
> how about changing the defaults all to Symphonia instead?

agreed! symphonia has proven great. And as you say if the license proves an issue (which it should not) users can fall back. The current situation is also a small footgun since symphonia implements more features (mainly seek). Having those missing while having enabled a feature for your format is confusing.

I would also like to separate codecs and containers. Most users do not know the difference and that is yet another footgun. You enable ogg but you miss vorbis etc (not the best  example). We could have three categories of codec/container features:
- `container`: enables all the codecs supported by that container
- `container_without_codec`: enables only that container
- `codec`: enables only that codec

Examples would be:
- `wav`: enables `symphonia/wav`, `symphonia/pcm` and `symphonia/adpcm`
- `wav_without_codecs`: enables only `symphonia/wav`. Useless without `pcm` or `adpcm`
- `pcm`, `adpcm`: both are useless without `wav_without_codecs`

Alternative and most user friendly would be to only have the container features.

PetrGlad (dismissed):
Check for no-default features was added to accommodate a no-playback build option that some users requested. In particular, with that option the whole output stream's code could (and should) be excluded.
Because Cargo does not support a "negative" feature specification, the playback feature was marked as default only to avoid surprises in the usual case when users do want playback. This way only users who actually want no playback would have to deal with more complicated dependency requiring  "no-default-features" flag.
I do not mind supporting even slimmer version of rodio as long as it is either required for some use-case or does not complicate things.

If you feel like this option is needed, see my other comments here.

roderickvd:
Not late at all! So that removes any need for an internal "decoder" feature right? Both are kinda hackish but yours seems the lesser of evils.

dvdsk:
> the fetching samples one-by-one approach that is currently used by `Source`

the "hope" is that the compiler will inline those calls, so having a consumer of Source that buffers should lead to the compiler unrolling 'some' loops. 

> How much of a hit is difficult to say without trying it out and profiling

Would be nice to see, you should know though there is a wider discussion regarding the `Iterator approach` vs alternatives. We are not ready to announce anything yet but the discussion is public. If your interested shoot me a mail and I'll send you a link (you'll find my email on davidsk.dev).

roderickvd:
Sorry, was occupied with other things before I can jump on this. 

But indeed I’d vote to prevent going the `dyn` route. We should go out of our way and enable the compiler to do smart things. 

I’m not sure what you mean with being too optimistic with the builder. I agree that it’s improbable that we’ll ever see more backends, let alone “lots more”. Honestly that also wasn’t the intent of the builder. It was there to expose some of Symphonia’s features, without tight coupling, and still allowing the other decoders to have their place. 

If it were me as a user, I would dispense with the other decoders and go full Symphonia. But me as a Rodio collaborator, I want to allow for other uses to use differently licensed decoders.

PetrGlad:
If `dyn` is necessary then we can consider using buffers? I remember that was discussed in part elsewhere here.
In that case the processing could even be vectorized (inside a filter), but the whole implementation will be more complex (and complicated).

roderickvd:
> Probably really late to this, but there is a third solution:
> 
> ```rust
> #[allow(clippy::large_enum_variant)]
> enum DecoderImpl<R: Read + Seek> {
>     #[cfg(all(feature = "wav", not(feature = "symphonia-wav")))]
>     Wav(wav::WavDecoder<R>),
>     #[cfg(all(feature = "vorbis", not(feature = "symphonia-vorbis")))]
>     Vorbis(vorbis::VorbisDecoder<R>),
>     #[cfg(all(feature = "flac", not(feature = "symphonia-flac")))]
>     Flac(flac::FlacDecoder<R>),
>     #[cfg(all(feature = "minimp3", not(feature = "symphonia-mp3")))]
>     Mp3(mp3::Mp3Decoder<R>),
>     #[cfg(feature = "symphonia")]
>     Symphonia(symphonia::SymphoniaDecoder, PhantomData<R>),
>     #[allow(dead_code)]
>     None(Unreachable, PhantomData<R>),
> }
> 
> enum Unreachable {}
> ```
> 
> This requires adding an unreachable branch to all `match` statements, but gets rid of all compile errors, leaving only warnings about unused items.

Anyone against this for a v0.21 target? I like it best of the various options.

roderickvd:
Just force pushed, replacing my previous take with the better proposal of @renski-dev.

dvdsk (approved):

roderickvd:
@PetrGlad do you have any further review comments? If not, I'd like to get this merged. It seems we're good to release v0.21 after that!

PetrGlad:
@roderickvd I am sorry for not answering earlier. It seems I took on more things than I could digest...

PetrGlad (commented):
@roderickvd I am sorry for not answering earlier. 
I can build with --no-default-features but the test build fail (mostly in examples)

cargo  test --no-default-features

Do you intent to clean that up here or is there another plan for these?

roderickvd:
> @roderickvd I am sorry for not answering earlier.

No worries, I know how life goes. Thanks for responding now.

> I can build with --no-default-features but the test build fail (mostly in examples)
>
> cargo test --no-default-features
> 
> Do you intent to clean that up here or is there another plan for these?

I agree that an absolutely minimal build should compile and test without error or warning. It turned out to be a bit more work than just in `examples/*.rs` but I made it so in 0498f72.

yara-blue:
I figured out a way to get the examples to work without needing conditional compile statements :partying_face: 

we need to put this in `Cargo.toml`
```toml
[[example]]
name = "basic"
required-features = ["playback"]
```

Then is you run:
```sh
cargo r --no-default-features --example basic
```
you get a nice error from cargo:
```
error: target `basic` in package `rodio` requires the features: `playback`
Consider enabling them by passing, e.g., `--features="playback"`
```

roderickvd:
Awesome I was looking for that. I'll refactor tomorrow or so.

roderickvd:
> I figured out a way to get the examples to work without needing conditional compile statements 🥳

This should do it.

tests/flac_test.rs
```diff
@@ -1,6 +1,6 @@
-#[cfg(all(feature = "flac", not(feature = "symphonia-flac")))]
+#[cfg(any(feature = "flac", feature = "symphonia-flac"))]
```
yara-blue, line 1: We could handle these the same way as with the [examples](https://doc.rust-lang.org/cargo/reference/cargo-targets.html?highlight=test#configuring-a-target). I think this is fine though. Examples should be as simple as possible so not having conditional comp in there is nice. These tests are not meant to help users so having them look a little more daunting is fine.

dvdsk (approved):

yara-blue:
thanks for all the hard work and patience everyone!

PR #734, opened 2025-04-20
fix non-symphonia builds

renski-dev:
Compiling with the following `Cargo.toml` fails:
```toml
rodio = { git = "https://github.com/RustAudio/rodio.git", default-features = false, features = [
  "vorbis",
  "noise",
  "playback",
] }
```

`cargo build` fails without the `"symphonia"` feature flag, because `try_seek` can create a `DecoderError::IoError`.

If different solution is preferred, let me know.


dvdsk:
thanks for the fix :+1:

PR #736, opened 2025-05-03
fix: remove decode error retry limit in SymphoniaDecoder

### LINKED ISSUE #735: decoding stops unexpected with malformed packets
I have an mp3 file from a podcast [link](https://mcdn.podbean.com/mf/web/xgkjyks9pfkp9w7b/EP_169_-_MLP_condamne_e_V27fwbg.mp3). The audio length is about 35 minutes.

Using the basic example to play a file, it plays just 34 seconds and then stops as if the audio is finished. There are no errors or logs shown when it happens, and it only happens with some mp3 files.
```
use std::{error::Error, io::BufReader, thread::sleep, time::Duration};
use std::env;
use rodio::Source;

fn main() -> Result<(), Box<dyn Error>> {
    let args: Vec<String> = env::args().collect();

    let file_path = &args[1];

    let (_outs, stream_handle) = rodio::OutputStream::try_default()?;
    let rsink = rodio::Sink::try_new(&stream_handle)?;

    let file = std::fs::File::open(file_path)?;
    let sink = rodio::Decoder::new(BufReader::new(file))?;
    rsink.append(sink);
    while !rsink.empty(){
        sleep(Duration::from_secs(10));
        println!("{:?}", rsink.get_pos());
    }

    Ok(())
}
```

with `Cargo.toml`
```
[package]
name = "rodio-example"
version = "0.1.0"
edition = "2024"

[dependencies]
rodio = "0.20.1"
```

I tested the same file with `symphonia-play` and it played without issues. What could be the issue?

roderickvd:
This PR changes the Symphonia decoder to just keep on decoding a file, no matter the amount of malformed packets. Previously, it would fail after 3 attempts. This could unexpectedly stop playback.

I don't see any particular danger in continuing endlessly, as at some point it will just hit end of file and stop anyway.

Fixes #735

### DIFF
--- a/src/decoder/symphonia.rs
+++ b/src/decoder/symphonia.rs
@@ -125,7 +125,12 @@ impl SymphoniaDecoder {
             match decoder.decode(&current_span) {
                 Ok(decoded) => break decoded,
                 Err(e) => match e {
-                    Error::DecodeError(_) => continue,
+                    Error::DecodeError(_) => {
+                        // Decode errors are intentionally ignored with no retry limit.
+                        // This behavior ensures that the decoder skips over problematic packets
+                        // and continues processing the rest of the stream.
+                        continue;
+                    }
                     _ => return Err(e),
                 },
             }
@@ -300,7 +305,12 @@ impl Iterator for SymphoniaDecoder {
                 let packet = self.format.next_packet().ok()?;
                 let decoded = match self.decoder.decode(&packet) {
                     Ok(decoded) => decoded,
-                    Err(Error::DecodeError(_)) => continue,
+                    Err(Error::DecodeError(_)) => {
+                        // Skip over packets that cannot be decoded. This ensures the iterator
+                        // continues processing subsequent packets instead of terminating due to
+                        // non-critical decode errors.
+                        continue;
+                    }
                     Err(_) => return None,
                 };
 
--- a/src/decoder/symphonia.rs
+++ b/src/decoder/symphonia.rs
@@ -20,11 +20,6 @@ use crate::{
     source, Source,
 };
 
-// Decoder errors are not considered fatal.
-// The correct action is to just get a new packet and try again.
-// But a decode error in more than 3 consecutive packets is fatal.
-const MAX_DECODE_RETRIES: usize = 3;
-
 pub(crate) struct SymphoniaDecoder {
     decoder: Box<dyn Decoder>,
     current_span_offset: usize,
@@ -115,7 +110,6 @@ impl SymphoniaDecoder {
             .zip(stream.codec_params.n_frames)
             .map(|(base, spans)| base.calc_time(spans).into());
 
-        let mut decode_errors: usize = 0;
         let decoded = loop {
             let current_span = match probed.format.next_packet() {
                 Ok(packet) => packet,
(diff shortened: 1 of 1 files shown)


dvdsk:
I think this might have some issues. 
- Don't we currently try every decoder until one works in `Decoder::new`? In that case its relatively important to fail fast. 
- When an audio stream gets corrupted, or a user accidentally provides random data this could cause a hang in their app.

An alternative:
Increase the allowed error rate, maybe even make it configurable (we've got that nice decoder builder we could use that)

Let me know what you think.

roderickvd:
I see your point. But how sure are you that we want to deal with random streams of infinite length?

Problem is that setting any retry maximum is entirely arbitrary. Both for us and for any user. 

How about this: continue when we have a `total_duration` (so we can rely on hitting EOF) or return `None` otherwise?

Optionally we could introduce a setting to the builder like `strict_mode` to influence that behavior. But honestly when we can keep things simple, let’s not add more knobs.

dvdsk:
> I see your point. But how sure are you that we want to deal with random streams of infinite length?

Its a supported use case now (think live audio streams). I would like to keep supporting that.

> Problem is that setting any retry maximum is entirely arbitrary. Both for us and for any user.

True, however at some point a file is too damaged and it no longer makes sense to try and play it. Right now the retry might be a little conservative. So we could make it 5 times larger and see what happens. Instead of arbitrary lets say this new retry limit has been found empirically :)

> How about this: continue when we have a total_duration (so we can rely on hitting EOF) or return None otherwise?

That would make the automatic decoder picking really slow right?

> Optionally we could introduce a setting to the builder like strict_mode to influence that behavior. But honestly when we can keep things simple, let’s not add more knobs.

Agreed suboptimal, rodio should be easy and without hidden nobs etc.

roderickvd:
> > I see your point. But how sure are you that we want to deal with random streams of infinite length?
> 
> Its a supported use case now (think live audio streams). I would like to keep supporting that.

I thought you were also referring to a stream that's both infinite _and_ random garbage. It would also require to first have a valid header, _then_ be random garbage, otherwise it would never initialize. What are the chances of that?

> > Problem is that setting any retry maximum is entirely arbitrary. Both for us and for any user.
> 
> True, however at some point a file is too damaged and it no longer makes sense to try and play it. Right now the retry might be a little conservative. So we could make it 5 times larger and see what happens. Instead of arbitrary lets say this new retry limit has been found empirically :)

I disagree. If we fix this one empirically, then some point down the line another will show up with more. I'd rather prevent drawing some arbitrary line and deal with it algorithmically.

> > How about this: continue when we have a total_duration (so we can rely on hitting EOF) or return None otherwise?
> 
> That would make the automatic decoder picking really slow right?

I don't think so. This is for Symphonia, and all the other decoders are attempted to initialize first. So no changes there.

dvdsk:
> What are the chances of that?

vanishingly small :+1:

> I disagree. If we fix this one empirically, then some point down the line another will show up with more.

That is true only if we see playing severely broken files (clear audio skipping due to damage) as a valid use-case. I had not thought about that, now that I do we _should_ support that usecase.

> I don't think so. This is for Symphonia, and all the other decoders are attempted to initialize first. So no changes there.

I just checked and you are right.

I'm in favor now

dvdsk (approved):

PR #743, opened 2025-06-01
feat: add distortion effect

setoelkahfi:
This PR adds a simple distortion effect to the source.

Run examples:
```
// Basic
cargo run --example distortion

// With music files
cargo run --example distortion_wav
cargo run --example distortion_wav_alternate
```


examples/distortion_wav_alternate.rs
yara-blue, line 1: nice example! its great to have more stuff showing how to use `periodic_access`!

src/source/distortion.rs
```diff
@@ -0,0 +1,116 @@
+use std::time::Duration;
+
+use super::SeekError;
+use crate::common::{ChannelCount, SampleRate};
+use crate::Source;
+
+/// Internal function that builds a `Distortion` object.
+pub fn distortion<I>(input: I, gain: f32, threshold: f32) -> Distortion<I>
```
yara-blue: we have decided to stop making these functions public, that still needs to be done to the rest of the source code. So if you want we can start with distortion

src/source/distortion.rs
```diff
@@ -0,0 +1,116 @@
...
+    I: Source,
+{
+    type Item = I::Item;
+
+    #[inline]
+    fn next(&mut self) -> Option<Self::Item> {
+        self.input.next().map(|value| {
+            let v = value * self.gain;
+            let t = self.threshold;
+            if v > t {
```
yara-blue: actually the if/else if/else here can be replaced by [f32::clamp](https://doc.rust-lang.org/std/primitive.f32.html#method.clamp) right? Might be a little cleaner (and maybe that gets optimized in the future)

yara-blue:
Looks great! just one more thing and we can merge. This deserves a changelog entry! :partying_face: :balloon:

yara-blue:
thanks! 

merged into 0.21 instead of master, fixed that with some manual gitting

setoelkahfi:
Thanks!

PR #744, opened 2025-06-01
a stream is now seekable if bytes_len is known

dvdsk:
Unfortunatly this reviels a bug(?) in symphonia regarding format recognition. M4a is no longer openable with symphonia. To resolve that we can set `seekable` to true without having `bytes_len` even though that is breaking an invarint. (seekable can not be true if bytes_len is unknown).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ hound = { version = "3.3.1", optional = true }
 lewton = { version = "0.10", optional = true }
 minimp3_fixed = { version = "0.5.4", optional = true }
 symphonia = { version = "0.5.4", optional = true, default-features = false }
+# symphonia = { git = "https://github.com/roderickvd/Symphonia", branch = "perf/faster-seeking", optional = true }
 crossbeam-channel = { version = "0.5.15", optional = true }
 
 rand = { version = "0.9.0", features = ["small_rng", "os_rng"], optional = true }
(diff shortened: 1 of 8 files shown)


PetrGlad:
Can this be automatically tested?

yara-blue:
> As I commented over at [pdeljanov/Symphonia#340 (comment)](https://github.com/pdeljanov/Symphonia/pull/340#issuecomment-2927671333), there are legitimate reasons why a source could have a known `byte_len` yet not be **random-access** seekable.

from that comment, copied here for context:

>is_seekable implies support for random-access seeking. Sources that could have their byte length known, yet only be forward-seekable include pipes, buffers from stdin or a read-once network stream.

Writing this PR I assumed such *forward-seekable* only applications are too niche to be worth supporting in rodio. As I understood it then it was either support them and make the general seek case less usable.

> Then in try_seek we can still support seek, but only forward. When a users tries to seek backward, we could catch this error and propagate it with a help text "set byte_len to enable random access seeking".

I was unaware this is an option. That seems like a good solution.

@roderickvd would you mind taking this out of my hands? My *symphonia* knowledge is minimal, I do not have the time to dive into it and as we have seen this issue is resisting my quick attempts at quickly throwing something together.

roderickvd:
Sure, I'll try and make some time for it this week. With this and #731 we can hopefully put a release out.

roderickvd:
Alternative proposed in #745

roderickvd:
Superseded by #745

PR #745, opened 2025-06-03
Improve seeking behavior and error handling

### LINKED ISSUE #740: Potential Regression: `try_seek` Panicking When Seeking Backwards
Switching from rodio `0.20.1` to the main branch (with symphonia-all enabled), I encountered an issue where try_seek produces a `SymphoniaDecoder(BaseSeek(SeekError(ForwardOnly)))` error when attempting to seek backwards. This behavior was not present in rodio `0.20.1`. I've tested with multiple file formats (flac, m4a, and mp3), all of which produce the same error when seeking to a previous time.

Additional Testing: When using **music.ogg** from the assets folder, the **try_seek** function does not seek to the specified position. Instead, it appears to seek to a slightly earlier position and then continues playback as if the seek was relative.

### Reproduction Build

[rodio_seek_err.zip](https://github.com/user-attachments/files/20189003/rodio_seek_err.zip)

roderickvd:
I think I've found the compromise I like best:

- Automatically enable `is_seekable` when `byte_len` is set
- Add helpful error message for random access seeking without `byte_len`
- Clarify documentation for builder methods with default states
- Test forward-only vs. random access seeking scenarios
- Ensure MP4s open without depending on `is_seekable`

I also tried setting `is_seekable` to `true` by default, but that only increases random access seeking weirdness without `byte_len` set. Which is why I think this is the sweet spot of compromises.

Mostly fixes #740
Supersedes #744

I say "mostly" fixes because `symphonia-isomp4` can still report it's successfully seeked backwards, when in fact it hasn't without `byte_len` set. In Symphonia v0.5 the MP4 support is working, but with a number of bugs that are targeted for some v0.6 release. In the meantime, I propose we regard MP4 as a "second tier" target and we don't go too far out of a way to fix upstream issues here.

### DIFF
--- a/tests/seek.rs
+++ b/tests/seek.rs
@@ -1,6 +1,8 @@
 #![allow(dead_code)]
 #![allow(unused_imports)]
 
+#[cfg(feature = "symphonia-mp3")]
+use rodio::{decoder::symphonia, source::SeekError};
 use rodio::{ChannelCount, Decoder, Source};
 use rstest::rstest;
 use rstest_reuse::{self, *};
@@ -239,6 +241,7 @@ fn seek_does_not_break_channel_order(
     }
 }
 
+#[cfg(feature = "symphonia-mp3")]
 #[test]
 fn random_access_seeks() {
     // Decoder::new::<File> does *not* set byte_len and is_seekable
@@ -249,7 +252,12 @@ fn random_access_seeks() {
         "forward seek should work without byte_len"
     );
     assert!(
-        decoder.try_seek(Duration::from_secs(1)).is_err(),
+        matches!(
+            decoder.try_seek(Duration::from_secs(1)),
+            Err(SeekError::SymphoniaDecoder(
+                symphonia::SeekError::RandomAccessNotSupported,
+            ))
+        ),
         "backward seek should fail without byte_len"
     );
 
--- a/tests/seek.rs
+++ b/tests/seek.rs
@@ -273,10 +273,7 @@ fn random_access_seeks() {
     );
 }
 
-fn second_channel_beep_range<R: rodio::Source>(source: &mut R) -> std::ops::Range<usize>
-where
-    R: Iterator<Item = f32>,
-{
+fn second_channel_beep_range<R: rodio::Source>(source: &mut R) -> std::ops::Range<usize> {
     let channels = source.channels() as usize;
     let samples: Vec<f32> = source.by_ref().collect();
 
--- a/tests/seek.rs
+++ b/tests/seek.rs
@@ -239,7 +239,36 @@ fn seek_does_not_break_channel_order(
     }
 }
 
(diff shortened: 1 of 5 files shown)


roderickvd:
@dvdsk if you have time, I feel we should slip this one into v0.21 as well.

dvdsk (commented):
I really like it! It clearly documents why things have to be (as complicated) as they are while simultaneously setting the defaults so that the average user will get a good result without looking at the docs.

tests/seek.rs
```diff
@@ -175,6 +175,32 @@ fn seek_does_not_break_channel_order(
...
+fn random_access_seeks() {
+    // Decoder::new::<File> does *not* set byte_len and is_seekable
+    let mp3_file = std::fs::File::open("assets/music.mp3").unwrap();
+    let mut decoder = Decoder::new(mp3_file).unwrap();
+    assert!(
+        decoder.try_seek(Duration::from_secs(2)).is_ok(),
+        "forward seek should work without byte_len"
+    );
+    assert!(
+        decoder.try_seek(Duration::from_secs(1)).is_err(),
```
yara-blue: This could check if the error returned matches what we expect. You might need `matches!` for that as I do not think symphonia error implements Eq.

dvdsk (approved):
one suggestion, feel free to ignore it though.

yara-blue:
> @dvdsk if you have time, I feel we should slip this one into v0.21 as well.

We will, documentation made this an easy review :)

If you made the branch on the rodio repo I would have solved that merge conflict for you :stuck_out_tongue_closed_eyes:

tests/seek.rs
roderickvd: Done in https://github.com/RustAudio/rodio/pull/745/commits/87f15d2e4227114d694be7becbeb6e13b228bc27.

roderickvd:
> If you made the branch on the rodio repo I would have solved that merge conflict for you 😝

What's the policy, you want me to do that (branch on the repo) instead going forward?

yara-blue:
> What's the policy, you want me to do that (branch on the repo) instead going forward?

I branch on the repo for three reasons:
- other maintainers can jump in and help out
- could help to see what is going on 
- if I have to take a step back its a little easier to take over

There are probably just as many reasons to branch on a fork. We could formulate a policy, I don't think 
there is one now? On the other hand its not really that important. I just wanted to make sure you 
knew that there is no objection from me to creating branches on this repo.

Semi-related: governance of Rodio. 

I think of us as a flat organization, none one has more say then anyone else. I would not know who 
would ultimately be in charge otherwise... maybe Pierre since he made rodio...

Of course on my CV I am *Project lead of Rodio* :smile:

roderickvd:
Thanks, I'll just branch here then. No need for any formalities at this point.

PR #746, opened 2025-06-05
Make accidentally dropping OutputStream impossible

dvdsk:
This adds a drop bomb to OutputStream. Dropping a OutputStream will panic
(unless we are dropping as part of unwinding). Technically OutputStream
is now a run time enforced linear type.

To get rid of a `OutputStream` you now have to call `OutputStream::close`.

For a (worse in my opinion) alternative see: #747

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,7 +18,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ed7572b7ba83a31e20d1b48970ee402d2e3e0537dcfe0a3ff4d6eb7508617d43"
 dependencies = [
  "alsa-sys",
- "bitflags 2.9.0",
+ "bitflags 2.9.1",
  "cfg-if",
  "libc",
 ]
@@ -35,9 +35,9 @@ dependencies = [
 
 [[package]]
 name = "anstyle"
-version = "1.0.10"
+version = "1.0.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "55cc3b69f167a1ef2e161439aa98aed94e6028e5f9a59be9a6ffb47aef1651f9"
+checksum = "862ed96ca487e809f1c8e5a8447f6ee2cf102f846893800b20cebdf541fc6bbd"
 
 [[package]]
 name = "approx"
@@ -74,21 +74,21 @@ checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
 name = "bitflags"
-version = "2.9.0"
+version = "2.9.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5c8214115b7bf84099f1309324e63141d4c5d7cc26862f97a0a857dbefe165bd"
+checksum = "1b8e56985ec62d17e9c1001dc89c88ecd7dc08e47eba5ec7c29c7b5eeecde967"
 
 [[package]]
 name = "bumpalo"
-version = "3.17.0"
+version = "3.18.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1628fb46dfa0b37568d12e5edd512553eccf6a22a78e8bde00bb4aed84d5bdbf"
+checksum = "793db76d6187cd04dff33004d8e6c9cc4e05cd330500379d2394209271b4aeee"
 
 [[package]]
 name = "bytemuck"
-version = "1.22.0"
+version = "1.23.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b6b1fc10dbac614ebc03540c9dbd60e83887fda27794998c6528f1782047d540"
+checksum = "5c76a5792e44e4abe34d3abf15636779261d45a7450612059293d1d2cfc63422"
 
 [[package]]
 name = "byteorder"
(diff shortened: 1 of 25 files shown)


roderickvd:
> For a (worse in my opinion) alternative see: #747

I agree that maintaining our own counter sucks, like in #747. But so is pulling in another dependency, kinda... it's late, but could we use [`ManuallyDrop`](https://doc.rust-lang.org/beta/std/mem/struct.ManuallyDrop.html) from `std` instead?

Or something like this? I'm actually thinking: isn't `Arc` designed precisely for this? But I thought we had already wrapped it in an `Arc` before. I need to go to bed 😄 but here's a rough draft with a wrapper type:

```rust
use std::sync::Arc;

// New internal type to wrap the actual stream
struct StreamHandle {
    stream: cpal::Stream,
}

pub struct OutputStream {
    mixer: Mixer,
    _stream: Arc<StreamHandle>,
}

impl OutputStream {  
    pub fn mixer(&self) -> &Mixer {
        &self.mixer
    }
    
    pub fn connect_sink(&self) -> Sink {
        Sink::connect_new(self.mixer())
    }
}
```

Mixer:

```rust
pub struct Mixer {
    _stream: Arc<StreamHandle>,
}

impl Mixer {
    fn new(stream: Arc<StreamHandle>) -> Self {
        Self {
            _stream: stream,
        }
    }

    fn get_stream_ref(&self) -> Arc<StreamHandle> {
        self._stream.clone()
    }
}
```

Sink:

```rust
pub struct Sink {
    _stream: Arc<StreamHandle>,
}

impl Sink {
    pub fn connect_new(mixer: &Mixer) -> Sink {
        let stream_ref = mixer.get_stream_ref();
        
        Sink {
            _stream: stream_ref,
        }
    }
}
```

yara-blue:
> I agree that maintaining our own counter sucks, like in #747. But so is pulling in another dependency, kinda... it's late, but could we use [`ManuallyDrop`](https://doc.rust-lang.org/beta/std/mem/struct.ManuallyDrop.html) from `std` instead?

Manually drop does something else, the drop bomb is crate is only a few lines so we can copy those over instead. The author is highly trusted and it made the code clean so I just added it as dependency.

> Or something like this? I'm actually thinking: isn't Arc designed precisely for this? But I thought we had already wrapped it in an Arc before. I need to go to bed 😄 but here's a rough draft with a wrapper type:

If only we could, unfortunately `cpal::Stream` is `!Send`. If we where to wrap it in an `Arc` our mixer's could no longer travel between threads. That is why #747 spawns a thread just to create a `cpal::Stream` in and keep it there.

Another issue is that its possible to drop all sinks and mixers and there should still be audio playing. Though that could probably be helped by adding a "`counter`" to the mixer source instead of the mixer handle....

roderickvd:
Argh, what a pain. Well, if this is the lesser of evils, so be it.

yara-blue:
> Argh, what a pain. Well, if this is the lesser of evils, so be it.

We could get #747 to work "perfectly" by moving the counter to the source. But we would have to introduce extra logic to deal with stuff like endless queue's. (We would need to track if the queue's handle has dropped and then (when the queue is empty) lower the counter).

It would still be kinda ugly due to the thread that is simply there to store the `cpal::Stream` :disappointed:.

Still in favor of this PR, though I really don't like making every downstream user add a custom drop implementations to whatever they are storing our `OutputStream` in.....

roderickvd:
> Still in favor of this PR, though I really don't like making every downstream user add a custom drop implementations to whatever they are storing our `OutputStream` in.....

The other option is of course to do nothing, and have users retain the stream in their struct.

yara-blue:
> The other option is of course to do nothing, and have users retain the stream in their struct.

maybe there is a middle ground: We emit a warning on drop unless close is called. Its not as disrupting as a panic but it could still help out new users.

roderickvd:
> maybe there is a middle ground: We emit a warning on drop unless close is called. Its not as disrupting as a panic but it could still help out new users.

In the case of the examples as they are today, would that mean that a warning is emitted for most of them?
I can imagine that a user wouldn't care when it just works with one file and exits, but a warning may surprise them.

yara-blue:
> In the case of the examples as they are today, would that mean that a warning is emitted for most of them?

No not if we call `.close()` on the stream before the end of the function. Instead of a drop bomb we have a drop shout. We can emit the shout at the `debug` log level for those using tracing or log, that its pretty non-disruptive.

roderickvd:
OK, let's do that.

yara-blue:
> OK, let's do that.

I'll get on that tomorrow and then afterwards see if I can prep the next release

PetrGlad:
Could it be an option to provide drop diagnostics instead? Drop may show the stack in debug mode, in lower log level settings this could be only a message mentioning the operation.

PetrGlad:
Another alternative - to recommend keeping the output stream in a static cell, maybe?

yara-blue:
> Could it be an option to provide drop diagnostics instead?

this is the plan now, we will print on drop unless close() was called first.

> Another alternative - to recommend keeping the output stream in a static cell, maybe?

Dropping the stream before it should close is mostly an issue about not reading the docs. If you read those you will see you should not drop the stream before you are done playing sound. 

I do not like users having to read the docs in detail (and remember it all!). Rodio should be really easy to use. No sound (without any warning) because you dropped a variable early makes Rodio hard to use. This PR adds a crash which is impossible to miss. The proposal is to make it a warning/diagnostic.

yara-blue:
lets remove close and the drop bomb, instead we always print something when the outputstream is dropped. We add a function to outputstream which one can use to disable the print. And we call that out from the print msg.

yara-blue:
closed in favor of https://github.com/RustAudio/rodio/pull/761

PR #747, opened 2025-06-05
Only close cpal::Stream once all mixers/sinks drop

dvdsk:
**This is an alternative to #746**

This resolves a common footgun in rodio, that is: accidentily dropping the stream handle causing all audio to stop. This can be hard to debug.

To solve that we make `cpal::Stream` undroppable by the user. Instead we drop it for the user once all mixers/sinks that are connected to the `cpal::Stream` are dropped. The only way to make `cpal::Stream` undroppable is to hold it in a sleeping thread. This is because `cpal::Stream` is neither send nor sync. If it where we could just store it on the heap in an `Arc` shared by all sinks and streams.

The `cpal::Stream` is now created on a seperate thread. After creation that thread sleeps until the OutputStreamTracker reaches a count of zero. The OutputStreamTracker is shared between all mixers and sinks that are connected to a `cpal::Stream`.

This design is not perfect, its still quite easy to drop the `cpal::Stream`. Simply take a normal mixer (not the `OutputMixer` introduced by this PR) and connect it to an `OutputMixer`. Then use the normal mixer instead of the `OutputMixer` and drop the `OutputMixer`.

An alternative, far simpler solution could be a drop bomb

### DIFF
--- a/examples/basic.rs
+++ b/examples/basic.rs
@@ -12,7 +12,7 @@ fn main() -> Result<(), Box<dyn Error>> {
     let beep1 = {
         // Play a WAV file.
         let file = std::fs::File::open("assets/beep.wav")?;
-        let sink = rodio::play(mixer, BufReader::new(file))?;
+        let sink = rodio::play(mixer.clone(), BufReader::new(file))?;
         sink.set_volume(0.2);
         sink
     };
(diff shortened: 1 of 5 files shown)


yara-blue:
Examples break, ill fix those if we decide this is the approach we go with.

PR #749, opened 2025-06-06
Allow accessing chosen OutputStreamConfig from OutputStream

will3942:
When opening an OutputStream there is no way to access the configuration associated with that Stream once it has been created.

The configuration ultimately used by the OutputStream may be different to what has been supplied (in the case when using  `OutputStreamBuilder::open_stream_or_fallback`).

This allows the user to access key configuration attributes that are associated with the OutputStream after build.

We specifically use this for accessing the SampleRate and BufferSize of the OutputStream at runtime.


roderickvd (commented):
I like it. You may want to add an entry to the changelog as well.

will3942:
@roderickvd Good shout - I've added those now. All good from my side, feel free to merge when you're happy & ready.

dvdsk (approved):

roderickvd:
Thanks!

PR #750, opened 2025-06-06
feat: add compressor effect

### LINKED ISSUE #748: Compressor effect
Is anyone working on adding a compressor effect? If not, I'd like to work on it.

setoelkahfi:
Fixes #748 

Add a dynamic range compressor. 

Run the examples:

```bash
$ cargo run --example compressor
$ cargo run --example compressor_wav
$ cargo run --example compressor_wav_alternate
```

### DIFF
--- a/examples/compressor_wav.rs
+++ b/examples/compressor_wav.rs
@@ -1,0 +1,18 @@
+use rodio::Source;
+use std::error::Error;
+
+fn main() -> Result<(), Box<dyn Error>> {
+    let stream_handle = rodio::OutputStreamBuilder::open_default_stream()?;
+    let sink = rodio::Sink::connect_new(stream_handle.mixer());
+
+    let file = std::fs::File::open("assets/music.wav")?;
+    let source = rodio::Decoder::try_from(file)?.compressor(0.05, 10.0, 0.001, 0.05);
+
+    sink.append(source);
+
+    println!("Playing music.wav with compressor until finished...");
+    sink.sleep_until_end();
+    println!("Done.");
+
+    Ok(())
+}
--- a/src/source/compressor.rs
+++ b/src/source/compressor.rs
@@ -1,0 +1,153 @@
+use std::time::Duration;
+
+use super::SeekError;
+use crate::common::{ChannelCount, SampleRate};
+use crate::Source;
+
+/// Internal function that builds a `Compressor` object.
+pub(crate) fn compressor<I>(
+    input: I,
+    threshold: f32,
+    ratio: f32,
+    attack: f32,
+    release: f32,
+) -> Compressor<I>
+where
+    I: Source,
+{
+    Compressor {
+        input,
+        threshold,
+        ratio,
+        attack,
+        release,
+        gain: 1.0,
+        envelope: 0.0,
+    }
+}
+
+/// Filter that applies a compressor effect to the source.
+#[derive(Clone, Debug)]
+pub struct Compressor<I> {
+    input: I,
+    threshold: f32,
+    ratio: f32,
+    attack: f32,
+    release: f32,
+    gain: f32,
+    envelope: f32,
+}
+
+impl<I> Compressor<I> {
+    /// Set the compression threshold.
+    #[inline]
+    pub fn set_threshold(&mut self, threshold: f32) {
+        self.threshold = threshold;
+    }
+
+    /// Set the compression ratio.
+    #[inline]
+    pub fn set_ratio(&mut self, ratio: f32) {
+        self.ratio = ratio;
+    }
+
+    /// Set the attack time (seconds).
+    #[inline]
+    pub fn set_attack(&mut self, attack: f32) {
+        self.attack = attack;
+    }
+
+    /// Set the release time (seconds).
+    #[inline]
+    pub fn set_release(&mut self, release: f32) {
+        self.release = release;
+    }
+
+    /// Returns a reference to the inner source.
+    #[inline]
+    pub fn inner(&self) -> &I {
+        &self.input
+    }
+
+    /// Returns a mutable reference to the inner source.
+    #[inline]
+    pub fn inner_mut(&mut self) -> &mut I {
+        &mut self.input
+    }
+
+    /// Returns the inner source.
+    #[inline]
+    pub fn into_inner(self) -> I {
+        self.input
+    }
+}
+
+impl<I> Iterator for Compressor<I>
+where
+    I: Source,
+{
+    type Item = I::Item;
+
+    #[inline]
+    fn next(&mut self) -> Option<Self::Item> {
+        let sample = self.input.next()?;
+        let abs_sample = sample.abs();
+
+        // Envelope follower (simple peak detector)
+        let sr = self.input.sample_rate() as f32;
+        let attack_coeff = f32::exp(-1.0 / (self.attack * sr).max(1.0));
+        let release_coeff = f32::exp(-1.0 / (self.release * sr).max(1.0));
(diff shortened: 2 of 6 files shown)


examples/compressor_wav.rs
yara-blue, line 1: We might not need 3 examples, I think the `sine` and `wav` one show off the same thing right?

setoelkahfi: Yeah. The `wav` one is easier for ear training, in my opinion. I can remove the `sine` example.

dvdsk (commented):
I'm not sure about the API for compressor. It follows what we have for sources however I find it hard to know which argument is what. Maybe we can do something with a builder pattern?

```rust
let compressed = SineWave::new(440.0)
        .amplify(0.5)
        .compressor()
        .with_threshold(0.01)
        .with_ratio(20.0)
        .with_attack(0.0005)
        .with_release(0.02)
       .take_duration(Duration::from_secs(3));
```

Now the issue with the above is that its not clear that `with_threshold` is a function on `compressor`. Maybe a compressor config?

 ```rust
let compressed = SineWave::new(440.0)
        .amplify(0.5)
        .compressor(CompressorSettings::new()
            .with_threshold(0.01)
            .with_ratio(20.0)
            .with_attack(0.0005)
            .with_release(0.02)
        )
       .take_duration(Duration::from_secs(3));
```

Definitely better, though not a big fan of the name CompressorSettings. Now we can get rid of that too but I do not think we will like the result. Lets see:

 ```rust
let compressed = SineWave::new(440.0)
        .amplify(0.5)
        .compressor(|c|
             c.set_threshold(0.01);
             c.set_ratio(20.0);
             c.set_attack(0.0005);
             c.set_release(0.02);
        )
       .take_duration(Duration::from_secs(3));
```

A bit too much magic right? The `c` variable in the closure would be a `Compressor` struct instance and the `.compressor` method on source takes as argument a function that mutates a `Compressor` instance. The implementation of the `.compressor` method would create a default instance of `Compressor` call the provided function to configure the instance and then return the now configured instance.

We could, provide a general mechanism for this (as another source). Similar too periodic callback but only called once. Lets call it `configure_source`. That would look like this:

 ```rust
let compressed = SineWave::new(440.0)
        .amplify(0.5)
        .compressor()
        .configure_source(|c|
             c.set_threshold(0.01);
             c.set_ratio(20.0);
             c.set_attack(0.0005);
             c.set_release(0.02);
        )
       .take_duration(Duration::from_secs(3));
```

Lots of options, I have preference however I am curious what you all think.

**Comment by yara-blue** [comment]
we should also document the normal range for these parameters and what a good start point is. Take a look at the docs for AGC for inspiration: https://github.com/RustAudio/rodio/blob/5df626a204566dbc8713b3c55bedc9124d21503a/src/source/mod.rs#L283

**Comment by yara-blue** [comment]
> I'm not sure about the API for compressor. It follows what we have for sources however I find it hard to know which argument is what. Maybe we can do something with a builder pattern?

AGC does the same (API wise), so we might also want to address that if we go for a similar API here.

File: examples/compressor_wav.rs
Comment (roderickvd) [reply examples/compressor_wav.rs:1]: I was thinking the same about the distortion examples?

**Comment by roderickvd** [comment]
Great work proposing these new DSPs @setoelkahfi!

> Lots of options, I have preference however I am curious what you all think.

I think passing a `CompressorSettings` is the clearest. Also it makes most sense to use with a `with_foo` builder syntax.

> we should also document the normal range for these parameters and what a good start point is.

I'd be in favor of also having `CompressorSettings::default`.

> AGC does the same (API wise), so we might also want to address that if we go for a similar API here.

I dislike the AGC API for passing Durations as seconds in `f32`. It would be more idiomatic to at least use `Duration`.

Further down the line, maybe as part of the future Firewheel integration that @dvdsk is working on, it'd be nice to also move the decibels from `f32` to its own `Decibel` type. That would allow for nice conversions and arithmetic.

File: src/source/compressor.rs [hunk 1]
```diff
@@ -0,0 +1,153 @@
...
+    type Item = I::Item;
+
+    #[inline]
+    fn next(&mut self) -> Option<Self::Item> {
+        let sample = self.input.next()?;
+        let abs_sample = sample.abs();
+
+        // Envelope follower (simple peak detector)
+        let sr = self.input.sample_rate() as f32;
+        let attack_coeff = f32::exp(-1.0 / (self.attack * sr).max(1.0));
```
roderickvd, line 98: These should be cached.

src/source/compressor.rs
```diff
@@ -0,0 +1,153 @@
...
+
+impl<I> Iterator for Compressor<I>
+where
+    I: Source,
+{
+    type Item = I::Item;
+
+    #[inline]
+    fn next(&mut self) -> Option<Self::Item> {
+        let sample = self.input.next()?;
```
roderickvd, line 93: For stereo or multi-channel audio, the samples are interleaved. Tracking a single `gain` for all channels can work, but may cause pumping and stereo imaging shifting when one channel is above the threshold and the other is not.

yara-blue:
> I've got a limiter done and optimized here, I just didn't get around to submitting it to Rodio: https://github.com/roderickvd/pleezer/blob/main/src/normalize.rs.
> 
> Note that this is a limiter with a soft knee, whereas your proposal seems to be a compressor with a hard knee. What is your design goal?
> 
> While I'm not married to it, please consider the merits of the limiter linked above and think about which implementation may be the best going forward. This limiter:
> 
>     * uses soft-knee processing so may sound more musical
> 
>     * works in the dB domain so sound quality may be better
> 
>     * has per-channel envelope detection with coupled gain reduction
> 
>     * implements two-stage envelope detection for better transient handling
> 
>     * has been profiled and optimized for performance already (< 2% CPU on RPi 4B)
> 
> 
> Yours of course is easier and overall faster with single-state envelope detection in the linear domain. Sometimes easier is better. Sometimes we won't mind some extra CPU cycles if that's worth the sound quality.
> 
> If you want, feel absolutely free to take and relicense it under MIT or Apache 2.0.

That looks very nice, would be lovely to have as a part of Rodio, thank you! @setoelkahfi How would you feel about merging rodericks code with the proposed API (I also favor the `CompressorSettings`, especially combined with a default impl).

setoelkahfi:
Hey @dvdsk I overestimated my time commitment for this. Slow response mode.

yara-blue:
> Hey @dvdsk I overestimated my time commitment for this. Slow response mode.

No worries, take all the time you need. And also feel free to drop it, its a hobby, its gotta stay fun. If you decide to drop us let us know.

roderickvd:
> > I've got a limiter done and optimized here, I just didn't get around to submitting it to Rodio: https://github.com/roderickvd/pleezer/blob/main/src/normalize.rs.
>
> That looks very nice, would be lovely to have as a part of Rodio, thank you! @setoelkahfi How would you feel about merging rodericks code with the proposed API (I also favor the `CompressorSettings`, especially combined with a default impl).

[...]

> Hey @dvdsk I overestimated my time commitment for this. Slow response mode.

Ported it at #751

yara-blue:
closing this in favor of #751. I do think this was a great starting point, you got the ball rolling and now we can finish things up in #751.

setoelkahfi:
> closing this in favor of #751. I do think this was a great starting point, you got the ball rolling and now we can finish things up in #751.

I think that's a good idea. I could try the branch and see if you can find an issue.

yara-blue:
> I think that's a good idea. I could try the branch and see if you can find an issue.

please do! A second look is always appreciated!

PR #751, opened 2025-06-08
feat: add audio peak limiting with configurable settings

roderickvd:
Provides a new LimitSettings API and limiter filter to prevent audio peaks from exceeding a threshold. Supports soft-knee limiting, per-channel detection, configurable attack/release, and efficient processing for mono, stereo, and multi-channel audio. Includes comprehensive tests and usage examples.


src/source/amplify.rs
```diff
@@ -14,7 +14,15 @@ where
 
 /// Internal function that converts decibels to linear
 pub(super) fn to_linear(decibels: f32) -> f32 {
```
roderickvd: Should we move these methods to `mod.rs` or some utility module?

yara-blue:
I love the settings example showing some analysis! It shows a different application of rodio then playback with effects. 

Could we augment that example with 2 of the examples from #750? I would say the "compressor_wav" and "compressor_wav_alternate.rs". We could rename those to "compressor_basic" and "compressor_changing".

examples/limit_settings.rs
```diff
@@ -0,0 +1,173 @@
...
+    // Example 1: Using default settings
+    println!("Example 1: Default LimitSettings");
+    let default_settings = LimitSettings::default();
+    println!("  Threshold: {} dB", default_settings.threshold);
+    println!("  Knee width: {} dB", default_settings.knee_width);
+    println!("  Attack: {:?}", default_settings.attack);
+    println!("  Release: {:?}", default_settings.release);
+    println!();
+
+    // Example 2: Using builder pattern for custom settings
```
yara-blue: nitpick:
The std user the builder pattern a lot but they never refer to it as a pattern. I think it is because the word pattern has some negative associations for some. 

Example: https://doc.rust-lang.org/nightly/std/fs/struct.OpenOptions.html this mentions a builder but not a pattern

examples/limit_settings.rs
```diff
@@ -0,0 +1,173 @@
...
+        .with_attack(Duration::from_millis(10))
+        .with_release(Duration::from_millis(50));
+
+    println!("  Threshold: {} dB", custom_settings.threshold);
+    println!("  Knee width: {} dB", custom_settings.knee_width);
+    println!("  Attack: {:?}", custom_settings.attack);
+    println!("  Release: {:?}", custom_settings.release);
+    println!();
+
+    // Example 3: Applying limiter to a sine wave with default settings
```
yara-blue: I do not think we need the comment (given the print)

examples/limit_settings.rs
```diff
@@ -0,0 +1,173 @@
...
+        "  Generated {} samples with custom settings",
+        custom_samples.len()
+    );
+    println!();
+
+    // Example 5: Different limiting scenarios
+    println!("Example 5: Comparing different limiting scenarios");
+
+    // Gentle limiting
+    let gentle_settings = LimitSettings::default()
```
yara-blue: similar here: `gentle_settings` could be renamed `gentle_limiting` and then the comment is no longer needed

examples/limit_settings.rs
```diff
@@ -0,0 +1,173 @@
...
+    let complex_chain = SineWave::new(220.0)
+        .amplify(1.5)                              // Add some gain
+        .limit(LimitSettings::default()            // Apply limiting with default + threshold
+            .with_threshold(-2.0))
+        .take_duration(Duration::from_millis(50));
+
+    let chained_samples: Vec<f32> = complex_chain.take(50).collect();
+    println!(
+        "  Generated {} samples from chained effects",
+        chained_samples.len()
```
yara-blue: You could also argue that this is showing off rodio's chaining and that is already done in other examples. Then given this example is on the long side we could skip this.

examples/limit_settings.rs
```diff
@@ -0,0 +1,173 @@
...
+    let target_linear = 10.0_f32.powf(strict_settings.threshold / 20.0);
+    let max_settled = test_samples[2000..]
+        .iter()
+        .fold(0.0f32, |acc, &x| acc.max(x.abs()));
+
+    println!(
+        "  {}dB threshold limiting results:",
+        strict_settings.threshold
+    );
+    println!("    Original max amplitude: {AMPLITUDE}");
```
yara-blue, line 140: again I love this, very nice!

src/source/amplify.rs
yara-blue: We have `src/math` it could fit there. Feel free to move it and rename it to `decibels_to_linear()`. But its also fine where it is right? this is always amplifying, sometimes a number sometimes a source.

src/source/amplify.rs
```diff
@@ -14,7 +14,15 @@ where
...
-    f32::powf(10f32, decibels * 0.05)
+    // ~3-4% faster than using `10f32.powf(decibels * 0.05)`,
```
yara-blue: hype :partying_face: :rocket:

src/source/amplify.rs
```diff
@@ -143,13 +169,120 @@ mod test {
 
     #[test]
     fn convert_decibels_to_linear() {
```
yara-blue: impressive tests! And thank you very much for the methodology description.

src/source/amplify.rs
```diff
@@ -143,13 +169,120 @@ mod test {
...
+    fn round_trip_conversion_accuracy() {
+        // Test that converting dB -> linear -> dB gives back the original value
+        let test_db_values = [-60.0, -20.0, -6.0, 0.0, 6.0, 20.0, 40.0];
+
+        for &original_db in &test_db_values {
+            let linear = to_linear(original_db);
+            let round_trip_db = to_db(linear);
+
+            let error = (round_trip_db - original_db).abs();
+            const MAX_ROUND_TRIP_ERROR: f32 = 16.0 * f32::EPSILON; // Measured max error: 8x ε (practical audio range), with 2x safety margin
```
yara-blue: I find the word "Measured" slightly confusing. Would be clearer to me without it. Though that could be my physics background biassing me (measure conjures images of measurement equipment for me)

src/source/limit.rs
```diff
@@ -0,0 +1,805 @@
+//! Audio peak limiting for dynamic range control.
+//!
+//! This module implements a feedforward limiter that prevents audio peaks from exceeding
+//! a specified threshold while maintaining audio quality. The limiter is based on:
+//! Giannoulis, D., Massberg, M., & Reiss, J.D. (2012). Digital Dynamic Range Compressor Design,
+//! A Tutorial and Analysis. Journal of The Audio Engineering Society, 60, 399-408.
+//!
+//! # What is Limiting?
```
yara-blue, line 8: This is sooooo useful to users new to audio! Thank you :)

src/source/limit.rs
```diff
@@ -0,0 +1,805 @@
...
+/// ```
+///
+/// ## Common Adjustments
+///
+/// ```rust
+/// use rodio::source::LimitSettings;
+/// use std::time::Duration;
+///
+/// // More headroom for dynamic content
+/// let conservative = LimitSettings::default()
```
yara-blue, line 118: we could make this presets. You would then have  `LimitSettings::dynamic_content()`, `LimitSettings::streaming()` & `LimitSettings::default()`.

src/source/limit.rs
```diff
@@ -0,0 +1,805 @@
...
+/// let broadcast = LimitSettings::default()
+///     .with_knee_width(2.0)                    // Narrower knee for firmer limiting
+///     .with_attack(Duration::from_millis(3))   // Faster attack
+///     .with_release(Duration::from_millis(50)); // Faster release
+/// ```
+#[derive(Debug, Clone)]
+pub struct LimitSettings {
+    /// Level where limiting begins (dB, must be negative)
+    pub threshold: f32,
+    /// Range over which limiting gradually increases (dB)
```
yara-blue: I get why you would like a decibel type. Would be pretty useful. We could also add doc comments to that explaining bB vs dBFS. Maybe next release?

src/source/limit.rs
```diff
@@ -0,0 +1,805 @@
...
+/// # Returns
+///
+/// Smoothing coefficient in the range [0.0, 1.0] for use in exponential filters
+#[must_use]
+fn duration_to_coefficient(duration: Duration, sample_rate: SampleRate) -> f32 {
+    f32::exp(-1.0 / (duration.as_secs_f32() * sample_rate as f32))
+}
+
+#[cfg(test)]
+mod tests {
```
yara-blue, line 1143: These could all be integration tests, I prefer integration over unit but choose what you prefer.

src/source/limit.rs
```diff
@@ -0,0 +1,805 @@
...
+            .fold(0.0f32, |acc, &x| acc.max(x.abs()));
+        assert!(
+            max_sample < 0.8,
+            "ALL samples should be well below 1.0: max={:.3}",
+            max_sample
+        );
+    }
+
+    #[test]
+    fn test_settings_api() {
```
yara-blue: I don't know if this test still provides a lot given the feature now works. I cant think of a scenario where this test breaks and we are not changing the API.

src/source/limit.rs
```diff
@@ -0,0 +1,805 @@
...
+///
+/// let source = SineWave::new(440.0).amplify(2.0);
+/// let limited = source.limit(LimitSettings::default().with_threshold(-6.0));
+/// ```
+///
+/// # Type Parameters
+///
+/// * `I` - The input audio source type
+#[derive(Clone, Debug)]
+pub enum Limit<I>
```
yara-blue: we could have:

```rust
pub struct Limit(LimitInner);

enum LimitInner {
  Mono(..)
  Stereo(..)
  Multi....
}
```

Then `LimitMono`, `LimitStereo` and `MultiChannel` do not need to be pub.

src/source/limit.rs
```diff
@@ -0,0 +1,805 @@
...
+    type Item = I::Item;
+
+    /// Provides the next processed sample.
+    ///
+    /// Routes processing to the appropriate channel-specific implementation:
+    /// * Mono: Direct single-channel processing
+    /// * Stereo: Optimized two-channel processing
+    /// * `MultiChannel`: Generic multi-channel processing
+    #[inline]
+    fn next(&mut self) -> Option<Self::Item> {
```
yara-blue, line 1019: I hate to ask this but..... what if the channel count changes in the next span? (I hate spans now)

src/source/amplify.rs
roderickvd: I moved and renamed them to `src/math.rs` in b1497f8. It doesn't feel right to me that we were adding a utility function `to_db` to a module that didn't use it. I also made them public, because they are useful as general-purpose audio functions.

examples/limit_settings.rs
roderickvd: Removed in b1497f8.

examples/limit_settings.rs
roderickvd: Removed this and more superfluous comments in b1497f8.

examples/limit_settings.rs
roderickvd: Removed in b1497f8.

examples/limit_settings.rs
roderickvd: Removed in b1497f8.

src/source/amplify.rs
roderickvd: Removed in de9deb832c062875c0e1eef7c504bbdf9d2ec6ad.

src/source/limit.rs
roderickvd: Added in b1497f8 with some additional presets: 
```markdown
* [`LimitSettings::default()`] - General-purpose limiting (-1 dBFS, balanced)
* [`LimitSettings::dynamic_content()`] - Music and sound effects (-3 dBFS, transparent)
* [`LimitSettings::broadcast()`] - Streaming and voice chat (fast response, consistent)
* [`LimitSettings::mastering()`] - Final production stage (-0.5 dBFS, tight peak control)
* [`LimitSettings::gaming()`] - Interactive audio (-3 dBFS, responsive dynamics)
* [`LimitSettings::live_performance()`] - Real-time applications (ultra-fast protection)
```

src/source/limit.rs
roderickvd: Added documentation explaining dB vs. dBFS in b1497f8.

src/source/limit.rs
roderickvd: Moved to integration tests in b1497f8.

src/source/limit.rs
roderickvd: Removed in b1497f8.

src/source/limit.rs
roderickvd: Done in b1497f8.

src/source/limit.rs
roderickvd: First and foremost: I bet that in 99.9% of cases, the limiter will be created by doing `Decoder::try_from(file1).limiter(...)`. Let's say `file1` is a 44.1 kHz stereo music file. It's extremely unlikely (and currently not well-supported) that the channel count (or sample rate, for that matter) changes within `file1`. Now let's say we have `file2` which is a 48 kHz mono podcast episode. Creating a new limiter with `Decoder::try_from(file2).limiter(...)` will then be fine.

Now for the unusual case of dynamic channel changes within one source:

You've probably spotted the performance optimizations for the mono and stereo case. And even the multi-channel case, which is the generic case that could equally handle mono or stereo, there's this optimization where the length of the vectors is determined at creation instead of during each iteration.

Theoretically speaking, this means that the limiter should be recreated or the imaging may suffer. Imaging here means that there may be audible pumping or shifting balance when one channel is much louder than the other.

Practically speaking, the limiter will continue to work. For music that's not hard-panned, it may not even matter much. For multi-channel gaming, probably more so.

Whatever we think of in the future of dealing with spans, should equally be implemented here.

src/source/amplify.rs
yara-blue: > they are useful as general-purpose audio functions

I don't follow exactly. We have different amplify methods on source for the different scales. How does converting a single float to another float help audio? Do you mean they are useful as parts for custom sources?

roderickvd: As a user of this library, I frequently run into the need to convert one into the other and display it to the end user. For example: Deezer, Qobuz and Spotify all send linear values and I want to print their dB equivalents on screen. Easy peasy when Rodio has them public.

dvdsk (approved):

roderickvd:
Thank you for your review.

> Could we augment that example with 2 of the examples from #750? I would say the "compressor_wav" and "compressor_wav_alternate.rs". We could rename those to "compressor_basic" and "compressor_changing".

I did not add anything like `_changing`. There currently is no way to change the limiter settings after it was created, and I don't really see a reason to do so.

src/source/limit.rs
yara-blue: I completely agree, its quite rare and probably not supported properly by other rodio parts currently. We might not even properly address spans until the new rodio audio core is done (or something else depending on what we decide in the audio ecosys effort). 

Taking all that into account it might be best to add what you wrote above as comment to the `next` function. Would you agree?

yara-blue:
> There currently is no way to change the limiter settings after it was created, and I don't really see a reason to do so

Then lets not do that :+1:

src/source/limit.rs
roderickvd: Documented in [77a246b](https://github.com/RustAudio/rodio/pull/751/commits/77a246b824cda184bc76e36ac967aa4837d5102d).

yara-blue:
ready to merge now right?

roderickvd:
👍 I will squash merge now.

PetrGlad:
I may have some confusion regarding terminology, how such limiter is different to a compressor? Compressor also has attack/release time parameters, with perhaps only difference that below-the-knee-threshold samples are also affected.
Is the difference that limiter works with amplitudes, not loudness?

roderickvd:
A limiter is also a compressor, a particular type of it. Any compressor has a threshold (amplitude) beyond which it starts reducing dynamic range with a certain ratio. For a limiter, that ratio is very high or infinite.

PR #752, opened 2025-06-09
fix: minimp3 decoder compilation

roderickvd:
A last-minute one as I was preparing another PR for a release beyond 0.21: the `minimp3` decoder is broken. This PR fixes that and adds a CI check to prevent regression.


dvdsk (approved):

yara-blue:
Thanks! Are the other decoders tested by default?

roderickvd:
> Thanks! Are the other decoders tested by default?

Yes, because they are enabled by default.

PR #753, opened 2025-06-09
feat: change default decoder to Symphonia and update feature flags

roderickvd:
Not sure how controversial this may be? Probably for a next version.

Pros:
- seeking support
- unified architecture
- enhanced format support
- better error handling
- works in floating point
- support for metadata and channel maps

Cons:
- not as minimal as the other decoders
- probably biggest one: MPL license is weak copyleft

To counter that licensing point, the other decoders are still available as optional features.

Further changes:
- Cleaned up and documented `Cargo.toml`
- Added MP4 to the default features
- Added CAF/MKV/MP1/MP2 as optional features
- Changed ADPCM to become an optional feature (used to be enabled with `symphonia-wav`).


yara-blue:
> Not sure how controversial this may be? Probably for a next version.

We should check with Bevy and some other Rodio users. @will3942 whats your opinion on this change?

will3942:
> > Not sure how controversial this may be? Probably for a next version.
> 
> 
> 
> We should check with Bevy and some other Rodio users. @will3942 whats your opinion on this change?

Thanks for flagging.

A less permissive license is always less desirable for us. MPL isn't a problem though and we already use Symphonia.

yara-blue:
Bevy runs with default features false so it should not be an issue for them (as they have zero decoders enabled by default)

yara-blue:
Could you add a section to the upgrade guide? Do we need to say something in the release announcement? (see outreach folder for the current draft in the 0.21 release branch).

roderickvd:
Sure - you mean you want to target it for v0.21?

yara-blue:
> Sure - you mean you want to target it for v0.21?

Why not :)

roderickvd:
> Could you add a section to the upgrade guide? Do we need to say something in the release announcement? (see outreach folder for the current draft in the 0.21 release branch).

Done in [dd2a49b](https://github.com/RustAudio/rodio/pull/753/commits/dd2a49bc99d1dcf3c6c8692218c86b0d18406a5f).

I think we can skip it in the release announcement; it should be all the same to 99.9% of existing users.

dvdsk (approved):

yara-blue:
@PetrGlad would you agree with these changes? (no need for full review though one is always welcome).

alice-i-cecile:
> Bevy runs with default features false so it should not be an issue for them (as they have zero decoders enabled by default)

Bevy maintainer here: yep, this won't cause any problems for our users as a result.

PR #755, opened 2025-06-13
Add noise generators and improve their distribution

roderickvd:
### Fixes

The current `WhiteNoise` generator has a problem where its distribution is not at all in range of `[-1.0, 1.0]` due to precision loss, truncation and bias to zero from doing this:

```rust
let rand = self.rng.next_u32() as f32 / u32::MAX as f32;
let scaled = rand * 2.0 - 1.0;
```

This PR fixes that by using a proper `f32` generated uniform distribution.

Then the current `PinkNoise` builds on top of the incorrect `WhiteNoise` but worse: has coefficients that are valid for 44.1 kHz only. This PR fixes that by algorithmically generating proper pink noise.

Finally, `try_seek` should provide deterministic seeking (same results after seeking) and `PinkNoise` cannot provide that. This PR correctly implements `try_seek` for noise generators that are deterministic (or uniformly random).

### New noise generators

Threw in some freebies:

- Gaussian white noise
- Triangular white noise
- Blue noise
- Pink noise
- Brownian noise
- Velvet noise

Including documentation on when you'd want to use which.

I noticed that docs.rs didn't document the feature-gated noise generators, so I've set it to generate documentation for all features.

### Advanced construction options

The current API remains as-is: `source::white()` will give you a `WhiteGenerator<SmallRng>`. But you see that I made the generators generic for more choice to the user: `source::WhiteNoise::<R: Rng>::new{_with_seed}()`.

This now provides a rather complete suite of high-quality noise generators for audio synthesis, testing, and dithering.


yara-blue:
> Finally, `try_seek` should provide deterministic seeking (same results after seeking)

should it? I've never seen seek as reproducing something rather navigating in an underlying stream. I would say the seek implementation for noise like this should do nothing. But if we do go on with `seek in noise -> error` then we should provide a wrapper that "eats" the seek operation. Basically:

```rust
struct DisableSeek;

impl Source for DisableSeek {
    ....
    
    fn try_seek() -> Result<() , _> {
        Ok(())
    }
}
```

Cargo.toml
```diff
@@ -64,6 +62,9 @@ symphonia-wav = ["symphonia/wav", "symphonia/pcm", "symphonia/adpcm"]
 symphonia-alac = ["symphonia/isomp4", "symphonia/alac"]
 symphonia-aiff = ["symphonia/aiff", "symphonia/pcm"]
 
+[package.metadata.docs.rs]
+all-features = true
```
yara-blue: thanks, I had no idea this was not working. Good find and fix!

src/source/mod.rs
```diff
@@ -87,7 +87,11 @@ mod zero;
 #[cfg(feature = "noise")]
 mod noise;
 #[cfg(feature = "noise")]
-pub use self::noise::{pink, white, PinkNoise, WhiteNoise};
+pub use self::noise::{
```
yara-blue: We could consider making the noise module public in source. Then the end user could access them like:
```rust
use rodio::source::noise;

let a = noise::brownian(); 
```

The advantage is that it makes it possible to have the word noise next to the noise generator. Alternative would be naming them all: `<color>_noise`. 

If we do make the noise folder public we could also remove the `Noise` suffix from the structs.

tests/noise_generators.rs
```diff
@@ -0,0 +1,352 @@
...
+        assert!(
+            sample1 >= -1.0 && sample1 <= 1.0,
+            "Triangular white noise sample out of range: {}",
+            sample1
+        );
+    }
+}
+
+#[test]
+fn test_pink_noise_with_seed_deterministic() {
```
yara-blue: the equality tests can be de-duplicated using `rstest` by making the functions generic over the noise kind.

See the `seek.rs` tests for how rstest works

tests/noise_generators.rs
```diff
@@ -0,0 +1,352 @@
...
+        assert!(
+            sample1 == 0.0 || sample1 == 1.0 || sample1 == -1.0,
+            "Velvet noise sample has invalid value: {}",
+            sample1
+        );
+    }
+}
+
+#[test]
+fn test_convenience_functions_work() {
```
yara-blue: I'm not sure if this tests anything not already tested by the above or verified by the compiler.

tests/noise_generators.rs
```diff
@@ -0,0 +1,352 @@
...
+    let mut velvet = VelvetNoise::<SmallRng>::new_with_seed(sample_rate, seed);
+
+    assert!(pink.try_seek(Duration::from_secs(1)).is_err());
+    assert!(blue.try_seek(Duration::from_secs(1)).is_err());
+    assert!(brownian.try_seek(Duration::from_secs(1)).is_err());
+    assert!(velvet.try_seek(Duration::from_secs(1)).is_err());
+}
+
+#[test]
+fn test_source_trait_properties() {
```
yara-blue: same here, there's not really any complex logic that leads to these values. (They are hard-coded in the two macros).

tests/noise_generators.rs
```diff
@@ -0,0 +1,352 @@
...
+    assert_eq!(pink.total_duration(), None);
+    assert_eq!(violet.total_duration(), None);
+
+    assert_eq!(white.current_span_len(), None);
+    assert_eq!(pink.current_span_len(), None);
+    assert_eq!(violet.current_span_len(), None);
+}
+
+#[test]
+fn test_different_seeds_produce_different_outputs() {
```
yara-blue: This feels more like a unit test to me. Though one could argue this test prevents a regression I cant really imagine that happening. I think the compilers unused variable warnings would catch that right?

src/source/noise.rs
```diff
@@ -1,141 +1,698 @@
...
+//! use rodio::source::{white, pink, triangular_white, blue, WhiteNoise, PinkNoise, NoiseGenerator};
+//! use rand::rngs::SmallRng;
+//!
+//! // Basic: create different noise types (all at 44.1kHz)
+//! let white = white(44100);                 // For testing equipment linearly
+//! let pink = pink(44100);                   // For pleasant background sound
+//! let triangular = triangular_white(44100); // For TPDF dithering
+//! let blue = blue(44100);                   // For high-passed dithering applications
+//!
+//! // Advanced: create with custom RNG (useful for deterministic output)
```
yara-blue: What is the use-case for bringing a custom RNG? Is it reproducibility in future versions, or performance vs quality? 

Having these generic adds a small amount of friction to the API.

src/source/noise.rs
```diff
@@ -1,141 +1,698 @@
...
+/// - As a base for creating other noise types
+///
+/// **Sound character**: Harsh, static-like, evenly bright across all frequencies.
+/// **Distribution**: RPDF (Rectangular Probability Density Function) - uniform distribution.
+///
+/// ```rust
+/// use rodio::source::white;
+/// let white_noise = white(44100);  // 44.1kHz sample rate
+/// ```
+pub fn white(sample_rate: SampleRate) -> WhiteNoise<SmallRng> {
```
yara-blue: Not sure if we want to continue having these factories. Maybe we should just have the structs with a new method each. What do you think?

src/source/noise.rs
```diff
@@ -1,141 +1,698 @@
...
+    Rng, SeedableRng,
+};
+use rand_distr::{Normal, Triangular};
+
+use crate::{ChannelCount, Sample, SampleRate, Source};
+
+/// Trait providing common constructor patterns for noise generators.
+/// Provides default implementations for `new()` and `new_with_seed()`
+/// that delegate to `new_with_rng()`.
+pub trait NoiseGenerator<R: Rng + SeedableRng> {
```
yara-blue: Is there a use-case for a unified API for noise generating? If not then a macro might be a better fit. Then the user does not need to have the trait in scope. That will be especially relevant if we decided to drop the factory functions.

dvdsk (changes requested):
Did not have time to look at the implementation in depth I do have some feedback/questions regarding the API and tests.

src/source/mod.rs
roderickvd: I agree, I'll do that.

tests/noise_generators.rs
roderickvd: I'll clean up the tests.

src/source/noise.rs
roderickvd: As developer of downstream apps and libraries myself, I often have random generators in my own code. To create message or session UUIDs for example. Then, I want to keep it clean and have my entire codebase draw from that single generator. I dislike the bloat and redundancy of dependencies pulling in different `Rng`s, increasing binary size and memory usage.

I believe the slight complexity in the API is OK, by keeping the simple `white`, `pink`, etc. functions that hide this from the user who is either novice or doesn't care.

src/source/noise.rs
roderickvd: I have no particular preference. The current way doesn't break the API, is not super idiomatic, but also not super un-idiomatic. We could also keep just the existing `white` and `pink` ones and tag them as deprecated.

src/source/noise.rs
roderickvd: I think that's a good point. I just did it for the purpose of DRY.

src/source/noise.rs
yara-blue: The SignalGenerator uses the struct::new instead of factory so it would align with those. I like the idea of keep white and pink and deprecating them. We can throw them out later.

roderickvd:
> > Finally, `try_seek` should provide deterministic seeking (same results after seeking)
> 
> should it? I've never seen seek as reproducing something rather navigating in an underlying stream. I would say the seek implementation for noise like this should do nothing. But if we do go on with `seek in noise -> error` then we should provide a wrapper that "eats" the seek operation.

Here is an example of what would happen if we would "eat" seeks:

```rust
// Create with a fixed seed for determinism
let mut noise = pink::new_with_seed(44100, 12345);
// Play for a while, changing internal state
for _ in 0..1000 { noise.next(); }
// Rewind to the beginning
noise.try_seek(Duration::from_secs(0)).unwrap();

// Create another generator with the same seed
let mut noise2 = pink::new_with_seed(44100, 12345);
// Noises should be the same, but are not
assert_eq!(noise.next(), noise2.next());
```

...where you've got me however is that this is also true for my proposed implementation of `WhiteNoise`. We _can_ get around that by storing/resetting original seeds, but I'm not sure about the added value vs. added complexity.

Thinking out loud: maybe we should remove `_with_seed` altogether and prevent illusions of determinism.

src/source/noise.rs
yara-blue: > I dislike the bloat and redundancy of dependencies pulling in different Rngs, increasing binary size and memory usage.

I had no idea the increase was that big. I would think a RNG uses maybe a kB (for both). Especially if we use a small fast RNG.

yara-blue:
> Thinking out loud: maybe we should remove `_with_seed` altogether and prevent illusions of determinism.

I think a deterministic noise gen is pretty neat but I struggle to think of a use case for it. Noise pretty much sounds the same regardless of seed right? 

If we can not find a good use case now I would propose we scrap determinism. We can always add `deterministic_noise` later that goes through all kinds of complex slow tricks to stay perfectly deterministic.

src/source/noise.rs
roderickvd: You may be right that it's mostly a mindset thing. 

w.r.t. various RNGs: `rand`'s default `Rng` is `ThreadRng` and at some point `SmallRng` was behind a feature gate. Then there are a [couple more](https://github.com/rust-random/rngs) that users can experiment with.

We can still have best of both worlds:

```rust
impl WhiteNoise<SmallRng> {
    pub fn new(sample_rate: SampleRate) -> Self {
        Self::new_with_rng(sample_rate, SmallRng::from_os_rng())
    }
}

impl<R: Rng + SeedableRng> NoiseGenerator<R> for WhiteNoise<R> {
    fn new_with_rng(sample_rate: SampleRate, rng: R) -> Self {
        // ...
    }
}
```

yara-blue: > Then there are a [couple more](https://github.com/rust-random/rngs) that users can experiment with.

What is the effect of different rng's on the noise?

roderickvd:
> If we can not find a good use case now I would propose we scrap determinism. We can always add `deterministic_noise` later that goes through all kinds of complex slow tricks to stay perfectly deterministic.

Yes, let's scrap it.

src/source/noise.rs
roderickvd: Audibly nothing. Just pointing out that there's more than one or two `Rng`s out there that users care for because of speed/uniformity/academia/fun/belief/whatever, and not difficult to support with both `new` and `new_with_rng`.

src/source/mod.rs
roderickvd: Done in https://github.com/RustAudio/rodio/pull/755/commits/377fbb933061db648d099ce4108f6c6f2cc6fac2.

tests/noise_generators.rs
roderickvd: Removed in https://github.com/RustAudio/rodio/pull/755/commits/377fbb933061db648d099ce4108f6c6f2cc6fac2.

tests/noise_generators.rs
roderickvd: Removed in https://github.com/RustAudio/rodio/pull/755/commits/377fbb933061db648d099ce4108f6c6f2cc6fac2.

src/source/noise.rs
roderickvd: https://github.com/RustAudio/rodio/pull/755/commits/377fbb933061db648d099ce4108f6c6f2cc6fac2 does something which I think is pretty cool:

```rust
pub struct WhiteUniform<R: Rng = SmallRng> {
    sample_rate: SampleRate,
    sampler: NoiseSampler<R, Uniform<f32>>,
}
```

so you can do both:

```rust
let white = WhiteUniform::new(44100); // look ma, no generics!
let white_custom = WhiteUniform::<StdRng>::new_with_rng(44100, StdRng::seed_from_u64(12345)); // full monty
```

tests/noise_generators.rs
roderickvd: Cleaned up in https://github.com/RustAudio/rodio/commit/377fbb933061db648d099ce4108f6c6f2cc6fac2. I upgraded `rstest` and `rstest_reuse` dependencies because they contain a bug fix that was needed to use `#[apply]` and `lib` tests.

tests/noise_generators.rs
roderickvd: Removed in https://github.com/RustAudio/rodio/commit/377fbb933061db648d099ce4108f6c6f2cc6fac2.

src/source/noise.rs
roderickvd: Deprecated in https://github.com/RustAudio/rodio/commit/377fbb933061db648d099ce4108f6c6f2cc6fac2.

src/source/noise.rs
roderickvd: Changed in https://github.com/RustAudio/rodio/commit/377fbb933061db648d099ce4108f6c6f2cc6fac2.

yara-blue:
Thanks for all the hard work, it looks _really_ nice.

Only thing I'm still unsure about is the seeking behavior. You've explained you want seeking to be reproducible/deterministic. I think we currently use seeking mostly for decoders. Effects just forward the seek or adjust it in the case of `speed`. Seeking in decoders is not 100% accurate but we find that okay. Isn't it strange to then disable seeking as soon as a piece of audio gets mixed with noise?

roderickvd:
> Isn't it strange to then disable seeking as soon as a piece of audio gets **mixed** with noise?

[emphasis mine]

Are you referring to `Mix` specifically? Because `Mix::try_seek` currently is always non-seekable regardless of its underlying sources.

But yes, you're probably right to not be too academic about this. It's probably OK to "eat" the seek and always return `Ok(())` as any generator will continue to produce continuous noise as long as we don't reset the state.

yara-blue:
> Are you referring to `Mix` specifically? Because `Mix::try_seek` currently is always non-seekable regardless of its underlying sources.

your right! In my head that was already fixed. Well we should be able to fix that now that we have `TrackPosition`.

roderickvd:
OK, [a952f97](https://github.com/RustAudio/rodio/pull/755/commits/a952f972f3b5e78e5a790f844fd32a6a08c54b50) should do it, making all noise generators seekable. Once you agree, I can squash merge.

dvdsk (approved):

yara-blue:
I think this epic is now done :)

Lets get it merged! I'll see if I can finish up get 0.21 ready for release tomorrow.

PR #759, opened 2025-06-26
docs: Update Recommended `release_time` for `AGC` to `0.0`

UnknownSuperficialNight:
After several months of using `AGC` I've concluded that the `release_time` recommendation should be `0.0`

`0.005` has the small possibility of distortion

I tried `0.0001` that fixed it, but in order to guarantee that there are no issues, recommend `0.0` to ensure that `AGC` can always lower the gain when needed


yara-blue:
Super valuable, merging!

(as a side the CI breaking is just clippy getting even better, someone needs to fix that)

PR #760, opened 2025-07-04
fix: AGC set_enabled to use atomic flag with experimental feature

roderickvd:
Caught this oversight while running the test suite manually. Updated to CI to prevent regressions.


dvdsk (approved):
thanks!

yara-blue:
strange the brown noise is failing CI now while it was fine pre merge.... Is it non deterministic?

edit: see you've already solved it

roderickvd:
> strange the brown noise is failing CI now while it was fine pre merge.... Is it non deterministic?

Somewhat, statistics... increasing the sample count to decrease the likelihood of false positives.

PR #761, opened 2025-07-05
Print/log when dropping the OutputStream

dvdsk:
This was a real footgun for new users, they would drop the OutputStream and be suprised their audio stopped playing. The message explicitly mentions dropping stops the audio.

We print on stderr unless the tracing is enabled. If for some reason a user needs a clean stderr and does not want to use tracing they can disable the print using `stream.log_on_drop(false)`.

We do not print if the stream is dropped because the program is panicking. Any print there could confuse the user as to the cause of the panic. It could also interleave with the panic output.

This replaces:
- https://github.com/RustAudio/rodio/pull/746
- https://github.com/RustAudio/rodio/pull/747


roderickvd:
@dvdsk want to slip this into v0.21? Looks good to go.

src/stream.rs
```diff
@@ -58,10 +58,10 @@ impl OutputStream {
         &self.config
     }
 
-    /// [`OutputStream`] is dropped a message is logged to stderr or
+    /// When [`OutputStream`] is dropped a message is logged to stderr or
     /// emitted through tracing if the tracing feature is enabled.
     pub fn log_on_drop(&mut self, enabled: bool) {
-        self.log_on_drop = enabled
```
yara-blue: I don't mind the change since it does not break the semantics but I would like to state that it is AI hallucinated bullshit. An assign expression returns unit which is the same as the functions return type.

yara-blue:
> @dvdsk want to slip this into v0.21? Looks good to go.

Yeah lets go

PR #768, opened 2025-07-13
Implement `try_seek` for MixerSource

azrogers:
Wanted to seek a `MixerSource`, and I noticed most of an implementation in the code already that was commented out in anticipation of #510. Well, #510 has been closed in favor of #585, so this uses the `TrackPosition` added in #585 to implement that code.

### DIFF
--- a/src/mixer.rs
+++ b/src/mixer.rs
@@ -1,7 +1,7 @@
 //! Mixer that plays multiple sounds at the same time.
 
 use crate::common::{ChannelCount, SampleRate};
-use crate::source::{SeekError, Source, UniformSourceIterator};
+use crate::source::{SeekError, Source, TrackPosition, UniformSourceIterator};
 use crate::Sample;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, Mutex};
@@ -69,7 +69,7 @@ impl Mixer {
 /// The output of the mixer. Implements `Source`.
 pub struct MixerSource {
     // The current iterator that produces samples.
-    current_sources: Vec<Box<dyn Source + Send>>,
+    current_sources: Vec<TrackPosition<Box<dyn Source + Send>>>,
 
     // The pending sounds.
     input: Mixer,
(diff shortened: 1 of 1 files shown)


yara-blue:
Very nice, thank you! I'll review it tomorrow when my brain is a little less mushy :moon:

roderickvd:
From a code review this looks good. Can we add some tests, also for rollback scenarios?

Question I've got is: what do we do when a rollback fails? Currently it immediately returns an error, without trying rollback of further sources. I think that makes sense: in any case the caller knows that the mixer is now in an inconsistent state and needs to handle that.

yara-blue:
I'm not sure if these are the best semantics for seek. 

Let me explain with an example:

Here the user wants to create a [canon](https://en.wikipedia.org/wiki/Canon_(music)). They do so by adding another sound to the mix after a few seconds. Seeking now breaks their canon.  
| time | event | mixer sound |
| --- | --- | ---- |
| 0 | mixer starts playing | no sound |
| 1 | sound A added to mixer | start of A |
| 11 | sound B added to mixer | start of B mixed with 10th second of A |
| 20 | seek to second 1 | start of A and start of B mixed |

One could argue users should not rely on when they add things to the mixer. For the example use-case they should prepend silence to the song. 

What do you all think?

roderickvd:
Yeah, that's interesting. Although relative seeks should work right?
I was naively only thinking about all sources having a common T=0.

We could make this _very_ complex. Even relative seeks are not fool-proof, because when one is bounded and the other seeks the full requested value, then you're still out of sync. We could handle that. Or not.  And we could try converting absolute seeks into relative seeks. Or not.

Best to think first of the use cases we want to cater to. I don't use this myself, but could imagine users expecting a mix to function like a linked group of tracks like in a DAW. That'd be quite complex. Any actual users who can chime in?

yara-blue:
Lets see if we can find some middle ground that covers most (simple) use cases.

> but could imagine users expecting a mix to function like a linked group of tracks like in a DAW. That'd be quite complex.

In that case we should provide a different API something like `TrackMixer`. Mix is a core part of Rodio and I would not like it to get too complex.

roderickvd:
> Lets see if we can find some middle ground that covers most (simple) use cases.

All for it. I just don't know (as in "have knowledge of real-world usage") what compromises "most" use cases.

Here is another example: we've got a decent-length music file playing. At some point we mix in 3 seconds of a sound effect, which finishes playing. Sometime later we seek the mix, which we presume is just seeking the music file. This seek should succeed and probably ignore the finished sound effect.

PR #633, opened 2024-10-17
remove `thiserror` dependency

joseluis:
I saw [this comment](https://github.com/RustAudio/rodio/pull/376/commits/41ae680bc79afd3baa6b1a382c164460fc0bee44#r659571587) on the commit that added support for Symphonia, regarding it as an unneccesary and cumbersome dependency, which I subscribe 100%, and since it's a simple change I just did it.

PR #778, opened 2025-08-09
reintroduces thiserror + improve output_to_wav error

yara-blue:
## Thiserror
This reverts PR #633 in which thiserror was removed. At the time there where only two Error types. I expected that number not to increase. Clearly I was wrong as we now we have 6. Therefore this re-introduces thiserror.

Note the original argument against this thiserror by est31 in 2021:

> I'm not in favour of adding thiserror, it pulls in a huge set of
dependencies for very little improvement in comfort.

Since then thiserror has become *the* error handling crate for use in crates (24580 published crates depend on thiserror). The crate most used for error handling in binaries (anyhow) even relies on thiserror. Thiserror is also increadibly stable.

Therefore the chance is really quite slim a rodio user will not already have the right thiserror version somewhere in their dependency tree. Which means the compile time impact is neglible in most cases.

## Output to wav
Gets its own type, it was `Box<dyn Error>` which does not work for wrapping it in anything requiring `Box<dyn Error + Send + Sync + 'static>`. Unfortunaly making wav_output return a
`Box<dyn Error + Send + Sync + 'static>` will break usage in functions returning `Box<dyn Error>` (yeah this suprised me too).

## Errors Clone
This also makes all our errors clonable by swapping Box with Arc and wrapping non clone errors in an Arc.

## Errors trait check macro
Finally I've added a macro (`assert_error_traits`) in `common` to verify an Error type is Send + Sync + Clone + Debug + Display + 'static.

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -1,5 +1,5 @@
-use std::num::NonZero;
 use std::fmt::{Debug, Display};
+use std::num::NonZero;
 
 /// Stream sample rate (a frame rate or samples per second per channel).
 pub type SampleRate = NonZero<u32>;
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -46,14 +46,20 @@
 //! if the corresponding features are enabled.
 
 use std::{
-    io::{BufReader, Read, Seek}, marker::PhantomData, sync::Arc, time::Duration
+    io::{BufReader, Read, Seek},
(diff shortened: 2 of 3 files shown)


roderickvd:
Looks great. I'm in favor.

yara-blue:
rebased and merged :)

PR #785, opened 2025-08-23
Adds Microphone input source

dvdsk:


yara-blue:
@roderickvd I've no idea what the cpal buffersize does for input streams. I saw you made an example over at cpal, could you when you review this carefully check the buffer size section? Feel free to also rewrite that if that's easier.

yara-blue:
Btw I quite like the builder for the microphone (especially the part where you can try to set a parameter using the fact that the builder members take &self). Shall I rewrite the stream one to be similar?

yara-blue:
Note to self: should we just pick f32 as sample format and ignore the default output config? Maybe only if doing so does not limit the other options (sample_rate and channel_count)...

Cargo.toml
```diff
@@ -105,6 +105,7 @@ rand_distr = { version = "0.5", optional = true }
 tracing = { version = "0.1.40", optional = true }
 
 atomic_float = { version = "1.1.0", optional = true }
+rtrb = "0.3.2"
```
roderickvd: To keep the dependency tree slim, I propose that we feature-gate the recording functionality similar to `playback`.

src/microphone.rs
```diff
@@ -0,0 +1,240 @@
...
+//! use rodio::microphone::MicrophoneBuilder;
+//! use rodio::Source;
+//! use std::time::Duration;
+//!
+//! # fn main() -> Result<(), Box<dyn std::error::Error>> {
+//! let mut builder = MicrophoneBuilder::new()
+//!     .default_device()?
+//!     .default_config()?;
+//!
+//! // Try to set stereo recording (2 channels), but continue with default if unsupported
```
roderickvd, line 67: Generally for Rodio I also veer into the direction of automatic fallback. As in: let the user set up as he wants, but in the end make a best effort attempt at truly starting recording, rather than following the requested setup exactly.

For the output devices we are stricter than that though: choose a non-supported sample rate or channel count, and it will fail without automatic fallback. I would favor that we are consistent.

Analogous to the output stream we could also offer `open_stream_or_fallback` methods to have the user opt-in explicitly to fallback behavior.

src/microphone.rs
```diff
@@ -0,0 +1,240 @@
...
+
+    fn channels(&self) -> crate::ChannelCount {
+        self.channels
+    }
+
+    fn sample_rate(&self) -> crate::SampleRate {
+        self.sample_rate
+    }
+
+    fn total_duration(&self) -> Option<std::time::Duration> {
```
roderickvd, line 184: It should be possible to implement `size_hint` using `rtrb::Consumer::slots()`.

src/microphone/builder.rs
```diff
@@ -0,0 +1,418 @@
...
+};
+
+use super::Microphone;
+
+/// Error configuring or opening microphone input
+#[allow(missing_docs)]
+#[derive(Debug, thiserror::Error, Clone)]
+pub enum Error {
+    /// No input device is available on the system.
+    #[error("There is no output device")]
```
roderickvd: `/s/output/input`

src/microphone/builder.rs
```diff
@@ -0,0 +1,418 @@
...
+        source: cpal::SupportedStreamConfigsError,
+        device_name: String,
+    },
+    /// The requested input configuration is not supported by the device.
+    #[error("The input configuration is not supported by input device: '{device_name}'")]
+    UnsupportedByDevice { device_name: String },
+}
+assert_error_traits! {Error}
+
+#[allow(missing_docs)]
```
roderickvd: I wouldn't mind the docs for what you're conjuring here 😅

src/microphone/builder.rs
```diff
@@ -0,0 +1,418 @@
...
+    /// # Example
+    /// ```no_run
+    /// # use rodio::microphone::MicrophoneBuilder;
+    /// let builder = MicrophoneBuilder::new()
+    ///     .default_device()?
+    ///     .default_config()?
+    ///     .samplerate(44_100.try_into()?)?;
+    /// # Ok::<(), Box<dyn std::error::Error>>(())
+    /// ```
+    pub fn samplerate(
```
roderickvd: `/s/samplerate/sample_rate`

src/microphone/builder.rs
```diff
@@ -0,0 +1,418 @@
...
+    ///     channel_count: NonZero::new(2).expect("2 is not zero"),
+    ///     buffer_size: cpal::BufferSize::Fixed(42_000),
+    ///     sample_format: cpal::SampleFormat::U16,
+    /// };
+    /// let builder = MicrophoneBuilder::new()
+    ///     .default_device()?
+    ///     .config(config)?;
+    /// # Ok::<(), Box<dyn std::error::Error>>(())
+    /// ```
+    pub fn config(
```
roderickvd, line 247: In the `DecoderBuilder` we use `with_foo` idiom. Change this to `with_config` and apply the same everywhere else?

src/microphone.rs
```diff
@@ -0,0 +1,240 @@
...
+        )
+        .map_err(OpenError::BuildStream)?;
+        stream.play().map_err(OpenError::Play)?;
+
+        Ok(Microphone {
+            _stream_handle: stream,
+            buffer: rx,
+            channels: config.channel_count,
+            sample_rate: config.sample_rate,
+            time_between_20_frames: Duration::from_secs_f64(1.0 / config.sample_rate.get() as f64),
```
roderickvd: I would change it into `poll_interval: Duration` and then poll every 5-10 ms regardless of sample rate.

src/microphone.rs
```diff
@@ -0,0 +1,240 @@
...
+assert_error_traits! {OpenError}
+
+impl Microphone {
+    fn open(
+        device: Device,
+        config: InputConfig,
+        error_callback: impl FnMut(cpal::StreamError) + Send + 'static,
+    ) -> Result<Self, OpenError> {
+        let timeout = Some(Duration::from_millis(100));
+        let (mut tx, rx) = RingBuffer::new(20_000);
```
roderickvd: I think this should target like 100 ms worth of samples and so scale with the sample rate and channel count.

src/microphone/builder.rs
```diff
@@ -0,0 +1,418 @@
...
+    ///   - Rodio to log errors like: `alsa::poll() returned POLLERR`
+    ///
+    /// # Recommendation
+    /// If low latency is important to you consider offering the user a method
+    /// to find the minimum buffer size that works well on their system under
+    /// expected conditions. A good example of this approach can be seen in
+    /// [mumble](https://www.mumble.info/documentation/user/audio-settings/)
+    /// (specifically the *Output Delay* & *Jitter buffer*.
+    ///
+    /// These are some typical values that are a good starting point. They may also
```
roderickvd, line 448: I think that earlier you were clamping to actually supported buffer sizes, right? In which case they should not break audio completely and adapt to the system.

Though from experience I know sound card drivers can be quite broken, and report all sorts of "supported" buffer sizes they actually don't support (or in powers of two only, or only by moonlight, etc.)

src/microphone.rs
```diff
@@ -0,0 +1,240 @@
...
+}
+
+impl Iterator for Microphone {
+    type Item = f32;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        loop {
+            if let Ok(sample) = self.buffer.pop() {
+                return Some(sample);
+            } else {
```
roderickvd, line 198: So this will loop indefinitely, even if the input is broken and the error callback triggered. Is that something we want to signal here and end the iterator? Or if not using a channel, then by providing a timeout here? Otherwise calls to `next()` may hang.

roderickvd:
> Note to self: should we just pick f32 as sample format and ignore the default output config? Maybe only if doing so does not limit the other options (sample_rate and channel_count)...

Definitely recommend to support all formats you can, then convert. Particularly on Alsa there is no guarantee that the input device can support `f32` without routing it through Alsa plugins - which users don't want because it adds latency and increases setup complexity.

yara-blue:
> Particularly on Alsa there is no guarantee that the input device can support f32 without routing it through Alsa plugins

So if a device reports F32 capability (through cpal supportedconfig) then thats actually alsa converting?

src/microphone.rs
yara-blue: > Generally for Rodio I also veer into the direction of automatic fallback

Here might I disagree, if the user wants things to "just work" they can take the defaults:

```rust
let mut builder = MicrophoneBuilder::new()
    .default_device()?
    .default_config()?;
```
Otherwise it could get horrible to debug: I requested stereo and I get mono? I rather have to handle and error _before opening the stream_.

This builder validates when the setting (in this case channel) is requested if it will be supported by the device.

Theoretically its possible some settings do not commute. For example setting channels to 1 and then sample rate to 91khz works but in the reverse order it does not because the device does not support 2 channels and 91khz. We might need to document that...

> Analogous to the output stream we could also offer open_stream_or_fallback

I'm not sure, I like to steer users to think *what* they want to happen when an error occurs. On the other hand we can just add it and mention in its docs you can also manually do fail-over. 

Open does not consume the builder so we could also have a 'next_so_called_supported_config` method which returns a builder with another config that "should" work.

src/microphone.rs
yara-blue: Does it make sense though? The queue (rtrb) is only there to convert from pushing samples (cpal input) to pulling (rodio). If everything is working there should hardly be any samples in that queue. 

The iterator is expected to yield *far* more samples then what is in the queue, so the hint would be rather misleading. For example if collect (into vec) would allocate the next vec size based on that size_hint it will get terribly inefficient (pretty sure collect does something smarter like min(current_len * 2, size_hint_lower_bound).

src/microphone.rs
yara-blue: Good catch! That would have been an ugly bug!

src/microphone/builder.rs
yara-blue: Not sure what you mean here. What I am unsure about is not whether the buffer size will work but what the effect of a larger buffer on latency is. If the OS always copies at least buffer_size bytes to rodio then a larger buffer size increases latency (which I wish to minimize).

Regarding supported config which they dont actually support: I have a mic that reports being a stereo device but then has one channel all zeros (the hardware is def. not stereo)

src/microphone.rs
roderickvd: Yes, scrap that - I was triggered too much by the sentence "Try to set stereo recording (2 channels), but continue with default if unsupported" without properly looking at the example below it. Sorry for having you write such an extensive reply.

src/microphone.rs
roderickvd: I'd expect that in most cases, it'd yield a fairly stable and truthful `(x, None)` with `x` probably biased towards `buffer.capacity()`.

Compared to returning `(0, None)` it'd spare the need of initializing an empty `Vec` and then constantly growing. Yes the growth factor is about ~2x I believe, but still that's a lot of allocations before you reach the buffer capacity.

Long story short I think it's the _right_ thing to do with no negative and likely better effects.

roderickvd:
> > Particularly on Alsa there is no guarantee that the input device can support f32 without routing it through Alsa plugins
> 
> So if a device reports F32 capability (through cpal supportedconfig) then thats actually alsa converting?

No if it reports that, then that's fine and it'll support that 99.99% of the times.

I thought you meant to *always* assume F32 capability.

src/microphone/builder.rs
roderickvd: > Not sure what you mean here.

I was complimenting your implementation of `supported_given` which I believe is the right approach. Though indeed it's not clamping but anyway checking whether it's supported.

> What I am unsure about is not whether the buffer size will work but what the effect of a larger buffer on latency is. If the OS always copies at least buffer_size bytes to rodio then a larger buffer size increases latency (which I wish to minimize).

`cpal` fires the callback delivering the samples whenever the buffer is full. So yes, larger buffers increase latency.

src/microphone.rs
yara-blue: Let me know if that comments needs work.

src/microphone/builder.rs
yara-blue: I did it but I don't really like it to be honest. std also does not have the with_ (see OpenOptions). How would you feel about dropping the with and scheduling an overhaul of stream to bring it in line with this (use the check_config trick there too)

roderickvd: Sure, I'm after consistency but not married to the `with_` approach. I can pick it up with the mega-PR - shall I deprecate the `with_` ones first or immediately rip them out?

Cargo.toml
```diff
@@ -105,7 +107,7 @@ rand_distr = { version = "0.5", optional = true }
 tracing = { version = "0.1.40", optional = true }
 
 atomic_float = { version = "1.1.0", optional = true }
-rtrb = "0.3.2"
+rtrb = { version = "0.3.2", optional = true }
```
roderickvd, line 110: `inquire` as `dev-dependency` is also optional right?

yara-blue: dev-dependencies are not allowed to be optional

src/microphone/builder.rs
yara-blue: For the decoder-builder, depends on how nice you want to be to our users :), I'm okay with ripping them out. Maintainability outweighs ease of upgrade for me. And its not that hard of an upgrade right?

I'll make an issue to bring stream inline.

src/microphone.rs
yara-blue: I made it better anyway :)

You can now also pass in a list of params (like sample_rate) and it will try to use those in order of preference. If none work no changes are made to the already set config nor is an error returned.

yara-blue:
Should be good to merge now

yara-blue:
Merging this as I kinda need it :sweat_smile:. Anything left over can be fixed later right? Feel free to revert if you feel the need.

roderickvd:
Sure we can do it incrementally. I've got less time as I'm traveling a lot. I see you've already covered a lot of the review comments, thanks. Here's two further ideas:

**Buffer size**

Although I did say that a 100 ms buffer should be fine most of the time, how about using the builder to give the user choice?

```rust
pub struct MicrophoneBuilder<Device, Config, E> {
    // ...
    buffer_duration: Option<Duration>,
    buffer_samples: Option<usize>,
}

impl MicrophoneBuilder<DeviceNotSet, ConfigNotSet> {
    pub fn new() -> Self {
        Self {
            // ...
            buffer_duration: None,
            buffer_samples: None,
        }
    }

    /// - Real-time applications can use smaller buffers (10-20ms)
    /// - Batch processing can use larger buffers (500ms+)
    pub fn buffer_duration(mut self, duration: Duration) -> Self {
        self.buffer_duration = Some(duration);
        self.buffer_samples = None;
        self
    }

    pub fn buffer_samples(mut self, samples: usize) -> Self {
        self.buffer_samples = Some(samples);
        self.buffer_duration = None;
        self
    }
}

impl Microphone {
    fn calculate_buffer_size(
        config: &InputConfig,
        buffer_duration: Option<Duration>,
        buffer_samples: Option<usize>,
    ) -> usize {
        match (buffer_samples, buffer_duration) {
            (Some(samples), _) => samples,
            (None, Some(duration)) => {
                let samples_per_sec = config.channel_count.get() as u32 * config.sample_rate.get();
                (samples_per_sec as f64 * duration.as_secs_f64()) as usize
            },
            (None, None) => {
                let samples_per_sec = config.channel_count.get() as u32 * config.sample_rate.get();
                (samples_per_sec / 10) as usize
            }
        }
    }
}
```

**Polling interval**

We could extend it to cover the polling interval:

```rust
impl MicrophoneBuilder<DeviceIsSet, ConfigIsSet, E> {
    /// Set polling interval for the microphone iterator
    /// 
    /// Lower values = lower latency but higher CPU usage
    /// Higher values = higher latency but lower CPU usage
    /// 
    /// Recommended values:
    /// - Real-time effects: 1-2ms
    /// - Interactive applications: 5ms (default)
    /// - Recording/batch processing: 10-20ms
    pub fn poll_interval(mut self, interval: Duration) -> Self {
        self.poll_interval = Some(interval);
        self
    }
}
```

**Forcefully stopping recording**

Finally I was thinking that you could use the "error channel" to also signal a "should stop" and introduce a `Microphone::stop()`? So that the user can choose to stop the polling and CPU usage when done recording.

By the way, when we've nailed this down and made the decoder builder consistent, by itself this would warrant a v0.22 release I think!

yara-blue:
> Finally I was thinking that you could use the "error channel" to also signal a "should stop" and introduce a Microphone::stop()? So that the user can choose to stop the polling and CPU usage when done recording.

Wen the user is done with recording they should drop the Microphone instance. At that point the underlying cpal stream will drop, with it they polling and CPU usage. Are you proposing we add a pausing system? If so I feel like that should be its own issue and PR.
 
Regarding Buffer size and Polling interval:

yagni. I am not against having those, but lets wait until they are requested.

> By the way, when we've nailed this down and made the decoder builder consistent, by itself this would warrant a v0.22 release I think!

Yeah lets put another one out :+1: 
After 0.22 we should look into adding a second spanless source trait though. I have some rodio sources I've kept out as they do not support spans. And it would solve some long standing issues with queue.

Issue #775, opened 2025-07-24
Examples with ogg files no longer work


mfluehr:
Examples that use ogg files still run, but no longer produce sound. Based on my testing, this problem seems to have been introduced in #753.

yara-blue:
Could you share a little more about your system? Since everything works on our systems.

mfluehr:
I'm running Arch Linux, EndeavourOS. My sound server is Pipewire.

I noticed that `cargo r --example reverb`  prints out this message immediately: "Dropping OutputStream, audio playing through this stream will stop." If behaving as desired, it should instead play audio for a while, and then drop. If I use an mp3 instead of an ogg, the example works correctly.

yara-blue:
I can reproduce the reverb issue 👍

yara-blue:
Switching to the lewton decoder backend addresses it. We might need to make that the default again. It might be an issue similar to #769. As reverb uses buffered which needs the full duration.

Still strange that the music_ogg example does not work. @mfluehr Could you double check that `cargo r --example music_ogg` does not work?

roderickvd:
I'd like to have this issue dug a little deeper before switching defaults again. First, to understand. Second, because unlike MP3, Ogg Vorbis files should have their duration reliably determinable always. So reverb, buffered reads and total duration *should* all be fine even if they are not today.

mfluehr:
@dvdsk, I must've been a little sloppy when trying examples the first time. The only ones that seem broken are "reverb" and "spatial."

yara-blue:
> [@dvdsk](https://github.com/dvdsk), I must've been a little sloppy when trying examples the first time. The only ones that seem broken are "reverb" and "spatial."

no worries :)

Then I can reproduce it all locally :)

martinetd:
Ah, I was banging my head trying rodio wondering why the sink stopped producing sounds whenever I added `.buffered()` or `.repeat_infinite()`...

I can confirm changing the backend (`cargo add rodio --no-default-features --features lewton,playback`) works around the issue.


Also, it's different from #769 in that `.total_duration()` returns a proper duration.. And actually with lewton it's not :)

This might help; printing this:
```rust
    let a = rodio::Decoder::try_from(file)?;
    println!("file duration {:?}", a.total_duration());
    let b = a.buffered();
    println!("buffered duration {:?}", b.total_duration());
    println!("sample rate {:?}", b.sample_rate());
    println!("channels {:?}", b.channels());
[...]

Issue #458, opened 2022-12-05
MP3 current_ frame_ Len and total_ Duration will never change


yunfengwangluo:
MP3 source. current_ frame_ The value returned by len() is always Some (2304)

source. total_ The value returned by duration is always None

Several different music files have been tested. This cannot be a problem with the music file itself. Other music players can obtain it normally

Or is there any other similar open source project recommended!!

pablo-lua:
I have seen the same trouble, total duration doesn't return nothing at all...

Nickiel12:
I am having the same issue with total_duration returning None

Nickiel12:
A fix is proposed in PR #481?

PraxTube:
This issue seems to still be present. Does anyone have a workaround for this, or is .wav and other files the way to go here?

roderickvd:
Implemented in #786.

PR #786, opened 2025-08-24
feat(decoders)!: comprehensive seeking and bit depth detection across all decoders

roderickvd:
This PR implements a major enhancement to Rodio's decoder capabilities, adding universal seeking support and bit depth detection while fixing some issues, improving performance, and reducing user-facing seek errors.

# Key features

 ## Universal seeking support

- All decoders implement `try_seek()` and `new_with_settings()`
- Configurable seeking modes: `SeekMode::Fastest` (coarse) vs `SeekMode::Nearest` (accurate)
- Automatic fallback: always perform best attempt at performing seek regardless of `SeekMode`
- Enhanced `DecoderBuilder` with seeking configuration options

##  Bit depth detection

- New `Source::bits_per_sample()` method returns bit depth for lossless formats

# Bug fixes

- Fixed zero span length issue in Symphonia Ogg Vorbis decoder that broke buffered sources and effects like reverb
- Improved `size_hint()` accuracy across all decoders

# Performance improvements

- `VorbisDecoder`: improved packet reading
- `FlacDecoder`: better block loading with `FlacReaderOptions`
- Major Symphonia refactoring: enhanced performance and multi-track/chained stream support

# API enhancements

## New `DecoderBuilder` methods

```rust
  DecoderBuilder::new()
      .with_seek_mode(SeekMode::Fastest)     // Configure seeking precision
      .with_scan_duration(true)              // Enable duration scanning  
      .with_total_duration(known_duration)   // Pre-computed duration
```

# Test coverage

- Seeking tests across all decoder formats
- Duration accuracy validation across all decoder formats
- Bit depth detection tests for lossless formats
- Hi-res audio support testing (24-bit, 32-bit)

# Breaking changes

1. `SeekError::NotSupported` renamed to `SeekError::SeekingNotSupported`
2. `WavDecoder` no longer implements `ExactSizeIterator` (for accuracy)
3. `DecoderBuilder::with_coarse_seek()` deprecated (use `with_seek_mode()`)
4. `Source::source_intact()` removed (too many unknowns)

# Decoder feature overview

| Decoder   | Seeking | Bit Depth          | Duration   | Size hint   | Span length |
|-----------|:-------:|:------------------:|:----------:|:-------------:|:--------:|
| claxon      | ✅  | ✅      | ✅ | ✅   | ✅ |
| hound       | ✅  | ✅         | ✅ | ✅   | ✅ |
| minimp3       | ✅  | ❌ Lossy            | ✅ | ✅   | ✅ |
| lewton    | ✅  | ❌ Lossy            | ✅ | ✅ | ✅ |
| Symphonia | ✅  | ✅ Format-dependent | ✅ | ✅  | ✅ |

Closes:
- #141
- #190
- #458
- #775
- #802

### DIFF
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -249,18 +249,18 @@ impl<R: Read + Seek> DecoderImpl<R> {
 
     /// Attempts to seek to a given position in the current source.
     #[inline]
-    fn try_seek(&mut self, pos: Duration) -> Result<(), SeekError> {
+    fn try_seek(&mut self, _pos: Duration) -> Result<(), SeekError> {
         match self {
             #[cfg(feature = "hound")]
-            DecoderImpl::Wav(source) => source.try_seek(pos),
+            DecoderImpl::Wav(source) => source.try_seek(_pos),
             #[cfg(feature = "lewton")]
-            DecoderImpl::Vorbis(source) => source.try_seek(pos),
+            DecoderImpl::Vorbis(source) => source.try_seek(_pos),
             #[cfg(feature = "claxon")]
-            DecoderImpl::Flac(source) => source.try_seek(pos),
+            DecoderImpl::Flac(source) => source.try_seek(_pos),
             #[cfg(feature = "minimp3")]
-            DecoderImpl::Mp3(source) => source.try_seek(pos),
+            DecoderImpl::Mp3(source) => source.try_seek(_pos),
             #[cfg(feature = "symphonia")]
-            DecoderImpl::Symphonia(source, PhantomData) => source.try_seek(pos),
+            DecoderImpl::Symphonia(source, PhantomData) => source.try_seek(_pos),
             DecoderImpl::None(_, _) => unreachable!(),
         }
     }
--- a/src/decoder/looped.rs
+++ b/src/decoder/looped.rs
@@ -1,9 +1,11 @@
 use std::{
     io::{Read, Seek},
-    marker::PhantomData,
     sync::Arc,
     time::Duration,
 };
+
+#[cfg(feature = "symphonia")]
+use std::marker::PhantomData;
 
 use crate::{
     common::{ChannelCount, SampleRate},
@@ -80,14 +82,14 @@ where
         let mut fast_settings = self.settings.clone();
         fast_settings.total_duration = self.cached_duration;
 
-        let (new_decoder, sample) = match decoder {
+        match decoder {
             #[cfg(feature = "hound")]
             DecoderImpl::Wav(source) => {
                 let mut reader = source.into_inner();
                 reader.rewind().ok()?;
                 let mut source = wav::WavDecoder::new_with_settings(reader, &fast_settings).ok()?;
                 let sample = source.next();
-                (DecoderImpl::Wav(source), sample)
+                Some((DecoderImpl::Wav(source), sample))
             }
             #[cfg(feature = "lewton")]
             DecoderImpl::Vorbis(source) => {
@@ -96,7 +98,7 @@ where
                 let mut source =
                     vorbis::VorbisDecoder::new_with_settings(reader, &fast_settings).ok()?;
                 let sample = source.next();
-                (DecoderImpl::Vorbis(source), sample)
+                Some((DecoderImpl::Vorbis(source), sample))
             }
             #[cfg(feature = "claxon")]
             DecoderImpl::Flac(source) => {
@@ -105,7 +107,7 @@ where
                 let mut source =
                     flac::FlacDecoder::new_with_settings(reader, &fast_settings).ok()?;
                 let sample = source.next();
-                (DecoderImpl::Flac(source), sample)
+                Some((DecoderImpl::Flac(source), sample))
             }
             #[cfg(feature = "minimp3")]
             DecoderImpl::Mp3(source) => {
@@ -113,7 +115,7 @@ where
                 reader.rewind().ok()?;
                 let mut source = mp3::Mp3Decoder::new_with_settings(reader, &fast_settings).ok()?;
                 let sample = source.next();
-                (DecoderImpl::Mp3(source), sample)
+                Some((DecoderImpl::Mp3(source), sample))
             }
             #[cfg(feature = "symphonia")]
             DecoderImpl::Symphonia(source, PhantomData) => {
@@ -122,11 +124,10 @@ where
                 let mut source =
                     symphonia::SymphoniaDecoder::new_with_settings(reader, &fast_settings).ok()?;
                 let sample = source.next();
-                (DecoderImpl::Symphonia(source, PhantomData), sample)
+                Some((DecoderImpl::Symphonia(source, PhantomData), sample))
             }
-            DecoderImpl::None(_, _) => return None,
-        };
-        Some((new_decoder, sample))
+            DecoderImpl::None(_, _) => None,
+        }
     }
 }
 


roderickvd:
I feel a bit guilty about this mega-commit. I got carried away while coding on the camping and this is what happened 😊

yara-blue:
6k lines.... AAAHHHHGHHHHHH

yara-blue:
This might take more then one weekend to review :sweat_smile: 
Hope you dont mind :)

src/decoder/builder.rs
```diff
@@ -14,10 +39,10 @@
 //!     let file = File::open("audio.mp3")?;
 //!     let len = file.metadata()?.len();
 //!
-//!     Decoder::builder()
+//!     let decoder = Decoder::builder()
 //!         .with_data(file)
 //!         .with_byte_len(len)      // Enable seeking and duration calculation
-//!         .with_hint("mp3")        // Optional format hint
+//!         .with_hint("mp3")        // Optional format hint for performance
```
yara-blue: // Reduce format detection overhead

src/decoder/builder.rs
```diff
@@ -46,59 +112,315 @@ use ::symphonia::core::io::{MediaSource, MediaSourceStream};
...
+/// - `seek_mode`: Balances seeking speed vs. accuracy
 #[derive(Clone, Debug)]
 pub struct Settings {
-    /// The length of the stream in bytes.
-    /// This is required for:
-    /// - Reliable seeking operations
-    /// - Duration calculations in formats that lack timing information (e.g. MP3, Vorbis)
+    /// The total length of the stream in bytes.
+    ///
+    /// This setting enables several important optimizations:
```
yara-blue: Depending on the format this can enable ...

src/decoder/builder.rs
```diff
@@ -46,59 +112,315 @@ use ::symphonia::core::io::{MediaSource, MediaSourceStream};
...
+    /// # Format Support
+    ///
+    /// - **MP3**: Removes encoder delay and padding frames
+    /// - **AAC**: Removes padding specified in container metadata
+    /// - **FLAC**: Generally gapless by nature, minimal effect
+    /// - **OGG Vorbis**: Handles sample-accurate boundaries
+    ///
+    /// # Duration Impact
+    ///
+    /// Disabling gapless may affect duration calculations as padding frames
```
yara-blue: This is unclear to me, does
- the reported total_duration change when switching gapless on/off
or
- is the reported total_duration wrong when gapless is off?

src/decoder/builder.rs
```diff
@@ -46,59 +112,315 @@ use ::symphonia::core::io::{MediaSource, MediaSourceStream};
...
+    /// Format extension hint for accelerated format detection.
+    ///
+    /// Providing an accurate hint significantly improves decoder initialization
+    /// performance by reducing the number of format probes required. Common
+    /// values include file extensions without the dot.
+    ///
+    /// # Common Values
+    ///
+    /// - Audio formats: "mp3", "flac", "wav", "ogg", "m4a"
+    /// - Container hints: "mp4", "mkv", "webm"
```
yara-blue: so what do I pass in for a mkv containing flac?

src/decoder/builder.rs
```diff
@@ -46,59 +112,315 @@ use ::symphonia::core::io::{MediaSource, MediaSourceStream};
...
+    pub gapless: bool,
 
-    /// An extension hint for the decoder about the format of the stream.
-    /// When known, this can help the decoder to select the correct codec.
-    pub(crate) hint: Option<String>,
+    /// Format extension hint for accelerated format detection.
+    ///
+    /// Providing an accurate hint significantly improves decoder initialization
+    /// performance by reducing the number of format probes required. Common
+    /// values include file extensions without the dot.
```
yara-blue: Maybe the decoder could have a `.open()` method that takes a path, then we can set this from that.

src/decoder/builder.rs
```diff
@@ -46,59 +112,315 @@ use ::symphonia::core::io::{MediaSource, MediaSourceStream};
...
+
+    /// Enable expensive file scanning for accurate duration computation.
+    ///
+    /// When enabled, allows decoders to perform comprehensive file analysis
+    /// to determine accurate duration information. This can be slow for large
+    /// files but provides the most accurate duration data.
+    ///
+    /// # Prerequisites
+    ///
+    /// This setting only takes effect when:
```
yara-blue: The builder validates this and returns an error if there are conflicting settings right? If so that could be mentioned here.

src/decoder/builder.rs
```diff
@@ -135,7 +487,15 @@ impl<R> Default for DecoderBuilder<R> {
 impl<R: Read + Seek + Send + Sync + 'static> DecoderBuilder<R> {
     /// Creates a new decoder builder with default settings.
     ///
+    /// Initializes the builder with sensible defaults suitable for most use cases:
+    /// - Gapless playback enabled
+    /// - Nearest seeking mode (accuracy over speed)
+    /// - No format hints (universal detection)
+    /// - Seeking disabled (streaming-friendly)
```
yara-blue, line 363: why mention seeking mode if seeking is disabled anyway?

src/decoder/builder.rs
```diff
@@ -164,10 +549,17 @@ impl<R: Read + Seek + Send + Sync + 'static> DecoderBuilder<R> {
...
     /// - Reliable seeking operations
     /// - Duration calculations in formats that lack timing information (e.g. MP3, Vorbis)
+    ///   when `scan_duration` is enabled
+    /// - Symphonia decoders may also use this for internal scanning operations
     ///
     /// Note that this also sets `is_seekable` to `true`.
     ///
-    /// The byte length should typically be obtained from file metadata:
+    /// To enable duration scanning for formats that require it, use `with_scan_duration(true)`.
+    /// File-based decoders (`try_from(File)`) automatically enable scanning.
```
yara-blue: s/File-based decoders (/when using

yara-blue: actually the same is said 2 lines down so I think we can scrap this line entirely right?

src/decoder/builder.rs
```diff
@@ -46,59 +112,315 @@ use ::symphonia::core::io::{MediaSource, MediaSourceStream};
...
+///
+/// # Performance Impact
+///
+/// Several settings significantly affect performance:
+/// - `byte_len`: Enables efficient seeking and duration calculation
+/// - `hint`/`mime_type`: Reduce format detection overhead
+/// - `scan_duration`: Controls expensive file analysis operations
+/// - `seek_mode`: Balances seeking speed vs. accuracy
 #[derive(Clone, Debug)]
 pub struct Settings {
```
yara-blue: I would consider keeping all these fields private, since they are set by the builders methods anyway.

src/decoder/builder.rs
```diff
@@ -263,29 +674,119 @@ impl<R: Read + Seek + Send + Sync + 'static> DecoderBuilder<R> {
...
+
     /// Creates the decoder implementation with configured settings.
+    ///
+    /// This internal method handles the format detection and decoder creation process.
+    /// It attempts to create decoders in a specific order, passing the data source
+    /// between attempts until a compatible format is found.
+    ///
+    /// # Format Detection Order
+    ///
+    /// Decoders are tried in this order for optimal performance:
```
yara-blue: I find the source code pretty readable, don't need the doc-comment. More afraid that the comment and code will get out of sync, and if they dont the presence of the doc makes it harder to refactor (as yo have to fix the docs too).

src/decoder/flac.rs
```diff
@@ -1,62 +1,302 @@
...
-use std::time::Duration;
+//! FLAC audio decoder implementation.
+//!
+//! This module provides FLAC decoding capabilities using the `claxon` library. The FLAC format
+//! is a lossless audio compression format that preserves original audio data while reducing file
+//! size through sophisticated entropy coding and decorrelation techniques.
+//!
+//! # Features
+//!
+//! - **Bit depths**: Full support for 8, 16, 24, and 32-bit audio (including 12 and 20-bit)
```
yara-blue: So.... full support for 8, 12, 16, 20, 24 and 32 bit?

src/decoder/flac.rs
```diff
@@ -1,62 +1,302 @@
...
+    ///
+    /// Stores raw i32 samples from the current FLAC block in the decoder's native
+    /// interleaved format. Capacity is pre-allocated based on stream's maximum block size.
     current_block: Vec<i32>,
+
+    /// Number of samples per channel in current block.
+    ///
+    /// Used for calculating the correct memory layout when accessing interleaved samples.
+    /// FLAC blocks can have variable sizes, so this changes per block.
     current_block_channel_len: usize,
```
yara-blue: s/current_block_channel_len/current_block_samples_per_channel/

src/decoder/flac.rs
```diff
@@ -1,62 +1,302 @@
...
+    Source,
+};
+
+/// Reader options for `claxon` FLAC decoder.
+///
+/// Configured to skip metadata parsing and vorbis comments for faster initialization.
+/// This improves decoder creation performance by only parsing essential stream information
+/// needed for audio playback.
+///
+/// # Fields
```
yara-blue, line 84: why not document the fields? themselves?

src/decoder/flac.rs
```diff
@@ -1,62 +1,302 @@
...
 }
 
 impl<R> FlacDecoder<R>
 where
     R: Read + Seek,
 {
-    /// Attempts to decode the data as FLAC.
-    pub fn new(mut data: R) -> Result<FlacDecoder<R>, R> {
-        if !is_flac(data.by_ref()) {
+    /// Attempts to decode the data as FLAC with default settings.
```
yara-blue, line 191: note to self: got till here (dont trust git to keep my review comments unless I send them now)

dvdsk (changes requested):
Review still in progress

roderickvd:
> This might take more then one weekend to review 😅 Hope you dont mind :)

Sure take the time! I got that coming 😜 

Anyway incorporating your review comments iteratively is a fine approach, because one comment may apply to a lot of other decoders as well without you needing to go through all of them.

src/decoder/builder.rs
roderickvd: Only with the Symphonia decoder, when gapless is on, the `total_duration` is usually different from what other codecs and even metadata reports, because it trims silence from full blocks / pages where other decoders don't. For this reason, the `total_duration` tests explicitly use `DecoderBuilder::with_gapless(false)`.

src/decoder/builder.rs
roderickvd: Good point, I will add it. The answer is: `audio/x-matroska`. Like with MP4 you need to specify `audio/mp4` instead of `audio/aac`.

_Note to self_: I need to complete these proposed docs with proper MIME types.

src/decoder/builder.rs
roderickvd: You mean besides `new()` and `new_with_settings()`? What do you think about using the builder instead, and adding a knob `with_auto_hints` to do that when a manual one isn't specified?

src/decoder/builder.rs
roderickvd: Currently the builder delegates invariant checking to the decoder implementations. I thought about auto-setting it in the builder when `is_seekable` and `byte_len` both become available etc. but steered away. It could be that a source _is_ seekable and _has_ a known length, yet the user still doesn't want it scanned to limit I/O.

src/decoder/builder.rs
roderickvd: So that the users know which mode they're getting the moment they *do* enable it.

src/decoder/builder.rs
roderickvd: > s/File-based decoders (/when using

You mean: "When using `try_from<file)` scanning is enabled automatically." ?

src/decoder/builder.rs
roderickvd: I also asked myself: "What is against having users create their own, or inspect/modify the fields after creation?"
In the end I thought: choice is good, there's very limited foot-shooting possibilities. What do you think?

src/decoder/flac.rs
roderickvd: Tried it: Rustdoc doesn't support documenting fields in a constant.

src/decoder/vorbis.rs
```diff
@@ -1,153 +1,1114 @@
...
-    }
-    pub fn from_stream_reader(mut stream_reader: OggStreamReader<R>) -> Self {
-        let mut data = match stream_reader.read_dec_packet_generic::<InterleavedSamples<Sample>>() {
-            Ok(Some(d)) => d.samples,
-            _ => Vec::new(),
+        // Calculate total duration using the new settings approach (before consuming data)
+        let last_granule = if settings.total_duration.is_some() {
+            None // Use provided duration directly
+        } else if settings.scan_duration && settings.is_seekable && settings.byte_len.is_some() {
+            find_last_granule(&mut data, settings.byte_len.unwrap()) // Scan for duration
```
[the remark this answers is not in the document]
roderickvd: `byte_len` is guaranteed by the `if` statement above it, but I can see if derestructuring makes sense without increasing verbosity.

src/decoder/builder.rs
yara-blue: Could add something like: When using the symphonia decoder the reported total_duration still matches the actual duration with gapless enabled. Other decoders fail to take this into account.

src/decoder/builder.rs
yara-blue: I think it makes the PR longer :stuck_out_tongue_closed_eyes: 

More seriously: It increases the maintenance burden as we now have the same docs in two different places. But we could solve that be having the field docs refer to the member docs?

src/decoder/builder.rs
yara-blue: Yes I meant an extra factory for the builder that takes a path. It could directly return a result, immediately returning if the file could not be opened.

src/decoder/builder.rs
yara-blue: Oh I agree with not auto setting it. I would just reword the docs to say something like: the Builder::build method returns an error if this is set but ... and ... are not ...

src/decoder/builder.rs
yara-blue: Okay

src/decoder/builder.rs
yara-blue: Yeah, that communicates the same as `File-based decoders (`try_from(File)`) automatically enable scanning.` right?

src/decoder/builder.rs
roderickvd: Actually it _is_ enabled by default, but now only comes in effect when `is_seekable` and `byte_len` are also set. I think that that's what users normally expect, that `total_duration` is available.

src/decoder/builder.rs
roderickvd: Rather than adding a `with_auto_hints` knob? I'm not so sure about the distinction between `new()` and `open()`.

src/decoder/builder.rs
roderickvd: Yes, just wanted to be sure I understood you.

roderickvd:
I got some smaller updates to this PR committed locally. Would you rather have me wait pushing them, so your review doesn't get mixed up? As you want.

src/decoder/flac.rs
```diff
@@ -69,63 +309,289 @@ where
...
+    /// The returned value is constant for the lifetime of the decoder and matches
+    /// the channel count specified in the FLAC stream metadata.
     #[inline]
     fn channels(&self) -> ChannelCount {
         self.channels
     }
 
+    /// Returns the sample rate in Hz.
+    ///
+    /// Common rates that FLAC supports are:
```
yara-blue: Doesn't flac support every sample range between 1 and ....big number?

In that case this should be: common sample rates

src/decoder/flac.rs
```diff
@@ -69,63 +309,289 @@ where
...
     #[inline]
     fn total_duration(&self) -> Option<Duration> {
         self.total_duration
     }
 
+    /// Returns the bit depth of the audio samples.
+    ///
+    /// FLAC is a lossless format that preserves the original bit depth:
+    /// - 16-bit: Standard CD quality
+    /// - 24-bit: Professional/high-resolution audio
```
yara-blue: I read this as saying 24 & 32 bit audio sound better then 16 bit audio. That might not be what you mean here as there  is plenty of (blind) A/B testing showing that humans *can not* tell the difference between 16 and 24 bit. 

There is a use-case for 24bit I believe: recovering some very faint signal. Regarding 32 bit audio.... that's a total waste of space there is no use case at all. The second you read that into rodio we will throw away the 8 extra bits anyway..... (f32 has a 24 bit mantissa).

roderickvd: Or doing recording, mixing or DSP. The point is not to make any qualifications, just that it exists and where it's used.

```shell
roderick@Mac rodio % ffprobe -hide_banner assets/audacity32bit_int.wav 
Input #0, wav, from 'assets/audacity32bit_int.wav':
  Metadata:
    encoder         : Lavf55.33.100
  Duration: 00:00:00.51, bitrate: 2823 kb/s
  Stream #0:0: Audio: pcm_s32le ([1][0][0][0] / 0x0001), 44100 Hz, stereo, s32, 2822 kb/s
                                                                           ^^^
```

> The second you read that into rodio we will throw away the 8 extra bits anyway..... (f32 has a 24 bit mantissa).

Yes, I've been thinking about what to do with that. As the `decoder_returns_hi_res_bit_depths` now shows, `bits_per_sample()` will return `Some(32)` for this file while we know it'll be truncated to 24 bit mantissa. This seems a case of no matter what we do, it'll be wrong:
1. return `Some(24)` and users may say it's a 32-bit file and `ffprobe` proves so
2. return `Some(32)` and users may say what you do about mantissa.

I'm inclined to do number 2 and then document with `Source::bits_per_sample` that it's the *source* bit depth, not the Rodio pipeline bit depth.

src/decoder/flac.rs
```diff
@@ -145,27 +611,130 @@ where
...
-                _ => return None,
+                Ok(None) | Err(_) => {
+                    // No more blocks or error, current_block becomes empty
+                    // (buffer was consumed by read_next_or_eof)
+                    return None;
+                }
             }
         }
     }
+
```
yara-blue, line 587: note to self, reviewed till here

src/decoder/builder.rs
yara-blue: `with_auto_hints` would then take a path as argument?

src/decoder/builder.rs
yara-blue: I, see... I wonder if there is a way we can make it so users get an error when they indicate in the builder they want total_duration while is_seekable and byte_len are false...

yara-blue:
> I got some smaller updates to this PR committed locally. Would you rather have me wait pushing them, so your review doesn't get mixed up? As you want.

Only see these now, lets keep them locally for now. Feel free to resolve the things I brought up once you've addressed them locally (or let me know why I'm wrong and I'll close em).

I fear it might take me quite some time (few weekends) to review this as I don't have that much time throughout the week...

src/decoder/builder.rs
roderickvd: Problem is there's so many different cases. Some codecs can just return `total_duration` from headers, without `is_seekable` or `byte_len`. But for sure you need random-access seeking and a bounded stream when you want to scan.

src/decoder/builder.rs
roderickvd: Ah yes, I was not thinking straight. You are right, we'd need something takes a `Path` or `Into<Path>`. Still not sure about `open()` though. We could implement `Decoder::try_from<Path>`?

roderickvd:
> > I got some smaller updates to this PR committed locally. Would you rather have me wait pushing them, so your review doesn't get mixed up? As you want.
> 
> Only see these now, lets keep them locally for now. Feel free to resolve the things I brought up once you've addressed them locally (or let me know why I'm wrong and I'll close em).

Will do.

> I fear it might take me quite some time (few weekends) to review this as I don't have that much time throughout the week...

No haste. I'm back to work starting tomorrow, too.

yara-blue:
sorry I spend all my time on #791 in response to #782. I wont be able to look at this again for a while :(

roderickvd:
I'm in no haste. Though as it touches a lot of sources and I want to prevent rebasing as much as possible, I'll wait to have this reviewed before I start other Rodio development.

src/decoder/symphonia.rs
```diff
@@ -175,160 +577,721 @@ impl Source for SymphoniaDecoder {
...
-        // Seeking is a demuxer operation without the decoder knowing about it,
-        // so we need to reset the decoder to make sure it's in sync and prevent
-        // audio glitches.
-        self.decoder.reset();
+            // TODO: remove when Symphonia has fixed linear seeking for Vorbis
+            if self.decoder.codec_params().codec == CODEC_TYPE_VORBIS {
+                for _ in self.samples_read..target_samples {
+                    let _ = self.next();
+                }
+                return Ok(());
```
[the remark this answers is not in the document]
roderickvd: This will not work, because this precisely is a workaround for Symphonia when `is_seekable` is false, in which case Ogg demuxer seeking would fail silently.

PetrGlad:
I wanted to play with it, are there any uses of the `Builder` uses in examples/* ?

src/decoder/builder.rs
```diff
@@ -26,18 +51,59 @@
...
+//!         .with_byte_len(len)
+//!         .with_hint("flac")
+//!         .with_mime_type("audio/flac")
+//!         .with_seekable(true)
+//!         .with_seek_mode(SeekMode::Nearest)
+//!         .with_scan_duration(true)
+//!         .with_gapless(false)
+//!         .build()?;
+//!
+//!     // High-quality decoder with precise seeking
```
PetrGlad: // Move the comment up maybe?

src/decoder/builder.rs
PetrGlad: Are all decoders expected to support all the options?

roderickvd:
> I wanted to play with it, are there any uses of the `Builder` uses in examples/* ?

No, though the Rustdoc contains a lot of examples.

"Behind the scenes" the examples do use it, because `Decoder::TryFrom<File>` uses the builder to construct a decoder.

src/decoder/builder.rs
roderickvd: No, on various places it's documented that it's decoder-dependent.

Currently, options are only used by Symphonia. This PR expands it to the other decoders to the subset of options they can actually use. For example, `hint` still only is supported by Symphonia.

src/decoder/flac.rs
```diff
@@ -69,63 +309,289 @@ where
...
+    /// the channel count specified in the FLAC stream metadata.
     #[inline]
     fn channels(&self) -> ChannelCount {
         self.channels
     }
 
+    /// Returns the sample rate in Hz.
+    ///
+    /// Common rates that FLAC supports are:
+    /// - **44.1kHz**: CD quality (most common)
```
yara-blue: Instead having a description like this for every audio source we should probably point to one spot.

This for example should link to: https://docs.rs/rodio/latest/rodio/source/trait.Source.html#sampling

src/decoder/flac.rs
```diff
@@ -69,63 +309,289 @@ where
...
     }
 }
 
 impl<R> Iterator for FlacDecoder<R>
 where
     R: Read + Seek,
 {
+    /// The type of items yielded by the iterator.
+    ///
+    /// Returns `Sample` (typically `f32`) values representing individual audio samples.
```
yara-blue: typically? This is always rodio::Sample which is always f32 right?

src/decoder/flac.rs
```diff
@@ -69,63 +309,289 @@ where
...
 
+    /// Returns the next audio sample from the FLAC stream.
+    ///
+    /// This method implements efficient block-based decoding by maintaining an internal
+    /// buffer of the current FLAC block. It returns samples one at a time while
+    /// automatically decoding new blocks as needed.
+    ///
+    /// # Sample Format Conversion
+    ///
+    /// Raw FLAC samples are converted to Rodio's sample format based on bit depth:
```
yara-blue: This is very clear from the code for me. I do not think we need this comment.

src/decoder/flac.rs
```diff
@@ -145,27 +611,130 @@ where
...
+
+    /// Returns bounds on the remaining length of the iterator.
+    ///
+    /// Provides accurate size estimates based on FLAC metadata when available.
+    /// This information can be used by consumers for buffer pre-allocation
+    /// and progress indication.
+    ///
+    /// # Returns
+    ///
+    /// A tuple `(lower_bound, upper_bound)` where:
```
yara-blue: The meaning of the tuple is specified by the Iterator trait. Unless we are deviating from that users should depend on the Iterator docs not a doc comment in the implementation. If we are deviating from the Iterator trait that we should add our own method.

src/decoder/flac.rs
```diff
@@ -145,27 +611,130 @@ where
...
 }
 
-/// Returns true if the stream contains FLAC data, then tries to rewind it to where it was.
-fn is_flac<R>(mut data: R) -> bool
+/// Probes input data to detect FLAC format.
+///
+/// This function attempts to parse the FLAC magic bytes and stream info header to determine if the
+/// data contains a valid FLAC stream. The stream position is restored regardless of the result.
+///
+/// # Arguments
```
yara-blue: We have a type system for this info. I'd propose we not duplicate the type information in comments that will rot away over time (and add many many lines)

src/decoder/flac.rs
```diff
@@ -145,27 +611,130 @@ where
...
+/// * `data` - Mutable reference to the input stream to probe
+///
+/// # Returns
+///
+/// - `true` if the data appears to contain a valid FLAC stream
+/// - `false` if the data is not FLAC or is corrupted
+///
+/// # Implementation
+///
+/// Uses the common `utils::probe_format` helper which:
```
yara-blue: If you want to know what utils::probe_format does read the docs of utils probe_format

src/decoder/flac.rs
```diff
@@ -145,27 +611,130 @@ where
...
-    let result = FlacReader::new(data.by_ref()).is_ok();
-    let _ = data.seek(SeekFrom::Start(stream_pos));
-    result
+    utils::probe_format(data, |reader| FlacReader::new(reader).is_ok())
+}
+
+/// Converts claxon decoder errors to rodio seek errors.
+///
+/// This implementation provides error context preservation when FLAC decoding operations fail
+/// during seeking. The original `claxon` error is wrapped in an `Arc` for thread safety and
```
yara-blue: Implementation detail not relevant to API user. Viewable in standard way in the generated docs.

image

src/decoder/flac.rs
```diff
@@ -145,27 +611,130 @@ where
...
+/// converted to the appropriate Rodio error type.
+///
+/// # Error Mapping
+///
+/// All `claxon::Error` variants are mapped to `SeekError::ClaxonDecoder` with the
```
yara-blue: Same, and further down

src/decoder/mod.rs
```diff
@@ -98,46 +104,60 @@ pub struct Decoder<R: Read + Seek>(DecoderImpl<R>);
 /// let file = File::open("audio.mp3").unwrap();
 /// let looped_decoder = Decoder::new_looped(file).unwrap();
 /// ```
+#[allow(dead_code)]
 pub struct LoopedDecoder<R: Read + Seek> {
     /// The underlying decoder implementation.
     inner: Option<DecoderImpl<R>>,
     /// Configuration settings for the decoder.
     settings: Settings,
+    /// Cached metadata from the first successful decoder creation.
```
yara-blue: metadata?

If we want a comment the next line alone is enough right?

src/decoder/mod.rs
```diff
@@ -98,46 +104,60 @@ pub struct Decoder<R: Read + Seek>(DecoderImpl<R>);
...
+    /// Used to avoid expensive file scanning on subsequent loops.
+    cached_duration: Option<Duration>,
 }
 
 // Cannot really reduce the size of the VorbisDecoder. There are not any
```
yara-blue: what happened here?

src/decoder/mod.rs
```diff
@@ -98,46 +104,60 @@ pub struct Decoder<R: Read + Seek>(DecoderImpl<R>);
...
-// arrays just a lot of struct fields.
+/// Internal enum representing different decoder implementations.
+///
+/// This enum dispatches to the appropriate decoder based on detected format
+/// and available features. Large enum variant size is acceptable here since
+/// these are infrequently created and moved.
```
yara-blue, line 87: We could box/arc the decoders, that would reduce the size

src/decoder/mod.rs
```diff
@@ -98,46 +104,60 @@ pub struct Decoder<R: Read + Seek>(DecoderImpl<R>);
...
     Symphonia(symphonia::SymphoniaDecoder, PhantomData<R>),
-    // This variant is here just to satisfy the compiler when there are no decoders enabled.
-    // It is unreachable and should never be constructed.
+    /// Placeholder variant to satisfy compiler when no decoders are enabled.
+    /// This variant is unreachable and should never be constructed.
     #[allow(dead_code)]
     None(Unreachable, PhantomData<R>),
 }
 
+/// Placeholder type for the None variant that can never be instantiated.
```
yara-blue: Already documented on the variant right?

src/decoder/mod.rs
```diff
@@ -611,63 +734,33 @@ where
...
-                DecoderImpl::Symphonia(source, PhantomData) => {
-                    let mut reader = source.into_inner();
-                    reader.seek(SeekFrom::Start(0)).ok()?;
-                    let mut source =
-                        symphonia::SymphoniaDecoder::new(reader, &self.settings).ok()?;
-                    let sample = source.next();
-                    (DecoderImpl::Symphonia(source, PhantomData), sample)
-                }
-            };
+            let (new_decoder, sample) = self.recreate_decoder_with_cache(decoder)?;
```
yara-blue: Very nice, much more readable then the previous code. I'm wondering if it makes sense to lift looped decoder to its own module? But we can always do that later (its not important).

src/decoder/mp3.rs
```diff
@@ -1,129 +1,917 @@
...
+///
+/// The decoder tracks average bitrate over time to improve seeking accuracy,
+/// especially for VBR files where initial estimates may be inaccurate.
+///
+/// # Channel Changes
+///
+/// MP3 frames can theoretically change channel configuration, though this is
+/// rare in practice. The decoder handles such changes dynamically.
+///
+/// # Thread Safety
```
yara-blue: Should be clear from the missing Sync auto trait implementation on docs.rs. And compile errors

src/decoder/mp3.rs
```diff
@@ -1,129 +1,917 @@
...
+/// * `R` - The underlying data source type, must implement `Read + Seek`
 pub struct Mp3Decoder<R>
 where
     R: Read + Seek,
 {
-    // decoder: SeekDecoder<R>,
-    decoder: Decoder<R>,
-    // what minimp3 calls frames rodio calls spans
-    current_span: Frame,
+    /// The underlying minimp3 decoder, wrapped in Option for seeking operations.
```
yara-blue: The fields are private, so they (and these docs) will not show on docs.rs. I assume we want these fields to stay private so we can write them way less verbose (increase the signal to noise ratio).

This could be: `Option to allow owning the Decoder instance during seeking.`

src/decoder/mp3.rs
```diff
@@ -1,129 +1,917 @@
...
+
+    /// Number of audio channels.
+    ///
+    /// Can theoretically change between frames, though this is rare in practice.
+    /// Updated dynamically when frame channel count changes.
+    channels: ChannelCount,
+
+    /// Sample rate in Hz.
+    ///
+    /// Fixed for the entire MP3 stream. Common rates include 44.1kHz (CD quality),
```
yara-blue: I dont think we need to know what common rates there are in the comments on a private field.

yara-blue: It is usefull to note the sample rate is fixed though.

I would write this doc comment as:
`does not change after decoder creation`

src/decoder/mp3.rs
```diff
@@ -1,129 +1,917 @@
...
+    /// Attempts to decode the data as MP3 with default settings.
+    ///
+    /// This method probes the input data to detect MP3 format and initializes the decoder if
+    /// successful. Uses default settings with no seeking support or duration scanning enabled.
+    ///
+    /// # Arguments
+    ///
+    /// * `data` - Input stream implementing `Read + Seek`
+    ///
+    /// # Returns
```
yara-blue, line 201: Again the java style doc comments. This is a very very verbose way of saying what the function signature is

dvdsk (changes requested):
The code, that I've seen, is of good quality. You clearly put in effort to get every feature of every decoder supported exposed and used. Semantically this looks like a great improvement to Rodio's abilities.

A problem for me is that we seem to have different software engineering practices:
- I prefer a mixture of Robert C. Martin clean code and Martin Fowler approach to refactoring. This favors as few comments as possible instead express as much as possible in the type system and variable & function names. (Doc)comments are seen as almost an evil. There are many reasons for this, for one comments are not checked by a type system.
- From this PR I get the impression you highly value (doc) comments and prefer to have information in comments instead of relying solely on having them encoded in names and the type system.

Having so many (doc) comments makes this PR incredibly hard to review for me. It feels like for every line of code there are about 3 lines of documentation/comments. I have to make sure they are all correct. For me that means I have to go back and re-read the comments to check if that is indeed what the code is doing (instead of just reading the code).

Is there something we can do to remedy this?

roderickvd:
> A problem for me is that we seem to have different software engineering practices:
> 
> [...]
> 
> Having so many (doc) comments makes this PR incredibly hard to review for me.

Yes I know so all the more appreciated.

> Is there something we can do to remedy this?

Yes and it shouldn't be hard. I do like when APIs are well documented and believe we can do better. But as for the definition of "well documented" no need to debate: let's pick a file we consider as golden standard and we can work towards that.

roderickvd:
By the way let me know if you've completed your first pass? Then I can push local changes, start reducing the documentation verbosity when we've settled the golden standard, and ping you for a second review when ready.

yara-blue:
> let's pick a file we consider as golden standard and we can work towards that.

I quite like my microphone PR, though I have some issues with it like: 
- The sample rate docs are tool long and should be extracted to a separate page talking about buffers and their effect on recording or playing.
- Because of the lint some struct fields are which have very clear names. For example `InputConfig.channel_count` really does not need a comment, same for `sample_rate`. It would be better to have those fields undocumented to draw the reader to the field documents on `sample_format`.

There are also cases where extensive documentation is, in my opinion, very welcome like the different limiter configurations and their [documentation](https://docs.rs/rodio/latest/rodio/source/struct.LimitSettings.html). There is even a place for extensive comments like when the algorithm is new and not documented anywhere else like in the case of the AGC https://github.com/RustAudio/rodio/blob/master/src/source/agc.rs

yara-blue:
> By the way let me know if you've completed your first pass? 

Go ahead and push changes :+1:

src/decoder/mod.rs
roderickvd: Generally I prefer `enum` delegation over `Box<dyn T>` virtual dispatch. It may be a little more boilerplate, but the improved performance is generally worth it, and more important than saving RAM here.

roderickvd:
Here's the most prominent changes since last time:

- Added support for decoding from `Bytes`, `Arc<[u8]>`, `Box<[u8]>`, and `Vec<u8>`
- Moved `LoopedDecoder` to its own module
- Consolidated `ReadSeekSource` into the `symphonia` module
- Replaced raw `u32` with `BitDepth` newtype for bit depth values
- Decreased some of the decoder docs - more work can be done here
- Updated all examples to use the new `Decoder::try_from<Path>` pattern

That last bullet meant updating most examples, but it's a small change really.
Advantage over `Decoder::try_from<File>` is that this one uses auto-extension hints.

Jaxx497:
Just wanted to say this PR is incredible work- I don't consider myself knowledgeable enough to help with the review itself, but in testing with my application, a number of prior situations that would throw errors or panics are no longer occurring. Being able to reliably seek through OGG files is huge! Excellent work

yara-blue:
I have a few hours each weekend for open source. I now realize I can not review a PR of this size in those hours. I can also not spread it between weekends since I will lose context in between.

To get this merged it needs to be split into multiple PR's, each limited to around 1000 lines so it can be independently reviewed.

I realize this is not easy. My recommendation is to add the new decoders in parallel to the existing. The new decoders would live under a feature flag. We can then merge each smaller PR without breaking the main lib (this is also known as CI). Ideally the small PR's each add a testable modular part of the new decoders that can be run independently. Though I realize that is too much to ask. However lets architect new efforts so that this can be done. 

Once the whole new decoder suite is in we can remove all the old code and the flag in one PR.

roderickvd:
I am sorry, I don’t see the point in making that effort. The CI is passing and at least one user has given a favorite review. I am not going to put in many hours of free time in something I don’t see the value in. 

I’m still up for a beer sometime in the east, or meeting online, if that expedites things.

will3942:
This is some really good work and would be a shame to see go to waste, if there's value and you want testing from a more complex user I'd be happy to.

Jaxx497:
@dvdsk 

Is there some kind of recourse here?

I can appreciate the desire to implement this in a way, and I can appreciate life outside of OSS, but the rejection of this PR hurts rodio and its users more than it provides any protection to the codebase.

These are real and tangible changes. The premier audio playback crate of Rust should absolutely promote stable audio seeking. I'm not aware what the exact priorities of this project are, but I can't imagine that said functionality isn't of a considerable standing. 

I'm also willing to offer my help in resolving this however I can, although admittedly, I'm not a fraction the developer any of you are. Still, if there is no recourse here, I find it seriously unfair not only to the efforts of Roderick (who is a seriously talented and respectable developer), but to the broader community who will surely benefit from the enhanced functionality.

yara-blue:
> Is there some kind of recourse here?

I don't have anything against the core feature being added here, nor against the code. _I can not_ review a PR of _this size_. So it needs to be split up and ideally also slimmed down.

Some context on the size:
- Rodio without this PR: 14349L 23% comments/docs
- Rodio with this PR: 16868L 40% comments/docs
- This PR adds: 2481L of code and 3486L of comments/docs
  71% op the PR is comments/docs

In my opinion that means this PR can be slimmed down to 2481 * 1.23 ~ 3kL in total. Split that up in roughly 3 parts and its quite reviewable.

---------

Almost all our users seem to use the symphonia decoders, this also PR supports all the current decoders. The symphonia decoders all have the same interface. Supporting all decoders must add a lot of extra code. This matches what I have seen while trying (and failing) to review this the past few weekends. So lets just drop support for all the other decoders in Rodio and in this PR. Then adding this PR might even slim Rodio down, increasing its maintainability.

Instead of letting all that decoder code slip into the void we would move it to another crate: `rodio-extra-decoders`. We then point to that from the Readme.

roderickvd:
I am not sure what you want:
1. Feature-gate and move the decoders in progressively: [comment](https://github.com/RustAudio/rodio/pull/786#issuecomment-3306946607)
2. Extract the alternative decoders into a separate crate: [comment](https://github.com/RustAudio/rodio/pull/786#issuecomment-3314861349)

For item 2 I do not see obvious ways to make that work with the `DecoderBuilder` and `LoopedDecoder`.

I have the feeling you detest the code comments. I already said we can cut down if you want. Beyond that, I do not understand your apprehension. You have set up Rodio with an extensive CI suite for decoders. For seeking in particular that's really not so easy to pass, yet this PR does. That gives some pretty strong behavioral guarantees.

Sure, a passing CI suite does not ensure code quality. Beyond pulling the creds card hoping to have established some known quality standard, I could equally argue that CI also is releasing early and iterate often.

I already said sorry for the camping-built mega commit and ask that we move forward pragmatically.

roderickvd:
> I already said sorry for the camping-built mega commit and ask that we move forward pragmatically.

@dvdsk to raise it to you again.

LDprg:
@roderickvd I want to note something since I have seen this PR. As a user I might ask wheter it makes sense to keep any other backend than Symphonia?

The point I am trying to make is that all other backends seem rather unmaintaned (they all have no real commits the last 2 or more years). Also I have looked into the code of rodio, the different backends increase the complexity alot. Furthermore Symphonia covers all those codes anyways and is actively maintained and used, by alot of people. There is as far as I know no real situation, where you shouldn't use Symphonia (I assume).

So from a development effort point of view I think it might make sense to strip other Codecs to reduce complexity. Since most users will use Symphonia anyways and other could easily migrate. It would also give more time to improvements and new features. A single backend is also just less error prone, since it's less compile time conditionals to care about and more time to spend to fix it.

However I don't think that should effect this PR in particular. Since this work had already been done and it doesn't make sense to just throw it away. So, if the backends get removed, it should be it's own seperate PR.

Just keep in mind that I have little actual knowledge about how rodio actually works, so I might over simply a bit.

will3942:
@dvdsk bump on this too.

yara-blue:
> However I don't think that should effect this PR in particular. Since this work had already been done and it doesn't make sense to just throw it away. So, if the backends get removed, it should be it's own seperate PR.

I disagree there, this is currently blocked on being too big to review. If we consider removing the other backends then this will get significantly smaller and thus easier to cut into ~1000L pieces.

yara-blue:
> I am not sure what you want:
>
> Feature-gate and move the decoders in progressively: https://github.com/RustAudio/rodio/pull/786#issuecomment-3306946607
>  Extract the alternative decoders into a separate crate: https://github.com/RustAudio/rodio/pull/786#issuecomment-3314861349

I'd like to be able to review this PR, for that it needs to shrink to N independent PR's of about ~1k lines. The smaller the total size the faster we will have this merged. I've given some suggestions on how I think that can be achieved. It is up to whomever takes up the work how it is achieved. That does not have to be Roderick who is busy working on Cpal and many other open source projects. As there is clear intrest from the community I'd suggest one of you picks this up.

roderickvd:
> @roderickvd I want to note something since I have seen this PR. As a user I might ask wheter it makes sense to keep any other backend than Symphonia?

As some of the posts in this thread show: there indeed is demand and usage of it. Why - I don't _know_ but I can think of several reasons: licensing, performance, code size.

As Rodio contributor, I don't have very strong feelings about it, other than following the voice of the community.

roderickvd:
When I'm done getting cpal v0.17 out of the door, I'll be turning to some Rodio stuff that my projects need including this one. Not upset with anyone, but if it doesn't get merged here then I'll fork it. I still intend to remain active here and keep everything in sync, so again: not upset myself and don't mean to upset anyone else.

will3942:
I'd strongly suggest we get this in versus losing an active maintainer to the RustAudio ecosystem to work on a fork. As mentioned previously this PR is going to provide clear benefits to this crate.

Jaxx497:
> When I'm done getting cpal v0.17 out of the door, I'll be turning to some Rodio stuff that my projects need including this one. Not upset with anyone, but if it doesn't get merged here then I'll fork it. I still intend to remain active here and keep everything in sync, so again: not upset myself and don't mean to upset anyone else.

I'm very happy to hear this- a project I've recently made public is reliant on this particular PR. [Shameless plug.](https://github.com/Jaxx497/concertus) This PR specifically provided the polish and stability I needed to feel good about sharing my efforts.

I've been seriously considering migrating to an mpv or gstreamer backend, but with this news I may hold off on that or at least make them an opt in option rather than the default. Hopefully the holidays are good for all of you and we see some healthy progress in the new year!

PR #787, opened 2025-08-24
Show an example of compliling the sound into the binary.

dcampbell24:
Seems like something you would want to do fairly often. It is needed for example to package the software on flathub.

### DIFF
--- a/examples/music_ogg.rs
+++ b/examples/music_ogg.rs
@@ -1,11 +1,12 @@
-use std::error::Error;
+use std::{error::Error, io::Cursor};
 
 fn main() -> Result<(), Box<dyn Error>> {
     let stream_handle = rodio::OutputStreamBuilder::open_default_stream()?;
     let sink = rodio::Sink::connect_new(stream_handle.mixer());
 
-    let file = std::fs::File::open("assets/music.ogg")?;
-    sink.append(rodio::Decoder::try_from(file)?);
+    let file = include_bytes!("../assets/music.ogg");
+    let cursor = Cursor::new(file);
+    sink.append(rodio::Decoder::try_from(cursor)?);
 
     sink.sleep_until_end();
 


dcampbell24:
I was not sure if I should do this everywhere or make a separate example of doing this or how exactly I should handle this change, but it seems fairly important. At least I needed to do it.

yara-blue:
> I was not sure if I should do this everywhere or make a separate example of doing this or how exactly I should handle this change, but it seems fairly important. At least I needed to do it.

I don't think it warrants a separate example, but I like having it in there somewhere :+1: 

Thanks!

roderickvd:
Note that using `Decoder::try_from(cursor)` will *not* set `is_seekable` or `byte_len`. However, the [documentation](https://doc.rust-lang.org/std/io/struct.Cursor.html) states:

> A Cursor wraps an in-memory buffer and provides it with a [Seek](https://doc.rust-lang.org/std/io/trait.Seek.html) implementation.

What would you guys say to changing that?

```diff
diff --git a/src/decoder/mod.rs b/src/decoder/mod.rs
index 49647c7..f473b44 100644
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -392,7 +392,14 @@ where
     type Error = DecoderError;
 
     fn try_from(data: std::io::Cursor<T>) -> Result<Self, Self::Error> {
-        Self::new(data)
+        let len = data.get_ref().as_ref().len() as u64;
+
+        Self::builder()
+            .with_data(data)
+            .with_byte_len(len)
+            .with_seekable(true)
+            .with_scan_duration(true)
+            .build()
     }
 }
```

yara-blue:
> Note that using `Decoder::try_from(cursor)` will _not_ set `is_seekable` or `byte_len`. However, the [documentation](https://doc.rust-lang.org/std/io/struct.Cursor.html) states:
> 
> > A Cursor wraps an in-memory buffer and provides it with a [Seek](https://doc.rust-lang.org/std/io/trait.Seek.html) implementation.
> 
> What would you guys say to changing that?
> 
> ```diff
> diff --git a/src/decoder/mod.rs b/src/decoder/mod.rs
> index 49647c7..f473b44 100644
> --- a/src/decoder/mod.rs
> +++ b/src/decoder/mod.rs
> @@ -392,7 +392,14 @@ where
>      type Error = DecoderError;
>  
>      fn try_from(data: std::io::Cursor<T>) -> Result<Self, Self::Error> {
> -        Self::new(data)
> +        let len = data.get_ref().as_ref().len() as u64;
> +
> +        Self::builder()
> +            .with_data(data)
> +            .with_byte_len(len)
> +            .with_seekable(true)
> +            .with_scan_duration(true)
> +            .build()
>      }
>  }
> ```

Lets do it :+1:

PR #789, opened 2025-08-27
nz macro

dvdsk:
I really really miss this macro as a rodio user. So lets make it public.

### DIFF
--- a/src/math.rs
+++ b/src/math.rs
@@ -76,16 +76,27 @@ pub fn linear_to_db(linear: f32) -> f32 {
     linear.log2() * std::f32::consts::LOG10_2 * 20.0
 }
 
-/// short macro to generate a `NonZero`. It panics during compile if the
-/// passed in literal is zero. Used for `ChannelCount` and `Samplerate`
-/// constants
+/// Utility macro for getting a `NonZero` from a literal. Especially 
+/// useful for passing in `ChannelCount` and `Samplerate`.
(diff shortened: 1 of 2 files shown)


yara-blue:
@PetrGlad I'm gonna take that thumbs-up as a positive review :stuck_out_tongue_closed_eyes:

src/math.rs
```diff
@@ -76,16 +76,27 @@ pub fn linear_to_db(linear: f32) -> f32 {
...
+///
+/// # Example
+/// ```
+/// use rodio::nz;
+/// use rodio::static_buffer::StaticSamplesBuffer;
+/// let buffer = StaticSamplesBuffer::new(nz!(2), nz!(44_100), &[0.0, 0.5, 0.0, 0.5]);
+/// ```
+///
+/// # Panics
+/// If the literal passed in is not zero this panicks.
```
roderickvd, line 91: typo: if the literal passed in is ~~not~~ zero this panicks.

yara-blue: addressed on main

PetrGlad:
By the way I found out this one today https://doc.rust-lang.org/std/num/struct.NonZero.html
Maybe it may be a bit heavy for `rodio`.

yara-blue:
> By the way I found out this one today https://doc.rust-lang.org/std/num/struct.NonZero.html
> Maybe it may be a bit heavy for `rodio`.

And this macro makes them :) 

(sample_rate and channel_count are NonZero now)

Issue #38, opened 2015-10-01
Add an Opus decoder


tomaka:

ishitatsuyuki:
libopus looks like BSD. Please consider using the C bindings.

jcgruenhage:
there is someone trying to implement opus in pure rust: [est31/werk](https://github.com/est31/werk)

teknico:
Another possibility is using the [duckerman/magnum-opus](https://github.com/duckerman/magnum-opus) safe bindings.

Beiri22:
Are there any new regarding opus support?

vlyr:
Would have use for it too.

CutestNekoAqua:
I think we need a contributor to tell us what would be a good way to add it to the library so we can work on doing it.

schneiderfelipe:
Maybe relevant: https://github.com/SpaceManiac/opus-rs/pull/20

dvdsk:
> I think we need a contributor to tell us what would be a good way to add it to the library so we can work on doing it.

we now have a contributors guide, see: https://github.com/RustAudio/rodio/blob/master/CONTRIBUTING.md#implementing-a-new-decoder if anyone wants to work on this we will be happy to help out.

PR #790, opened 2025-08-29
add wav output method which takes writer

dvdsk:
This will enable sending the wav somewhere else then the local filesystem. Examples are: send over tcp/http, collect multiple in a tar (my usecase).

### DIFF
--- a/src/wav_output.rs
+++ b/src/wav_output.rs
@@ -92,20 +92,20 @@ pub fn wav_to_writer(
 
 #[cfg(test)]
 mod test {
-    use super::output_to_wav;
+    use super::wav_to_file;
     use crate::Source;
     use std::io::BufReader;
     use std::time::Duration;
 
     #[test]
-    fn test_output_to_wav() {
+    fn test_wav_to_file() {
         let make_source = || {
             crate::source::SineWave::new(745.0)
                 .amplify(0.1)
                 .take_duration(Duration::from_secs(1))
         };
         let wav_file_path = "target/tmp/save-to-wav-test.wav";
-        output_to_wav(&mut make_source(), wav_file_path).expect("output file can be written");
+        wav_to_file(&mut make_source(), wav_file_path).expect("output file can be written");
 
         let file = std::fs::File::open(wav_file_path).expect("output file can be opened");
         // Not using crate::Decoder bcause it is limited to i16 samples.
--- a/src/wav_output.rs
+++ b/src/wav_output.rs
@@ -2,21 +2,24 @@ use crate::common::assert_error_traits;
 use crate::Source;
 use hound::{SampleFormat, WavSpec};
 use std::path;
+use std::io;
 use std::sync::Arc;
 
 #[derive(Debug, thiserror::Error, Clone)]
 pub enum ToWavError {
-    #[error("Could not create wav file")]
+    #[error("Opening file for writing")]
+    OpenFile(#[source] Arc<std::io::Error>),
+    #[error("Could not create wav writer")]
     Creating(#[source] Arc<hound::Error>),
-    #[error("Failed to write samples to wav file")]
+    #[error("Failed to write samples writer")]
     Writing(#[source] Arc<hound::Error>),
     #[error("Failed to update the wav header")]
     Finishing(#[source] Arc<hound::Error>),
 }
 assert_error_traits!(ToWavError);
 
-/// This procedure saves Source's output into a wav file. The output samples format is 32-bit float.
-/// This function is intended primarily for testing and diagnostics. It can be used to see
+/// Saves Source's output into a wav file. The output samples format is 32-bit
+/// float. This function is intended primarily for testing and diagnostics. It can be used to see
 /// the output without opening output stream to a real audio device.
 ///
 /// If the file already exists it will be overwritten.
@@ -24,13 +27,27 @@ pub fn output_to_wav(
     source: &mut impl Source,
     wav_file: impl AsRef<path::Path>,
 ) -> Result<(), ToWavError> {
+    let file = std::fs::File::create(wav_file)
+        .map_err(Arc::new)
+        .map_err(ToWavError::OpenFile)?;
+    collect_to_wav(source, file)
+}
+
+/// Saves Source's output into a writer. The output samples format is 32-bit float. This function
+/// is intended primarily for testing and diagnostics. It can be used to see the output without
+/// opening output stream to a real audio device.
+pub fn collect_to_wav(
+    source: &mut impl Source,
+    writer: impl io::Write + io::Seek,
+) -> Result<(), ToWavError> {
     let format = WavSpec {
         channels: source.channels().get(),
         sample_rate: source.sample_rate().get(),
         bits_per_sample: 32,
         sample_format: SampleFormat::Float,
     };
-    let mut writer = hound::WavWriter::create(wav_file, format)
+    let writer = io::BufWriter::new(writer);
+    let mut writer = hound::WavWriter::new(writer, format)
         .map_err(Arc::new)
         .map_err(ToWavError::Creating)?;
     for sample in source {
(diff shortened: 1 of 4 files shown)


yara-blue:
needs
- [x] better name
- [x] change log entry
- [x] possibly rename the existing one

src/wav_output.rs
```diff
@@ -2,35 +2,68 @@ use crate::common::assert_error_traits;
...
 
 #[derive(Debug, thiserror::Error, Clone)]
 pub enum ToWavError {
-    #[error("Could not create wav file")]
+    #[error("Opening file for writing")]
+    OpenFile(#[source] Arc<std::io::Error>),
+    #[error("Could not create wav writer")]
     Creating(#[source] Arc<hound::Error>),
-    #[error("Failed to write samples to wav file")]
+    #[error("Failed to write samples writer")]
```
roderickvd, line 15: "Failed to write samples **to** writer"

src/wav_output.rs
```diff
@@ -2,35 +2,68 @@ use crate::common::assert_error_traits;
...
-/// This procedure saves Source's output into a wav file. The output samples format is 32-bit float.
-/// This function is intended primarily for testing and diagnostics. It can be used to see
+/// Saves Source's output into a wav file. The output samples format is 32-bit
+/// float. This function is intended primarily for testing and diagnostics. It can be used to see
 /// the output without opening output stream to a real audio device.
 ///
 /// If the file already exists it will be overwritten.
 pub fn output_to_wav(
-    source: &mut impl Source,
+    source: impl Source,
```
roderickvd: Should document this as a breaking change.

src/wav_output.rs
```diff
@@ -2,35 +2,68 @@ use crate::common::assert_error_traits;
...
+    writer: &mut (impl io::Write + io::Seek),
 ) -> Result<(), ToWavError> {
     let format = WavSpec {
         channels: source.channels().get(),
         sample_rate: source.sample_rate().get(),
         bits_per_sample: 32,
         sample_format: SampleFormat::Float,
     };
-    let mut writer = hound::WavWriter::create(wav_file, format)
+    let writer = io::BufWriter::new(writer);
```
roderickvd: I've never really worked with `BufWriter` before, but I believe there are some caveats to dropping and flushing:

> It is critical to call [flush](https://doc.rust-lang.org/std/io/struct.BufWriter.html#method.flush) before BufWriter<W> is dropped.

src/wav_output.rs
```diff
@@ -2,35 +2,68 @@ use crate::common::assert_error_traits;
...
+    source: impl Source,
     wav_file: impl AsRef<path::Path>,
+) -> Result<(), ToWavError> {
+    let mut file = std::fs::File::create(wav_file)
+        .map_err(Arc::new)
+        .map_err(ToWavError::OpenFile)?;
+    collect_to_wav(source, &mut file)
+}
+
+/// Saves Source's output into a writer. The output samples format is 32-bit float. This function
```
roderickvd, line 42: Might want to document that this wraps in a `BufWriter`.

Would there be use cases where you'd actually _not_ want buffered writing? e.g. when writing to a real-time stream, or a TCP stream that by itself already batches?

Alternative strategies:
1. offer a vanilla `collect_to_wav` and `collect_to_wav_buffered` (or vice versa: `unbuffered`)
2. document and suggest that the user wraps a `BufWriter` himself.

roderickvd (commented):
What are you recording? 😄

src/wav_output.rs
yara-blue: I did not realize that, makes sense. Good catch though!

src/wav_output.rs
yara-blue: > When writing to a real-time stream, or a TCP stream that by itself already batches?

Is real time streaming a use-case for wav?

1. Seems like the best approach but it does entail adding yet a third function for outputting to wav..... At this point do we want a builder? Are there other output formats we can and want to support?
2. We'll as we've seen with me not reading the BufWriter docs that does not often end well. :laughing:

yara-blue:
> What are you recording? 😄

Improving the VOIP in zed, needed a way to write recent audio to disk in case we run into audio issues. Need the writer call since I want to record each user in a call to a single tar file. (Which the user triggering the record can then send me to debug)

src/wav_output.rs
roderickvd: > > When writing to a real-time stream, or a TCP stream that by itself already batches?
> 
> Is real time streaming a use-case for wav?

Though I couldn't say it really is, I got triggered by your PR description saying "send over tcp/http".

> 1. Seems like the best approach but it does entail adding yet a third function for outputting to wav..... At this point do we want a builder?

Though I think we both like builder patterns, and it'd be consistent... what other settings do we foresee that would make it worth it now?

> Are there other output formats we can and want to support?

You mean besides WAV? It'd be amazing if we could turn it into a transcoder with flexible output format, but I don't know of other pure Rust audio encoding crates.

[mp3lame-encoder](https://github.com/doumanash/mp3lame-encoder) does what it says on the tin but is licensed under LGPL.

roderickvd:
\<offtopic> You are contributing to Zed and using VoIP calls? Cool, I'm a Zed user too. \</offtopic>

src/wav_output.rs
yara-blue: > Though I couldn't say it really is, I got triggered by your PR description saying "send over tcp/http".

Ahhh yes, I was thinking of uploading a file.

> Though I think we both like builder patterns, and it'd be consistent... what other settings do we foresee that would make it worth it now?

Different encoders for example. We could have `Source::outputbuilder`. The builder would then allow specifying output encoding etc. But lets not do that since there only is mp3lame and no request for it yet.

yara-blue:
> You are contributing to Zed and using VoIP calls? Cool, I'm a Zed user too.

Yeah I joined their team last month. Really like the collaborative editing mode, its so much nicer to pair on a PR then use gits terrible interface. We should try that for your giant PR (if you're up for it).

roderickvd:
> > You are contributing to Zed and using VoIP calls? Cool, I'm a Zed user too.
> 
> Yeah I joined their team last month. Really like the collaborative editing mode, its so much nicer to pair on a PR then use gits terrible interface. We should try that for your giant PR (if you're up for it).

Never used it before, but sure! Only question I have is how it works with beer.

src/wav_output.rs
roderickvd: > Different encoders for example. We could have `Source::outputbuilder`. The builder would then allow specifying output encoding etc. But lets not do that since there only is mp3lame and no request for it yet.

Searching two minutes more, I see more opportunities:
1. https://github.com/yotarok/flacenc-rs
2. https://github.com/SpaceManiac/opus-rs
3. https://github.com/ComunidadAylas/vorbis-rs

Ad 2) Should add a decoder for it first though: #38. Symphonia is working on one too: https://github.com/pdeljanov/Symphonia/pull/398

Ad 3) There seem to be more Vorbis crates, didn't check which ones are the best maintained and have the right feature set.

All in all I'm thinking: why not get the API right now and do the builder thing. Then it'll be easy to add other encoders later when someone does request it or we feel like doing it.

yara-blue:
few notes on the builder, should be typed as in:
configuring `.wav` on the builder should return a `WavOutputBuilder`. That builder then has the specific options relevant to wav.

src/wav_output.rs
yara-blue: > All in all I'm thinking: why not get the API right now and do the builder thing. Then it'll be easy to add other encoders later when someone does request it or we feel like doing it.

In principle I agree, in reality I simply do not have the time as other things (merging your PR, spanless rodio) are more important. I needed wav to writer so I added it. I now have 2 projects using this Rodio branch.

Lets take the notes we learned split that off into an enhancement issue and merge this. If before the next release someone implements a builder great! If they dont no one probably will in the foreseeable future.

PR #794, opened 2025-09-12
feat: add audio dithering support

roderickvd:
Builds on our noise generators to add four dithering algorithms:
- TPDF (Triangular PDF) - Optimal decorrelation (default)
- RPDF (Rectangular PDF) - Lower noise floor, worse decorrelation
- GPDF (Gaussian PDF) - Higher noise floor, analog characteristics
- HighPass - Blue noise to reduce low-frequency artifacts

Gated by new `dither` feature flag, that's enabled by default.

## API

```rust
use rodio::source::{dither, DitherAlgorithm};
use rodio::BitDepth;

let source = SineWave::new(440.0);
let dithered = dither(source, BitDepth::new(16).unwrap(), DitherAlgorithm::TPDF);
```

## Parent branch

This is branched off of #786 which introduces `Source::bits_per_sample()`, because you'll typically want to dither down to:

```rust
let output_bits = config.bits_per_sample();  // in cpal master: https://github.com/RustAudio/cpal/pull/1008
let target_bits = source_bits.map_or(output_bits, |src_bits| src_bits.min(output_bits));
```

I'll mark this PR for review when that PR is merged.

### DIFF
--- a/src/source/dither.rs
+++ b/src/source/dither.rs
@@ -36,76 +36,32 @@ use crate::{
     BitDepth, ChannelCount, Sample, SampleRate, Source,
 };
 
-/// Dither algorithm selection - chooses the probability density function (PDF).
-#[derive(Debug, Default, Clone, Copy, PartialEq, Eq)]
-pub enum Algorithm {
-    /// GPDF (Gaussian PDF) - normal/bell curve distribution.
-    ///
-    /// Uses Gaussian white noise which more closely mimics natural processes and
-    /// analog circuits. Higher noise floor than TPDF.
-    GPDF,
-
-    /// High-pass dithering - reduces low-frequency artifacts.
-    ///
-    /// Uses blue noise (high-pass filtered white noise) to push dither energy
-    /// toward higher frequencies. Particularly effective for reducing audible
-    /// low-frequency modulation artifacts. Best for material with significant
-    /// low-frequency content where traditional white dither might be audible.
-    HighPass,
-
-    /// RPDF (Rectangular PDF) - uniform distribution.
-    ///
-    /// Uses uniform white noise for basic decorrelation. Simpler than TPDF but
-    /// allows some correlation between signal and quantization error at low levels.
-    /// Slightly lower noise floor than TPDF.
-    RPDF,
-
-    /// TPDF (Triangular PDF) - triangular distribution.
-    ///
-    /// The gold standard for audio dithering. Provides mathematically optimal
-    /// decorrelation by completely eliminating correlation between the original
-    /// signal and quantization error.
-    #[default]
-    TPDF,
-}
-
-/// Internal dithering implementation with a specific noise generator type.
-#[derive(Clone, Debug)]
-pub struct DitherImpl<I, N> {
-    input: I,
-    noise: N,
-    lsb_amplitude: f32,
-}
-
-impl<I, N> DitherImpl<I, N>
-where
-    I: Source,
-    N: Iterator<Item = Sample>,
-{
-    /// Creates a new dither source with a custom noise generator.
-    ///
-    /// This low-level internal constructor allows providing a custom noise generator.
-    /// The noise generator should produce samples with appropriate amplitude
-    /// for the chosen dither type.
-    #[inline]
-    pub(crate) fn new_with_noise(input: I, noise: N, target_bits: BitDepth) -> Self {
-        // LSB amplitude for signed audio: 1.0 / (2^(bits-1))
-        // This represents the amplitude of one quantization level
-        let lsb_amplitude = if target_bits.get() >= Sample::MANTISSA_DIGITS {
-            // For bit depths at or beyond the floating point precision limit,
-            // the LSB amplitude calculation becomes meaningless
-            Sample::MIN_POSITIVE
-        } else {
-            1.0 / (1_i64 << (target_bits.get() - 1)) as f32
-        };
-
-        Self {
-            input,
-            noise,
-            lsb_amplitude,
-        }
-    }
-}
+// /// Dither algorithm selection - chooses the probability density function (PDF).
+// #[derive(Debug, Default, Clone, Copy, PartialEq, Eq)]
+// pub enum Algorithm {
+//     /// High-pass dithering - reduces low-frequency artifacts.
+//     ///
+//     /// Uses blue noise (high-pass filtered white noise) to push dither energy
+//     /// toward higher frequencies. Particularly effective for reducing audible
+//     /// low-frequency modulation artifacts. Best for material with significant
+//     /// low-frequency content where traditional white dither might be audible.
+//     HighPass,
+//
+//     /// RPDF (Rectangular PDF) - uniform distribution.
+//     ///
+//     /// Uses uniform white noise for basic decorrelation. Simpler than TPDF but
+//     /// allows some correlation between signal and quantization error at low levels.
+//     /// Slightly lower noise floor than TPDF.
+//     RPDF,
+//
+//     /// TPDF (Triangular PDF) - triangular distribution.
+//     ///
+//     /// The gold standard for audio dithering. Provides mathematically optimal
+//     /// decorrelation by completely eliminating correlation between the original
+//     /// signal and quantization error.
+//     #[default]
+//     TPDF,
+// }
 
 impl<I, N> Iterator for DitherImpl<I, N>
 where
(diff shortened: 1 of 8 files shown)


yara-blue:
> ## Parent branch
> 
> This is branched off of #786 which introduces `Source::bits_per_sample()`, because you'll typically want to dither down to:

:sweat_smile: :sweat_smile:  For a minute or two I thought you introduced another ~6k PR.

Btw lets add a new *rule*/*convention* to keep PR's under 1500 lines. I'll hold myself to that for rewriting the entire audio pipeline.

src/source/mod.rs
```diff
@@ -696,7 +740,7 @@ pub trait Source: Iterator<Item = Sample> {
     /// the source is not known.
     #[allow(unused_variables)]
     fn try_seek(&mut self, pos: Duration) -> Result<(), SeekError> {
-        Err(SeekError::NotSupported {
+        Err(SeekError::SeekingNotSupported {
             underlying_source: std::any::type_name::<Self>(),
         })
```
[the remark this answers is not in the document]
roderickvd: Nonsense.

src/source/dither.rs
```diff
@@ -0,0 +1,379 @@
...
+    ///
+    /// This low-level internal constructor allows providing a custom noise generator.
+    /// The noise generator should produce samples with appropriate amplitude
+    /// for the chosen dither type.
+    #[inline]
+    pub(crate) fn new_with_noise(input: I, noise: N, target_bits: BitDepth) -> Self {
+        // LSB amplitude for signed audio: 1.0 / (2^(bits-1))
+        // This represents the amplitude of one quantization level
+        // Use i64 bit shifting to avoid overflow (supports up to 63 bits)
+        let lsb_amplitude = 1.0 / (1_i64 << (target_bits.get() - 1)) as f32;
```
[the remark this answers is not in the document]
roderickvd: Something like that in [a00fe4c](https://github.com/RustAudio/rodio/pull/794/commits/a00fe4c604d91c3f9cdbd649ddfa9c1675106ae0).

roderickvd:
Rebased on master.

src/source/dither.rs
```diff
@@ -0,0 +1,367 @@
...
+use rand::{rngs::SmallRng, Rng, SeedableRng};
+
+use crate::{
+    source::noise::{Blue, WhiteGaussian, WhiteTriangular, WhiteUniform},
+    BitDepth, ChannelCount, Sample, SampleRate, Source,
+};
+
+/// Dither algorithm selection - chooses the probability density function (PDF).
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq)]
+pub enum Algorithm {
```
yara-blue, line 38: I've spend some time trying to learn about these. I find it quite hard to find info on these. Would love to know (out of intrest, this is *not* blocking the PR in any way) how you got to these docs. Is it something you've learned over the years?

src/source/dither.rs
```diff
@@ -0,0 +1,367 @@
...
+    ///
+    /// This low-level internal constructor allows providing a custom noise generator.
+    /// The noise generator should produce samples with appropriate amplitude
+    /// for the chosen dither type.
+    #[inline]
+    pub(crate) fn new_with_noise(input: I, noise: N, target_bits: BitDepth) -> Self {
+        // LSB amplitude for signed audio: 1.0 / (2^(bits-1))
+        // This represents the amplitude of one quantization level
+        let lsb_amplitude = if target_bits.get() >= Sample::MANTISSA_DIGITS {
+            // For bit depths at or beyond the floating point precision limit,
```
yara-blue: Is the idea to not do the LSB amplitude calc. as a performance improvement? Or do we need to skip it because we get artifacting otherwise?

My gut tells me that the extra branch from the if will be more expensive then the calculation.

src/source/dither.rs
roderickvd: I think you will love reading Wannamaker's PhD seminal [dissertation](https://www.robertwannamaker.com/writings/rw_phd.pdf) on this. I believe [this](https://robertwannamaker.com/writings/ieee.pdf) is a shorter IEEE extract of it, but it's been years since I read all these. Yes, I actually dove into it like four years ago reading these, as well as source code of GStreamer, PortAudio and others.

Beyond high-passed dither there are more elaborate [noise-shaping](https://www.researchgate.net/publication/242019172_Psychoacoustically_Optimal_Noise_Shaping) techniques that I _may_ propose in a later PR. Emphasis on _may_.

src/source/dither.rs
```diff
@@ -0,0 +1,367 @@
...
+
+    #[inline]
+    fn try_seek(&mut self, pos: Duration) -> Result<(), crate::source::SeekError> {
+        self.input.try_seek(pos)
+    }
+}
+
+/// Dithering interface delegating to the supported dithering algorithms.
+#[derive(Clone)]
+pub enum Dither<I, R = SmallRng>
```
yara-blue: How about we merge this and the DitherImpl struct? That would get rid of the match in channels, sample_rate etc.

src/source/dither.rs
roderickvd: This branch is in the constructor, not the hot path.

The point is that some sources may report a bit depth of 32 that users would want to dither to for requantization (remember, this would be great to use with the `Source::bits_per_sample()` that #786 introduced). Because our audio pipeline in the end will be constrained to 24 bits, we then still want to select the LSB and dither before truncating.

src/source/dither.rs
roderickvd: How would that look like while still supporting the various PDFs and so, different noise generators? I want to prevent branching in the iterator. This delegation is faster, particularly when the algorithm is selected at compile time.

roderickvd:
> I think we can make `source/dither` smaller by merging the `Dither` and `DitherImpl` structs. Let me know what you think.

Thanks for the review - please check out my question about that. I'm not sure.

yara-blue:
> > 
> 
> Thanks for the review - please check out my question about that. I'm not sure.

I'll give a try and then we can compare :+1:

src/source/dither.rs
yara-blue: > This delegation is faster, particularly when the algorithm is selected at compile time.

Thats surprising since there is a match in the iterator. That will branch.

yara-blue: > This delegation is faster

Do we have a benchmark to back this up?

yara-blue:
I pushed a messy idea/prototype (we got a builder again). This gets rid of the branching and should end up a lot shorter. It does make it impossible to change dithering method on the fly but I dont think you would want to do that right? Feel free to revert if we decide not to go this route.

src/source/dither.rs
roderickvd: > > This delegation is faster, particularly when the algorithm is selected at compile time.
> 
> Thats surprising since there is a match in the iterator. That will branch.

`match` creates jump tables which are O(1). `if` branches are O(n).

roderickvd:
I do want the ergonomics of `sound.dither().amplify()`. Sure, we could have `Source::dither()` select `DitherBuilder::tpdf()` by default. But I think builders make most sense when they are "point of origin" like with the decoder or microphone. I do agree that this proposal shaves off a lot of code and should be even faster.

I don't immediately have a strong feeling either way. What do you say looking at that source-and-effect chain?

src/source/dither.rs
yara-blue: > match creates jump tables which are O(1). if branches are O(n).

On a modern optimizing compiler this is no longer true. But its a moot point as the number of comparisons is not at all what can kill performance with branches.

Branches prevent in-lining and a ton of other essential compiler optimizations. Further more if the CPU's branch predictor makes an incorrect prediction the entire pipeline needs to be flushed. 

For a for loop (basically a loop { if i > num break else do things} }this is not an issue since the branch predictor can figure out that if statement will usually be true. That means only once we go into the else we pay the price.

But as always compilers and CPU's have become far to complex for anyone to reason about performance in detail. So optimize for readability and benchmark after.

yara-blue:
> I don't immediately have a strong feeling either way. What do you say looking at that source-and-effect chain?

New idea I'll have to work it out. I'll send another very ugly commit your way.

yara-blue:
Curious what you think, I quite like it. No branching on algorithm shorter code and no builder :tada:

roderickvd:
Your proposal is clever, however cannot select a PDF at runtime. There sure would be reasons why someone would want that, like in a dialog or on the command line to configure an application’s behavior. 

I benchmarked the branching and monomorphization approaches with `divan` and the performance differences are negligible: 0.01ns  delta on my Apple M3. I’m gonna propose a middle ground: remove the `DitherInner` enum and replace it with a `NoiseGenerator` enum that dispatches. It does save a lot of code.

yara-blue:
> I benchmarked the branching and monomorphization approaches with `divan` and the performance differences are negligible: 0.01ns delta on my Apple M3.

The modern optimizing compiler, a true miracle.

> Your proposal is clever, however cannot select a PDF at runtime. There sure would be reasons why someone would want that, like in a dialog or on the command line to configure an application’s behavior.

:+1:

src/source/mod.rs
```diff
@@ -85,6 +85,26 @@ mod triangle;
...
+#[cfg_attr(docsrs, doc(cfg(feature = "dither")))]
+pub use self::dither::{Algorithm as DitherAlgorithm, Dither};
+
+/// Creates a dithered source using the specified algorithm.
+///
+/// Dithering eliminates quantization artifacts during digital audio playback
+/// and when converting between bit depths. Apply at the target output bit depth.
+#[cfg(feature = "dither")]
+#[cfg_attr(docsrs, doc(cfg(feature = "dither")))]
+pub fn dither<I>(input: I, target_bits: BitDepth, algorithm: DitherAlgorithm) -> Dither<I>
```
yara-blue: replace this with a default implementation on source? So you can do `sine().dither().amplify().pausable()` etc.

yara-blue: otherwise good to merge btw

roderickvd: Huh that must have fallen off somewhere! Good catch, will re-add it and merge when done. Thanks.

roderickvd:
> > I benchmarked the branching and monomorphization approaches with `divan` and the performance differences are negligible: 0.01ns delta on my Apple M3.
> 
> The modern optimizing compiler, a true miracle.

I forgot to say "0.01ns" **per sample**, but yes.

Besides the compiler there's speculative branch prediction in CPUs as well of course. You're right it's nigh impossible reason about it, and then it differs per architecture too. Though lately I'm making more effort to try and feed it code that at least *should* be easy to digest in the hot path.

PR #805, opened 2025-10-04
Add Symphonia SIMD flag

### LINKED ISSUE #804: Add flag to suport SIMD with symphonia
Presently it does not appear possible to use the SIMD feature of Symphonia via Rodio, it would be great to have a flag on Rodio that in turn enables the flag on Symphonia.

keirlawson:
Fixes #804


yara-blue:
Thank you!

PR #808, opened 2025-10-11
feat: Add optional floor value to AGC and performance optimizations

UnknownSuperficialNight:
This PR introduces an optional floor value for the AGC, enabling users to set a minimum output level that prevents the gain from dropping below a specified threshold.

For example, setting a floor value of `1.0` ensures that the audio output is always at or above the original source level, preventing excessive attenuation.

It also updates the example AGC to use the new defaults, I noticed forgot to do that in #759

### DIFF
--- a/examples/automatic_gain_control.rs
+++ b/examples/automatic_gain_control.rs
@@ -1,4 +1,4 @@
-use rodio::source::Source;
+use rodio::source::{AutomaticGainControlSettings, Source};
 use rodio::Decoder;
 use std::error::Error;
 use std::fs::File;
@@ -16,7 +16,7 @@ fn main() -> Result<(), Box<dyn Error>> {
     let source = Decoder::try_from(file)?;
 
     // Apply automatic gain control to the source
-    let agc_source = source.automatic_gain_control(1.0, 4.0, 0.0, 5.0);
+    let agc_source = source.automatic_gain_control(AutomaticGainControlSettings::default());
 
     // Make it so that the source checks if automatic gain control should be
     // enabled or disabled every 5 milliseconds. We must clone `agc_enabled`,
--- a/src/source/agc.rs
+++ b/src/source/agc.rs
@@ -14,6 +14,7 @@
 //
 
 use super::SeekError;
+use crate::math::duration_to_coefficient;
 use crate::Source;
 #[cfg(feature = "experimental")]
 use atomic_float::AtomicF32;
@@ -40,6 +41,37 @@ const fn power_of_two(n: usize) -> usize {
 /// A larger size provides more stable RMS values but increases latency.
 const RMS_WINDOW_SIZE: usize = power_of_two(8192);
 
+/// Settings for the Automatic Gain Control (AGC).
+///
+/// This struct contains parameters that define how the AGC will function,
+/// allowing users to customise its behaviour.
+#[derive(Debug, Clone)]
+pub struct AutomaticGainControlSettings {
+    /// The desired output level that the AGC tries to maintain.
+    /// A value of 1.0 means no change to the original level.
+    pub target_level: f32,
+    /// Time constant for gain increases (how quickly the AGC responds to level increases).
+    /// Longer durations result in slower, more gradual gain increases.
+    pub attack_time: Duration,
+    /// Time constant for gain decreases (how quickly the AGC responds to level decreases).
+    /// Shorter durations allow for faster response to sudden loud signals.
+    pub release_time: Duration,
+    /// Maximum allowable gain multiplication to prevent excessive amplification.
+    /// This acts as a safety limit to avoid distortion from over-amplification.
+    pub absolute_max_gain: f32,
+}
+
+impl Default for AutomaticGainControlSettings {
+    fn default() -> Self {
+        AutomaticGainControlSettings {
+            target_level: 1.0,                           // Default to original level
+            attack_time: Duration::from_secs_f32(4.0),   // Recommended attack time
+            release_time: Duration::from_secs_f32(0f32), // Recommended release time
+            absolute_max_gain: 7.0,                      // Recommended max gain
+        }
+    }
+}
+
 #[cfg(feature = "experimental")]
 /// Automatic Gain Control filter for maintaining consistent output levels.
 ///
@@ -49,12 +81,11 @@ const RMS_WINDOW_SIZE: usize = power_of_two(8192);
 pub struct AutomaticGainControl<I> {
     input: I,
     target_level: Arc<AtomicF32>,
-    floor: Option<f32>,
+    floor: f32,
     absolute_max_gain: Arc<AtomicF32>,
     current_gain: f32,
     attack_coeff: Arc<AtomicF32>,
     release_coeff: Arc<AtomicF32>,
-    min_attack_coeff: f32,
     peak_level: f32,
     rms_window: CircularBuffer,
     is_enabled: Arc<AtomicBool>,
@@ -69,12 +100,11 @@ pub struct AutomaticGainControl<I> {
 pub struct AutomaticGainControl<I> {
     input: I,
     target_level: f32,
-    floor: Option<f32>,
+    floor: f32,
     absolute_max_gain: f32,
     current_gain: f32,
     attack_coeff: f32,
     release_coeff: f32,
-    min_attack_coeff: f32,
     peak_level: f32,
     rms_window: CircularBuffer,
     is_enabled: bool,
@@ -96,8 +126,8 @@ impl CircularBuffer {
     #[inline]
     fn new() -> Self {
         CircularBuffer {
-            buffer: Box::new([0.0; RMS_WINDOW_SIZE]),
-            sum: 0.0,
+            buffer: Box::new([0f32; RMS_WINDOW_SIZE]),
+            sum: 0f32,
             index: 0,
         }
     }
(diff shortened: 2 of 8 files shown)


src/source/agc.rs
```diff
@@ -383,7 +400,14 @@ where
         let peak_gain = self.calculate_peak_gain();
 
         // Use RMS for general adjustments, but limit by peak gain to prevent clipping
-        let desired_gain = rms_gain.min(peak_gain);
+        let mut desired_gain = rms_gain.min(peak_gain);
+
+        // Apply the floor value to set a minimum output level
+        if let Some(floor_value) = self.floor {
+            if desired_gain < floor_value {
```
yara-blue: we could write this as: `desired_gain = desired_gain.min(floor_value.unwrap_or(0f32))`. We could then even chain it onto the `rms_gain.min` expression.

examples/automatic_gain_control.rs
```diff
@@ -16,7 +16,7 @@ fn main() -> Result<(), Box<dyn Error>> {
     let source = Decoder::try_from(file)?;
 
     // Apply automatic gain control to the source
-    let agc_source = source.automatic_gain_control(1.0, 4.0, 0.005, 5.0);
+    let agc_source = source.automatic_gain_control(1.0, 4.0, 0.0, 5.0);
```
yara-blue: (only if you want to) 

You could rewrite this to take a struct as parameter and then have a default implementation for that struct with these parameters. See the limiter for a great example on what I mean.

src/source/agc.rs
UnknownSuperficialNight: Good point

examples/automatic_gain_control.rs
UnknownSuperficialNight: You mean something like

```Rust
#[derive(Debug, Clone)]
pub struct AutomaticGainControlSettings {
    pub target_level: f32,
    pub attack_time: f32,
    pub release_time: f32,
    pub absolute_max_gain: f32,
}

impl Default for AutomaticGainControlSettings {
    fn default() -> Self {
        AutomaticGainControlSettings {
            target_level: 1.0,      // Default to original level
            attack_time: 4.0,       // Recommended attack time
            release_time: 0.0,      // Recommended release time
            absolute_max_gain: 5.0, // Recommended max gain
        }
    }
}
```

Then just put that in the public facing API?

UnknownSuperficialNight:
> Code nitpick, please have a look and let me know what you think.

Should we also remove most of the docs for [this](https://github.com/RustAudio/rodio/blob/836c0133533ad89f6df08b584d1e4c7269744062/src/source/mod.rs#L395) and add links to the struct and the examples?

Since i don't think we need this anymore

```
/// ```rust
/// // Apply Automatic Gain Control to the source (AGC is on by default)
/// use rodio::source::{Source, SineWave};
/// use rodio::Sink;
/// let source = SineWave::new(444.0); // An example.
/// let (sink, output) = Sink::new(); // An example.
///
/// let agc_source = source.automatic_gain_control(1.0, 4.0, 0.0, 5.0);
///
/// // Add the AGC-controlled source to the sink
/// sink.append(agc_source);
///
/// ```
```

roderickvd:
Nice to see your active maintenance on this one! As you're updating anyway, may I put in a couple of thoughts as well?

1. Change the `attack_time` and `release_time` in the public-facing API with `Duration` - more idiomatic.
2. It seems the `attack_coeff` and `release_coeff` calculation could reuse `duration_to_coefficient` from `src/source/limit.rs` - it should probably be moved to `src/math.rs` with `pub(crate)` visibility.

And three questions:

3. How correct is the logic of `min_attack_coeff`? The name implies it contains a coefficient, but instead it contains a duration as float in seconds.
4. What are the intended use cases for the recommended values? Maybe they're good for speech, I don't know, but I've never had success with them on music with large dynamic range and rather fast transients like classical.
5. `AutomaticGainControl` is the only user of the `experimental` feature. Could we by now consider it non-experimental and remove the feature flag?

examples/automatic_gain_control.rs
yara-blue: Yeah that seems perfect. You could add some defaults but I would have no idea what to call them. If it helps, these are parameters that work well for voip [microphone input](https://github.com/zed-industries/zed/blob/abe1fd5e16c9a9c3db84648f56336c7f25df3ddb/crates/audio/src/audio.rs#L208) and [speaker output](https://github.com/zed-industries/zed/blob/abe1fd5e16c9a9c3db84648f56336c7f25df3ddb/crates/audio/src/audio.rs#L238).

both: automatic_gain_control(0.90, 1.0, 0.0, 5.0)

that got us about -18dbA which seems to be the recommended gain to prevent clipping.

yara-blue:
> Should we also remove most of the docs for [this](https://github.com/RustAudio/rodio/blob/836c0133533ad89f6df08b584d1e4c7269744062/src/source/mod.rs#L395) and add links to the struct and the examples?

I would keep the docs on the source member and link from the struct actually. Since users will generally use AGC like this:

```rust
let stream = stream
    .possibly_disconnected_channels_to_mono()
    .constant_samplerate(SAMPLE_RATE)
    .limit(LimitSettings::live_performance())
    .process_buffer::<BUFFER_SIZE, _>(move |buffer| {
       <secret things> // not really look at zed/crates/audio/src/audio.rs if your curious
    })
    .denoise()
    .context("Could not set up denoiser")?
    .automatic_gain_control(0.90, 1.0, 0.0, 5.0)
    .periodic_access(Duration::from_millis(100), move |agc_source| {
        agc_source
            .set_enabled(LIVE_SETTINGS.auto_microphone_volume.load(Ordering::Relaxed));
        let denoise = agc_source.inner_mut();
        denoise.set_enabled(LIVE_SETTINGS.denoise.load(Ordering::Relaxed));
    });

```

yara-blue:
> 5\. `AutomaticGainControl` is the only user of the `experimental` feature. Could we by now consider it non-experimental and remove the feature flag?

I really really do not want to expand rodio's API/patterns if we can. As I am actively working on rewriting/re-architecturing the audio pipeline. The larger our API the harder it well get to re-do it. Right now we only really support periodic access as parameter change method. If we stabilize the use of atomics/handles as is experimental here it will become far harder to make the changes I need to make.

UnknownSuperficialNight:
>Nice to see your active maintenance on this one 

:)

> Change the attack_time and release_time in the public-facing API with Duration - more idiomatic.
>
>It seems the attack_coeff and release_coeff calculation could reuse duration_to_coefficient from src/source/limit.rs - it should probably be moved to src/math.rs with pub(crate) visibility.

I agree… Done.


> How correct is the logic of min_attack_coeff? The name implies it contains a coefficient, but instead it contains a duration as float in seconds.

Not very correct I looked more into it, we don't need a `min_attack_coeff` we can just use `0.0` as it looks like I missed it during development but `min_attack_coeff` was always `0.0` by default or very close too.

We already have:
```Rust
let attack_speed = if desired_gain > self.current_gain {
    attack_coeff
} else {
    release_coeff
};
```

To do this so peak doing the same thing for `if sample_value > self.peak_level {` is irrelevant so ill instead replace it with `0f32` to be as fast as possible. We still want `release_coeff` to be considered during the peak calculations to stop the peak from going down too rapidly.

> What are the intended use cases for the recommended values? Maybe they're good for speech, I don't know, but I've never had success with them on music with large dynamic range and rather fast transients like classical.

Maybe we should work together on the defaults because for me, it's working fine, see attached video:


https://github.com/user-attachments/assets/63dc002d-87c2-45d4-9a1c-e388b8dbf7f6

But your issue with `music with large dynamic range and rather fast transients like classical` might have been due to `update_peak_level` accounting for `attack_coeff` when it shouldn't

> AutomaticGainControl is the only user of the experimental feature. Could we by now consider it non-experimental and remove the feature flag?

@dvdsk responded to this, I don't feel that I have sufficient knowledge on the current state to provide valuable input on that.

Added some optimisations for AGC:

Increased default `absolute_max_gain` to `7.0` as this is the default i personally use (In my program) and there is no reason not too (if there is an issue with this let me know)

### 1. **Cached Atomic Loads**
- **What:** Load `target_level()`, `absolute_max_gain()`, `attack_coeff()`, `release_coeff()` once at the start of `process_sample()`
- **Why:** Each call performs an atomic load (with experimental feature), so calling them multiple times per sample was wasteful
- **Impact:** Reduced atomic loads per sample to just 4 total

### 2. **Optimized Floor Storage**
- **What:** Changed `floor: Option<f32>` to `floor: f32`
- **Why:** Unwrap happens once in `set_floor()` instead of `unwrap_or(0f32)` on every single sample
- **Impact:** Eliminates one branch + Option check per sample

### 3. **Modified Helper Methods with Parameters**
- **What:** Helper methods now accept cached values as parameters:
  - `update_peak_level(sample_value, release_coeff)`
  - `calculate_peak_gain(target_level, absolute_max_gain)`
- **Why:** Previously, `update_peak_level()` would call `self.release_coeff()` internally, causing another atomic load. Now it uses the cached parameter.
- **Impact:** Prevents redundant atomic loads within helper methods

yara-blue:
Good to merge if you ask me! :tada:

examples/automatic_gain_control.rs
UnknownSuperficialNight: Sorry, didn't notice this

These might be irrelevant now after changes:

automatic_gain_control(0.90, 1.0, 0.0, 5.0)

personally, I've had no issues with target_level `1.0` but that could just be me

yara-blue: > These might be irrelevant now after changes:

how so? There now is: AutomaticGainControlSettings.target_level for this purpose right?

UnknownSuperficialNight: I mean that as in using target_level `0.9` vs `1.0` as default seems you use `0.9` while for me `1.0` works just fine, so I was commenting on either could be right or wrong depending on setup.

However, when I have a song that the artist decided to clip intentionally with AGC at `0.9` it lowers it below the original intended sound which for me, I want the original sound volume thus I go with `1.0`

Before there might have been slight issues with `1.0`, but it should be fixed due to `update_peak_level` always attacking as fast as possible at least that's my observation look at the video in [this comment](https://github.com/RustAudio/rodio/pull/808#issuecomment-3395290127) for my final results (i.e. before on certain songs depending on AGC settings it could have causes slight clipping at `1.0`)

yara-blue: I'll see if `1.0` works better now and if not I'll report it back here :+1:

PR #810, opened 2025-11-06
Fix assertion

YichiZhang0613:
In rodio/src/source/signal_generator.rs, frequency should be > 0 according to semantic but code allows it to be negative, so assertion should be changed into `assert!(frequency > 0.0, "frequency must be greater than zero");`
```rust
pub fn with_function(
    sample_rate: SampleRate,
    frequency: f32,
    generator_function: GeneratorFunction,
) -> Self {
    assert!(frequency != 0.0, "frequency must be greater than zero");
    let period = sample_rate as f32 / frequency;
```


yara-blue:
thanks! :+1:

PR #812, opened 2025-11-07
SourcesQueueOutput can peek metadata from next source

### LINKED ISSUE #811: First 512 samples of a source added to an empty Sink are always mono
To reproduce this issue, try following the music_wav example with a 2-channel WAV file with audio within the first ~11ms. Record the audio output and compare with the input. Instead of being the same, the first 11ms of the recorded output should be about double-length and mono (due to the stereo file being read as mono and converted to stereo.)

This seems to be related to the interaction between the mixer's UniformSourceIterator and SourcesQueueOutput, which plays "silence" at 1 channel when there are no sources. In particular, changing the "silence" played by SourcesQueueOutput from 1 channel to 2 channels (by editing my local copy of [queue.rs](https://github.com/RustAudio/rodio/blob/master/src/queue.rs#L224)) causes the sound to play normally.

I'm suspicious of the conditional logic on [current_span_len](https://github.com/RustAudio/rodio/blob/master/src/queue.rs#L132) -- I wonder if it's transitioning from silence to an actual sound improperly, and returning a span length of THRESHOLD on that borderline. But I can't investigate it right now, so I'll leave this issue here until I get the time to look at it further.

0----0:
Fixes #811 

SourcesQueueOutput currently violates the span interface by returning incorrect values for sample_rate and channels while the iterator is positioned on a source boundary (i.e. at the end of one source, before the next source has started.).

This PR adjusts its sample_rate and channels implementations so that they use current_span_len == 0 as an admissible heuristic to check if the current source is empty. This means that it can obey the span interface contract as long as the sources are returning Some current_span_len values. (If they aren't, there's probably no way we can possibly obey that contract.)

In order to actually uncomment the `basic` test for the queue, I also had to adjust the `SamplesBuffer` implementation to return a Some current_span_len value. This showed some unrelated rounding errors in `SkipDuration`, so I made unrelated adjustments to the `SkipDuration` math to reduce rounding errors, just to keep tests passing.

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -126,7 +126,7 @@ impl Iterator for SamplesBuffer {
 
     #[inline]
     fn size_hint(&self) -> (usize, Option<usize>) {
-        (self.data.len(), Some(self.data.len()))
+        (self.data.len() - self.pos, Some(self.data.len() - self.pos))
     }
 }
 


src/buffer.rs
```diff
@@ -73,7 +73,7 @@ impl SamplesBuffer {
 impl Source for SamplesBuffer {
     #[inline]
     fn current_span_len(&self) -> Option<usize> {
-        None
+        Some(self.data.len() - self.pos)
```
roderickvd, line 76: The contract is that `current_span_len` returns the length of the span / buffer regardless of its current position. That's for `size_hint` to do.

src/buffer.rs
```diff
@@ -126,7 +126,7 @@ impl Iterator for SamplesBuffer {
 
     #[inline]
     fn size_hint(&self) -> (usize, Option<usize>) {
-        (self.data.len(), Some(self.data.len()))
+        (self.data.len() - self.pos, Some(self.data.len() - self.pos))
```
roderickvd, line 129: `SamplesBuffer` could therefore also implement `ExactSizeIterator`. I know that's not directly related to this change, but see it also doesn't implement it currently.

src/queue.rs
```diff
@@ -154,12 +156,28 @@ impl Source for SourcesQueueOutput {
 
     #[inline]
     fn channels(&self) -> ChannelCount {
-        self.current.channels()
+        // current_span_len() should never return 0 unless the source is empty, so this is a cheeky hint
+        if self.current.current_span_len() != Some(0) {
```
roderickvd, line 160: I like the cheekiness :smile:

How much sense do you think it would make to add an `is_empty` helper method to `Source`?

roderickvd (changes requested):
Thanks a lot for tackling this one! Sorry for the late response. Here are a few points and questions for further improvement.

roderickvd:
Superseded by #817 

@0----0 thanks a lot for finding the root cause of this - let me know if you have any comments to that other PR.

PR #813, opened 2025-11-25
Remove stale comments

casey:
Remove possibly stale comments related to #510.

I'm not sure if these are stale or not, but I noticed that #510 was merged (actually the successor PR, #585) a while ago, so I figure they might be.

### DIFF
--- a/src/mixer.rs
+++ b/src/mixer.rs
@@ -110,35 +110,6 @@ impl Source for MixerSource {
         Err(SeekError::NotSupported {
             underlying_source: std::any::type_name::<Self>(),
         })
-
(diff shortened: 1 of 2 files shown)


yara-blue:
The idea behind these comments was to eventually implement error recovery when seeking. 

A seek can fail and if we seek in a mix of two songs we can end up with one having successfully seeked and the other failing and staying at its original position. The commented code would make it so we would undo the seek in the first song if the second failed.

Looking at it now that is really unnecessary and over-complicated as it would require wrapping everything in TrackPosition to make it seekable. So yes we can strip these out. 

A good followup would be implementing seek for these. That could simply forward the seek to each underlying source and eagerly return any error.

PR #815, opened 2025-12-14
Adds a builder for configuring outputstreams

yara-blue:
Its basically the same as the microphone builder. It will replace the OutputStream. In the future we'll add more abstractions on top. Until its done it lives under an experimental flag.

Names are subject to change too, Speakers is probably not ideal but it conveys the meaning better then OutputStream. I'm thinking of having a Source -> Stream -> Sink terminolgy where a Sink could be the audio card, the network or a file (the wavwriter).

### DIFF
--- a/src/microphone/builder.rs
+++ b/src/microphone/builder.rs
@@ -479,7 +479,6 @@ where
     /// let builder = MicrophoneBuilder::new()
     ///     .default_device()?
     ///     .default_config()?
-    ///     // Multiples of two work well for us
     ///     .prefer_buffer_sizes([
     ///         2048.try_into().expect("not zero"),
     ///         4096.try_into().expect("not_zero"),
@@ -493,8 +492,6 @@ where
     /// let builder = MicrophoneBuilder::new()
     ///     .default_device()?
     ///     .default_config()?
-    ///     // We need a minimum buffer of 4096
-    ///     // or we get glitches.
     ///     .prefer_buffer_sizes(4096..);
     /// # Ok::<(), Box<dyn std::error::Error>>(())
     /// ```
--- a/src/queue.rs
+++ b/src/queue.rs
@@ -61,9 +61,10 @@ impl SourcesQueueInput {
     where
         T: Source + Send + 'static,
     {
-        let mut next_sounds = self.next_sounds.lock().unwrap();
-        next_sounds.push((Box::new(source) as Box<_>, None));
-        next_sounds.len();
+        self.next_sounds
+            .lock()
+            .unwrap()
+            .push((Box::new(source) as Box<_>, None));
     }
 
     /// Adds a new source to the end of the queue.
--- a/src/speakers.rs
+++ b/src/speakers.rs
@@ -148,8 +148,7 @@ impl fmt::Debug for Output {
                 &self
                     .inner
                     .description()
-                    .map(|d| d.name().to_string())
-                    .unwrap_or("unknown".to_string()),
+                    .map_or("unknown".to_string(), |d| d.name().to_string()),
             )
             .finish()
     }
@@ -162,8 +161,7 @@ impl fmt::Display for Output {
             "{}",
             self.inner
                 .description()
-                .map(|d| d.name().to_string())
-                .unwrap_or("unknown".to_string()),
+                .map_or("unknown".to_string(), |d| d.name().to_string())
         )
     }
 }
--- a/src/speakers/builder.rs
+++ b/src/speakers/builder.rs
@@ -78,8 +78,7 @@ where
                 "device",
                 &self.device.as_ref().map(|d| {
                     d.0.description()
-                        .map(|d| d.name().to_string())
-                        .unwrap_or("unknown".to_string())
+                        .map_or("unknown".to_string(), |d| d.name().to_string())
                 }),
             )
             .field("config", &self.config)
@@ -143,8 +142,7 @@ where
                 source,
                 device_name: device
                     .description()
-                    .map(|d| d.name().to_string())
-                    .unwrap_or("unknown".to_string()),
+                    .map_or("unknown".to_string(), |d| d.name().to_string()),
             })?
             .collect();
         Ok(SpeakersBuilder {
@@ -174,8 +172,7 @@ where
                 source,
                 device_name: default_device
                     .description()
-                    .map(|d| d.name().to_string())
-                    .unwrap_or("unknown".to_string()),
+                    .map_or("unknown".to_string(), |d| d.name().to_string()),
             })?
             .collect();
         Ok(SpeakersBuilder {
@@ -210,8 +207,7 @@ where
                 source,
                 device_name: device
                     .description()
-                    .map(|d| d.name().to_string())
-                    .unwrap_or("unknown".to_string()),
+                    .map_or("unknown".to_string(), |d| d.name().to_string()),
             })?
             .into();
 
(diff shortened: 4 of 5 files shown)


src/queue.rs
```diff
@@ -56,14 +56,15 @@ pub struct SourcesQueueInput {
 impl SourcesQueueInput {
     /// Adds a new source to the end of the queue.
     #[inline]
-    pub fn append<T>(&self, source: T)
+    pub fn append<T>(&self, source: T) -> usize
```
roderickvd: What's the rationale for changing this to return a `usize`? By analogy, [`Vec::append`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.append) doesn't return a value either.

src/speakers/builder.rs
```diff
@@ -0,0 +1,553 @@
...
+    speakers::{self, config::OutputConfig},
+    ChannelCount, OutputStream, SampleRate,
+};
+
+/// Error configuring or opening speakers input
+#[allow(missing_docs)]
+#[derive(Debug, thiserror::Error, Clone)]
+pub enum Error {
+    /// No output device is available on the system.
+    #[error("There is no input device")]
```
roderickvd: `s/input/output`

src/speakers.rs
```diff
@@ -0,0 +1,179 @@
...
+//! ```
+//!
+//! # Device Selection
+//!
+//! ```no_run
+//! use rodio::speakers::{SpeakersBuilder, available_outputs};
+//!
+//! # fn main() -> Result<(), Box<dyn std::error::Error>> {
+//! // List all available input devices
+//! let inputs = available_outputs()?;
```
roderickvd: `s/inputs/outputs/g`

src/speakers/builder.rs
```diff
@@ -0,0 +1,553 @@
...
+    pub fn device(
+        &self,
+        device: impl Into<cpal::Device>,
+    ) -> Result<SpeakersBuilder<DeviceIsSet, ConfigNotSet, E>, Error> {
+        let device = device.into();
+        let supported_configs = device
+            .supported_output_configs()
+            .map_err(|source| Error::OutputConfigs {
+                source,
+                device_name: device.name().unwrap_or_else(|_| "unknown".to_string()),
```
roderickvd: `Device::name` has been deprecated in favor of `Device::description().name()` in cpal v0.17. You could also choose to pretty-print `Device::description`.

src/speakers/config.rs
```diff
@@ -0,0 +1,83 @@
...
+            (cpal::BufferSize::Default, _) | (_, cpal::SupportedBufferSize::Unknown) => true,
+            (
+                cpal::BufferSize::Fixed(n_frames),
+                cpal::SupportedBufferSize::Range {
+                    min: min_samples,
+                    max: max_samples,
+                },
+            ) => {
+                let n_samples = n_frames * self.channel_count.get() as u32;
+                (*min_samples..*max_samples).contains(&n_samples)
```
roderickvd: For cpal v0.18 I have the idea to have something like a `cpal::BufferSize::Range { min, max }` that you can request. Open to API ideas: https://github.com/RustAudio/cpal/issues/956

roderickvd:
As for naming,  here are some suggestions:

* OutputBuilder, InputBuilder, Output, Input
* PlaybackBuilder, CaptureBuilder, Playback, Capture

If we want to use Source → Stream → Sink terminology (pretty accurate actually) then to disambiguate we should rename the existing Sink to something like MixerQueue, AudioQueue, or QueuedOutput.

src/queue.rs
yara-blue: Oh haha, I needed this for some tests in zed. I'll throw it out. Good catch!

src/speakers.rs
yara-blue: thank you :)

src/speakers/config.rs
yara-blue: I went with with offering a range here (see the prefer_... methods). That might work better for buffersize too

yara-blue:
> If we want to use Source → Stream → Sink terminology (pretty accurate actually) then to disambiguate we should rename the existing Sink to something like MixerQueue, AudioQueue, or QueuedOutput.

Agreed, we should. Sink would be the perfect name for this "concept". We can have a file sink as well (the current `to_wav` thing). We can do that after merging this though. 

Its already pretty big. Makes sense to land it as is and then tweak as I use it. The next thing I'd like to add after this is a new `Source` trait called `FixedSource`.

Hit that merge button if you think we can get it in.

src/queue.rs
```diff
@@ -61,10 +61,9 @@ impl SourcesQueueInput {
     where
         T: Source + Send + 'static,
     {
-        self.next_sounds
-            .lock()
-            .unwrap()
-            .push((Box::new(source) as Box<_>, None));
+        let mut next_sounds = self.next_sounds.lock().unwrap();
+        next_sounds.push((Box::new(source) as Box<_>, None));
+        next_sounds.len();
```
roderickvd: This line should be removed, now that `append` doesn't return anything anymore.

src/speakers/config.rs
```diff
@@ -0,0 +1,83 @@
+use std::num::NonZero;
+
+use crate::{math::nz, stream::OutputStreamConfig, ChannelCount, SampleRate};
+
+/// Describes the input stream's configuration
```
roderickvd: s/input/output

src/speakers/config.rs
roderickvd: The range should probably be inclusive like `min..=max`.

src/speakers.rs
```diff
@@ -0,0 +1,190 @@
...
+impl fmt::Debug for Output {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("Device")
+            .field(
+                "inner",
+                &self
+                    .inner
+                    .description()
+                    .map(|d| d.name().to_string())
+                    .unwrap_or("unknown".to_string()),
```
roderickvd: Can be simplified to `map_or` (I thought there was a clippy lint for that).

src/speakers.rs
```diff
@@ -0,0 +1,190 @@
...
+
+impl fmt::Display for Output {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
+            "{}",
+            self.inner
+                .description()
+                .map(|d| d.name().to_string())
+                .unwrap_or("unknown".to_string()),
```
roderickvd: Same.

src/microphone/builder.rs
```diff
@@ -489,8 +493,8 @@ where
     /// let builder = MicrophoneBuilder::new()
     ///     .default_device()?
     ///     .default_config()?
-    ///     // We want mono, if thats not possible give
-    ///     // us the lowest channel count
+    ///     // We need a minimum buffer of 4096
```
roderickvd: It's strange that it does, and I am doubtful that it would be universally applicable across sample rates, systems and input devices _unless_:
- we got something wrong in our ring buffer implementation
- it was bugged due to the `UniformSourceIterator` transitioning out of silence - the latest commits fixed that, does it help this use case too?

yara-blue: oh it absolutely does not!

I tried to add a comment describing a fictional situation where it could be useful to have a minimum buffer size. I'll change this!

yara-blue:
Thanks for the review, should be good to merge now :)

roderickvd (approved):
LGTM

PR #816, opened 2025-12-19
Better error on `OutputStreamBuilder::open_stream`

angeloanan:
I'd think the panic will make much more sense if there's the word `No` in front of it :D


yara-blue:
lmao, thanks for that!

PR #819, opened 2025-12-23
fix: remove rust-toolchain.toml to avoid forcing toolchain on all developers

roderickvd:
PR #814 added `rust-toolchain.toml` for Nix flake support, but this file affects all developers using rustup, not just Nix users. The pinned version (1.91) is outdated and conflicts with CI which tests against stable/beta/nightly.

This change:
- Removes `rust-toolchain.toml`
- Updates `flake.nix` to specify stable Rust directly using `rust-overlay`
- Keeps Nix-specific config in Nix files without affecting other devs

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -8,7 +8,7 @@
     inputs: inputs.utils.lib.eachDefaultSystem (
       system: let
         pkgs = inputs.nixpkgs.legacyPackages.${system}.extend inputs.rust-overlay.overlays.default;
-        rust = pkgs.rust-bin.stable.latest.default.override {
+        rust = pkgs.rust-bin.stable."1.87.0".default.override {
           extensions = [ "rust-src" "rust-analyzer" ];
         };
       in {
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,12 +14,11 @@ jobs:
   tests:
     name: Tests
     runs-on: ${{ matrix.os }}
-    continue-on-error: ${{ matrix.toolchain == 'nightly' }}
     strategy:
       fail-fast: false
       matrix:
         os: [macos-latest, windows-latest, ubuntu-latest]
-        toolchain: [stable, beta, nightly]
+        toolchain: ["1.87"]
         include:
           - os: macos-latest
             MACOS: true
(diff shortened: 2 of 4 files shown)


yara-blue:
> The pinned version (1.91) is outdated and conflicts with CI which tests against stable/beta/nightly.

Should we not test CI against a single known version? Right now every PR tends to have to fix the latest new clippy issues. On the other hand it does spread the load of dealing with new clippy lints :laughing:

Let me know what you think (I've no serious objection against CI running against the world)

flake.nix
```diff
@@ -8,7 +8,9 @@
     inputs: inputs.utils.lib.eachDefaultSystem (
       system: let
         pkgs = inputs.nixpkgs.legacyPackages.${system}.extend inputs.rust-overlay.overlays.default;
-        rust = pkgs.rust-bin.fromRustupToolchainFile ./rust-toolchain.toml;
+        rust = pkgs.rust-bin.stable.latest.default.override {
```
yara-blue: should be a specific version and not just latest stable, otherwise the project is not reproducible.

yara-blue: let me know if you dont use nix yourselves though

roderickvd: To tell you the truth I don't even know what it is, I just threw this together.

roderickvd:
Sure, then we'd have to find the MSRV, set it in Cargo.toml and set the CI accordingly. cpal just went to 1.78.

I don't mind running floppy on stable at the same time: I kinda like continuously fixing the new and improved lints.

roderickvd:
@yara-blue what do you want to do here? I don't use nix and don't really like the toolchain being forced this way.

yara-blue:
Can we keep it and set it to the MSRV? There should be a tool to find the minimum in an automated fashion.

Offtopic; I'm sick and might be for a little while so my responses will be somewhat erratic sorry.

roderickvd:
> Can we keep it and set it to the MSRV? There should be a tool to find the minimum in an automated fashion.

I am for setting an MSRV, I can work that out. 

I don't like forcing the toolchain to the MSRV though. My workflow is developing with Rust stable - better lints and compiler messages - and then ensuring it works on MSRV too. I think that's pretty typical. 


> Offtopic; I'm sick and might be for a little while so my responses will be somewhat erratic sorry.

Health first. Beterschap!

roderickvd:
I determined the MSRV to be 1.87 and updated the CI and Nix accordingly.

Opinionated change: I also removed testing on Rust beta and nightly. Testing the Rust compiler isn't in our perimeter. Though it doesn't happen often, but I do remember librespot failing its CI due to compiler bugs out of its control.

yara-blue:
Might be opinionated but I share your opinion :)

PR #821, opened 2025-12-28
feat: add optional 64-bit sample precision

roderickvd:
This started as a code quality journey when I noticed that in some places we still had `Item = f32` hardcoded where we should have been using the `Sample` type alias. While fixing those inconsistencies, I realized we could make sample precision configurable at compile time.

I know we've discussed that 32-bit floating point should be sufficient for all audio applications, and I agree that's true for the vast majority of use cases. However, similar to how [CamillaDSP does it](https://github.com/HEnquist/camilladsp/blob/f01fddcae9cf8675751b3b2a9d715e4433948169/src/lib.rs#L85), rodio can now be built with either `f32` (default) or `f64` samples using the `64bit` feature flag. The 64-bit mode helps when precision drift accumulates, like when chaining many audio operations together.

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -10,6 +10,21 @@ pub type ChannelCount = NonZero<u16>;
 /// Number of bits per sample. Can never be zero.
 pub type BitDepth = NonZero<u32>;
 
+// NOTE on numeric precision:
+//
+// While `f32` is transparent for typical playback use cases, it does not guarantee preservation of
+// full 24-bit source fidelity across arbitrary processing chains. Each floating-point operation
+// rounds its result to `f32` precision (~24-bit significand). In DSP pipelines (filters, mixing,
+// modulation), many operations are applied per sample and over time, so rounding noise accumulates
+// and long-running state (e.g. oscillator phase) can drift.
+//
+// For use cases where numerical accuracy must be preserved through extended processing (recording,
+// editing, analysis, long-running generators, or complex DSP graphs), enabling 64-bit processing
+// reduces accumulated rounding error and drift.
+//
+// This mirrors common practice in professional audio software and DSP libraries, which often use
+// 64-bit internal processing even when the final output is 16- or 24-bit.
+
 /// Floating point type used for internal calculations. Can be configured to be
 /// either `f32` (default) or `f64` using the `64bit` feature flag.
 #[cfg(not(feature = "64bit"))]


yara-blue:
Nothing against spreading the use of the Sample alias further, however changing so many public APIs from `f32` to `Float` makes me hesitate. 

> I know we've discussed that 32-bit floating point should be sufficient for all audio applications, and I agree that's true for the vast majority of use cases.
...
> The 64-bit mode helps when precision drift accumulates, like when chaining many audio operations together.

I'm having trouble imagining how to lose 8 bits of precision (24bit mantissa need 16bit for transparent audio). Could you give some examples (we can also reuse those in the docs). 

I could see how naively mixing 512 songs each at higher then 50% gain could overflow but is that now a fault in the mixer algorithm then?

roderickvd:
> Nothing against spreading the use of the Sample alias further, however changing so many public APIs from `f32` to `Float` makes me hesitate.

I can _kind of_ see how in public documentation it could throw users off for a second, only to see that it's just a type alias to `f32` or `f64` in the end. Stuff like `source.reverb(Duration::from_millis(40), 0.7)` or `source.distortion(4.0, 0.3)` continues to compile with no change. So will `source.distortion(foo, bar)` with `foo` and `bar` in `f32` and Rodio default features.

Advanced users can type their `foo` and `bar` as `rodio::Float` or do `source.distortion(foo as _, bar as _)` if they want to be compatible with both 32 and 64-bit Rodio. But they don't need to.

> I'm having trouble imagining how to lose 8 bits of precision (24bit mantissa need 16bit for transparent audio). Could you give some examples (we can also reuse those in the docs).

The anchor shouldn't be "is 16-bit transparent?" (it is), but rather "does our processing preserve the input fidelity?" If someone feeds in 24-bit audio, they shouldn't get lower quality back out, particularly when recording, editing, or mixing.

Each floating-point operation rounds to `f32`'s 24-bit precision, accumulating error as `sqrt(N) x epsilon` where `N` is the number of operations, with a number of bits lost equal to `log2(N) / 2`. So it shows that fidelity is reduced below 24-bit even after a single `f32` operation, and reduced with a full bit with four operations. That's after one `source.low_pass(x)` for example, which does 5 multiply-adds per sample.

Another case is long-running operations that accumulate error over time. For example, `SignalGenerator` phase drifts 180 degrees within one hour in `f32`. That doesn't matter for a single tone, but when mixing multiple generators or doing modulation, they drift out of sync.

Finally, by analogy, CamillaDSP (a reference Rust DSP library) actually defaults to `f64` and professional DAWs use 64-bit engines. This is why I propose to default Rodio to `f32` with zero impact on typical use cases, and advanced users can flip on `--features 64bit` to preserve full fidelity.

yara-blue:
> Each floating-point operation rounds to `f32`'s 24-bit precision, accumulating error as `sqrt(N) x epsilon` where `N` is the number of operations, with a number of bits lost equal to `log2(N) / 2`. So it shows that fidelity is reduced below 24-bit even after a single `f32` operation, and reduced with a full bit with four operations. That's after one `source.low_pass(x)` for example, which does 5 multiply-adds per sample.

Thank you for that proof, its better then anything I could find with a quick google last night. The audio world is so full of bigger number thus better that it can be hard to decern whats makes sense.

This is clearly a change worth having. I also would not mind having this example in the code. Both so any future maintainer does not rip it out and such that users can understand when its needed. If you agree could you please add that? I'm also good with merging as is though!

roderickvd:
> Thank you for that proof, its better then anything I could find with a quick google last night. The audio world is so full of bigger number thus better that it can be hard to decern whats makes sense.

_Proof_ is a strong word but yes: reasonable heuristics.

> This is clearly a change worth having. I also would not mind having this example in the code. Both so any future maintainer does not rip it out and such that users can understand when it's needed. If you agree could you please add that? I'm also good with merging as is though!

👍 I came up with a note in https://github.com/RustAudio/rodio/pull/821/commits/ed4af6dbc5cd99239166161ac027418ac0fde33e. Feel free to edit or signal if you'd like changes or improvements.

Merging now.

PR #823, opened 2026-01-03
Rename Sink -> Player and OutputStream -> OsSink

yara-blue:
As discussed in https://github.com/RustAudio/rodio/pull/815#issuecomment-3685623618 this renames parts of rodio so they actually make sense.

This builds upon: #815, it should be merged first

PS: ignore branch name, I wanted to introduce a new source trait here but the rename takes priority.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -179,7 +179,7 @@ pub use cpal::{
 mod common;
 mod player;
 mod spatial_player;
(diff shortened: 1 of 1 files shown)


yara-blue:
This diff seems huge, however it still includes the un-merged #815 changes. Once those land it should be more reasonable.

yara-blue:
closed (renamed branch)

PR #825, opened 2026-01-03
perf: improve mixer and sample rate conversion

roderickvd:
This PR speeds up `cargo bench` by approximately 7.5% on my MacBook Air M3.

I had expected that `--features crossbeam-channel` would have been even faster because its operations are lock-free, where `std::sync::mpsc` are not. However, the speedup is reduced slightly to 7.4%, probably because there isn't any contention to speak of.

Improvements:

* Replace `Vec` with `VecDeque` for queue and output buffering and avoid O(n) `remove(0)` operations
* Replace `Mutex` + `Atomic` pending list with a channel and consume via `try_recv`
* `retain_mut` sources in-place to reduce allocations and copies


roderickvd:
1.7% faster on my Raspberry Pi 4 where other math operations dominate.

yara-blue (approved):
Looks good 👍 

LGTM

Issue #712, opened 2025-03-03
Span-less rodio?


dvdsk:
I have had some time and been pouring it into _rodio_ lately, having written
about 3x span/parameter_change related code I have some opinions now :)
I have paused progress on the `parameters_changed()` PR while working out if an
alternative (see below) works better.

A lot of performance is being lost by dealing with spans in various sources.
Lets take `buffered` as an example. Since we got to notify any consumer at the
right moment of a sample rate/channel count changes we up having to keep track
of a lot.

This is (just a part of) the `Buffered::next()` method I have been working on this morning:
```rust
if self.shared.samples_in_memory.len() < self.samples_index {
	let sample = self.shared.samples_in_memory[self.samples_index];

	if self.samples_index == self.next_parameter_change_at {
		let new_params = &self.shared.parameters[self.parameter_changes_index];
		self.sample_rate = new_params.sample_rate;
		self.channel_count = new_params.channel_count;
	}

	// sample after sample where flag a parameter_change
	if self.samples_index > self.next_parameter_change_at {
		self.next_parameter_change_at =
			if self.parameter_changes_index > self.shared.parameters.len() {
				usize::MAX
			} else {
				self.shared.parameters[self.parameter_changes_index].index
			};
	}

	self.samples_index += 1;
	return Some(sample);
}
```

Now if instead we got entirely rid of them *1 this (part of the) method would become trivial:
```rust
if self.shared.samples_in_memory.len() < self.samples_index {
	let sample = self.shared.samples_in_memory[self.samples_index];
	self.samples_index += 1;
	return Some(sample);
} 
```

## Alternative: no spans
Get rid of spans, source will no longer have member functions `channels()` &
`sample_rate()` instead it gets `set_channel_count()` & `set_sample_rate()`. The
consumer (outputstream) uses those to communicate the picked sample_rate to the
edges of the audio `tree`.

### Schematic example
```txt
                                     OutputStream @ sr1,chs 4
									        |
										  mixer
										/       \
									 low_pass    mixer
									  /           /  \_______
									mixer       queue        \
								    /   \         |       convertor
							    Square  Noise  convertor      |
							    @44.1/4  @44.1/4  |        decoder
								               decorder      @48.0/1
											    @ 96.1/2
```
Here `OutputStream` picks a sample rate and channel count, then forwards it too
the edges. Generators like Square & Noise use that sample rate while between
every variable sample rate edge and the tree a resampler is inserted. As an
[...]

PR #706, opened 2025-02-24
Parameters changed

dvdsk:
implements #694

currently draft, due to high number of non trivial changes this PR also aims to dramatically extend the test suite of rodio. Hopefully that will catch most bugs.

@roderickvd I could use some help fixing the decoders, they now just have `todo!()` as `parameters_changed` implementation. Let me know if you want to work on that. The branch is under the rodio repo so you can just pull it down and push your work to it.

status (ill edit this along the way)
- [x] trivial sources implemented (empty, sine etc)
- [x] normal sources implemented (skip etc)
- [ ] extra tests for normal sources
- [ ] complex sources implemented (queue and friends) 
- [ ] extra tests for complex sources
- [ ] decoder sources implemented
- [ ] extra tests for decoder sources (unsure if needed?)

Issue #694, opened 2025-01-26
span_length is the wrong abstraction


dvdsk:
When decoding audio you get chunks with constant sample rate and channel count. It seemed natural to extend that into rodio's audio pipeline. The consumer of audio gets the guarantee that the sample rate and channel count will not change for the next n-samples. The consumer can then deal with that like this:

```rust
let mut resampler = Resampler::new();
loop {
  resampler.set_sample_rate(source.sample_rate());
  for _ in 0..source.current_span_len() {
    resample.resample(source.next()?);
  }
}
```

This seems very efficient right? It is however no different then:
```rust
let mut resampler = Resampler::new();
loop {
  resampler.set_sample_rate(source.sample_rate());
  while !source.parameters_changed() {
    resample.resample(source.next()?);
  }
}
```

Especially if we let the compiler know that the parameters usually do not change using:
```rust
impl Source {
  #[cold]
  fn parameters_changed() -> bool;

  ..
}
```

Currently at least skipping, seeking and queue are non-trivial in rodio since we *may not* break our promised span_len. That span len is however never used except to *create* the boolean `parameters_changed` via a comparison to an increasing index. Replacing `current_span_len` by `parameters_changed` makes rodio code far simpeler. For example seek:

```rust
impl Source {
  fn seek(&mut self, some_params: some_type) {
    self.parameters_changed = true
    // and some code to guarantee frame alignment
   //  but nothing for span alignment since spans do not exist anymore!
    ..
  }

  fn parameters_changed() -> bool {
    self.parameters_changed
  }
}
```

All the complex code in queue can just be removed, it can simply forward parameters_changed until a source ends: then it just switches it to true. Right now it is has to play silence to fill the current span if a source ends early. So this change will not only make the code far far simpler it will improve audio latency!

Regarding breaking the Source API, an argument along the line of https://github.com/RustAudio/rodio/issues/690#issuecomment-2612265055 can be made. This would also replace #690.

I think this is a wonderful fix/improvement. I might be wrong. Looking forward to your input.

PetrGlad:
I have to read it again with the fresh mind but even with this version the source should guarantee that whole frame (samples for all channels) is received, otherwise it invites some complicated logic again.
Instead of boolean I'd consider a revision counter that is increased every time parameters change, this is usually my favorite in such situations. Or if it is a boolean it should be be reset every time target sink sees it's value.
Even if spans are not visible, I guess the our decoder wrappers will still need to handle a lot of cases?

[...]

Issue #691, opened 2025-01-23
Seek breaks span


yara-blue:
Seek might need padding silence when resample rate or channel count change. Since the consumer has no way to know they have.

Alternative, make seek an exception and make the consumer recall sample_rate, channel count & span len.

PetrGlad:
We may require that all variable sources are wrapped into something like UniformSourceIterator, so in most cases spans can only change channels, not sample rate...

dvdsk:
would not help since seek is implemented for decoders, we can not resample to such a uniform source if the spans are not correct.

PetrGlad:
At least it would help to not care about the sample rate changes downstream, in sources that are inserted after decoders.

With decoders themselves, I am thinking of more comprehensive test harness that would generate both various source audio files and random seeks, something like property testing but for our case.

dvdsk:
> At least it would help to not care about the sample rate changes downstream

it would, but it would come at a performance cost. Hifi-resampling is performance intensive, resampling everything at the edges might be wasteful. Maybe we can make the resampler very lazy, that would help....

Honestly I think we would need a prototype before making any decision. A rodio fork that only supports a few sources (lets say queue & mix) and works by informed resampling at the input. With informed I mean that idea I had earlier with the output communicating its preferred sample rate down and everything adjusting along the way. 

After I finish fixing queue I can work on the informed sample rate bit, once that's done you could make the prototype. Then we have to figure out some workloads to compare the implementations with. If the performance difference is positive or trivial the code simplifications might be worth it.

PR #828, opened 2026-01-10
Introduce FixedSource as experimental

yara-blue:
This introduces a small component from [rodio-experiments](https://github.com/RustAudio/rodio-experiments) into rodio under the experimental flag. 

For full context see that repo, though in short this enables:
- consumers like other sources and sink to express they need sources that do not change their parameters
- us to write more optimized versions of queue, channelcount convertor and resample

### DIFF
--- a/src/fixed_source.rs
+++ b/src/fixed_source.rs
@@ -1,14 +1,19 @@
-//! TODO
+//! Sources of sound and various filters which never change sample rate or
+//! channel count.
 use std::time::Duration;
 
 use crate::{ChannelCount, Sample, SampleRate};
 
-/// TODO
+/// Similar to `Source`, something that can produce interleaved samples for a
+/// fixed amount of channels at a fixed sample rate. Those parameters never
+/// change.
 pub trait FixedSource: Iterator<Item = Sample> {
     /// May NEVER return something else once its returned a value
     fn channels(&self) -> ChannelCount;
     /// May NEVER return something else once its returned a value
     fn sample_rate(&self) -> SampleRate;
-    /// TODO
+    /// Returns the total duration of this source, if known.
+    ///
+    /// `None` indicates at the same time "infinite" or "unknown".
     fn total_duration(&self) -> Option<Duration>;
 }
--- a/src/speakers.rs
+++ b/src/speakers.rs
@@ -125,7 +125,7 @@ pub struct Output {
 }
 
 impl Output {
-    /// TODO doc comment also mirror to microphone api
+    /// Whether this output is the default sound output for the OS
     pub fn is_default(&self) -> bool {
         self.default
     }
--- a/src/speakers/builder.rs
+++ b/src/speakers/builder.rs
@@ -551,12 +551,13 @@ where
         crate::stream::MixerDeviceSink::open(&device, &config.into_cpal_config(), error_callback)
     }
 
-    /// TODO
-    pub fn open_queue() -> Result<QueueSink, DeviceSinkError> {
-        todo!()
-    }
+    // TODO
+    // pub fn open_queue() -> Result<QueueSink, DeviceSinkError> {
+    //     todo!()
+    // }
 
-    /// TODO
+    /// Open the device with the current configuration and play a single
+    /// `FixedSource` on it.
     pub fn play(
         self,
         mut source: impl FixedSource + Send + 'static,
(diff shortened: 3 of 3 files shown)


roderickvd:
Sorry for my buddy Claude posting on my behalf earlier. Looks good to both him and me :smile:

How much effort do we want to put into salvaging the current situation with `current_span_len` (#712)? To recount a few pain points:

- Currently the contract isn't implemented nor even documented consistently. In my view it _should_ represent the total length of the current span - pretty much what it says on the tin (#694). The _remaining_ length would be for `size_hint`. What filters actually do is all over the place.

- `size_hint`s contract is that it's a hint, not a guarantee. That's actually true for decoders, because you cannot ensure that you can successfully decode or even read all bytes before you have. Then for signal generators that don't suffer from that, we should implement `ExactSizeIterator`. Also not done consistently.

- Just to spell it out, `current_span_len` returning remaining length would therefore also be a hint and no guarantee.

- Because we lack something like a `parameters_changed` function or signal (#706), each filter that needs to deal with that pretty much needs to copy the same code. That's poorly maintainable code duplication as well as an ugly runtime performance hit.

- `try_seek` and `skip_duration` are more interesting cases (#688, #691), because after seeking typically we will be somewhere mid-span, and we can only estimate from `size_hint` when we hit the next boundary to check for a parameter update. Depending on the fidelity of that size hint, that may not be accurate and lead to glitches. I think it can be quite OK, we just need to fix `size_hint` here and there.

I've got a couple things brewing to basically ensure consistent usage of `current_span_len`, `size_hint`, and `ExactSizeIterator`. I don't mind putting some time in it and put a PR in, but want to align with you first. Assuming it'll be some time before your experiments get in as mainstream, it'd be nice to have `current_span_len` and friends working better than they do today.

yara-blue:
> How much effort do we want to put into salvaging the current situation with `current_span_len` (#712)?

I personally zero :) I hate spans now :stuck_out_tongue_closed_eyes:. Once the experimental stuff gets stable I would like to recommend to users to convert to FixedSource ASAP.

We should at some point benchmark: `DynamicSource (the current one) -> Queue -> Resampler (for output)` vs `DynamicSource -> Resampler -> FixedSource -> Queue -> Resampler`. There are quite some scenarios to try (like: constant input sample rate matching/not matching output & changing input sample rate), however I would not be surprised if double resampling to get to FixedSource ends up cheaper as its Queue implementation can be trivial.

> Just to spell it out, current_span_len returning remaining length would therefore also be a hint and no guarantee.

That would break resampling I think? Or do you suggest a signal to watch that changes? The latter would wreck performance (which is why I gave up on it).

> In my view it should represent the total length of the current span - pretty much what it says on the tin

Good point the current name & contract contradict each-other. Though we could also rename the method to resolve this. In my perfect world the DynamicSource is just there for backwards compatibility, decoders would get a re-sampler build. With a re-sampler build in they can optimally use all the info they have about parameter stability.

yara-blue:
> Sorry for my buddy Claude posting on my behalf earlier. Looks good to both him and me 😄

We can keep the discussion going here :+1: after I merge this.

roderickvd:
> > How much effort do we want to put into salvaging the current situation with `current_span_len` (#712)?
> 
> I personally zero :) I hate spans now 😝. Once the experimental stuff gets stable I would like to recommend to users to convert to FixedSource ASAP.

If you see no reason against it, then I'll put in some effort to fix `current_span_len` up for now.

> We should at some point benchmark: `DynamicSource (the current one) -> Queue -> Resampler (for output)` vs `DynamicSource -> Resampler -> FixedSource -> Queue -> Resampler`. There are quite some scenarios to try (like: constant input sample rate matching/not matching output & changing input sample rate), however I would not be surprised if double resampling to get to FixedSource ends up cheaper as its Queue implementation can be trivial.

The resampler at the end of the pipeline would basically be a pass-through, when `DynamicSource` has already been resampled appropriately. I imagine it's only when users mix multiple sources of different sampling rates, that there could be double resampling. Say source A in 96 kHz, source B in 44.1 kHz and an output to 48 kHz.

> > Just to spell it out, current_span_len returning remaining length would therefore also be a hint and no guarantee.
> 
> That would break resampling I think?

No, the current resampler just takes frames (not spans) without even calling `current_span_len`, ever.

For the future resampler with Rubato, I think we should use a fixed-out regime. That means that:
- anything comes after always receives a fixed number of frames
- we don't care about anything that comes before, and don't need to zero-pad or anything

I'm also toying around with `dasp_interpolate`. It's easy in that it also just works with frames, iterator-style, but Rubato's performance is just much better both in terms of CPU usage and anti-aliasing attenuation.

> Or do you suggest a signal to watch that changes? The latter would wreck performance (which is why I gave up on it).

I imagine it'd be possible to signal over a broadcast channel. But with your architecture, we don't need to.

PR #829, opened 2026-01-14
fix: span inconsistencies

roderickvd:
Continuing from discussion at #828.

Changes:
- `current_span_len` consistently returns _total_ samples in the current span
- `size_hint` returns the _remaining_ samples, if known
- `ExactSizeIterator` is implemented where the inner source has a fixed length
- `try_seek` takes into account mid-span seeks where necessary
- some opportunistic bug fixes

I did not fix up the decoders, as I had already done so under the infamous #786.


roderickvd:
GitHub commit log is broken after squashing: opened a new PR.

PR #830, opened 2026-01-14
SamplesBuffer::new() example in CHANGELOG

ghenry:
Updated usage of SamplesBuffer to show the new nz! macro which caught me out as per our chat in https://github.com/RustAudio/cpal/pull/1021#issuecomment-3751972803


yara-blue:
Thanks! These kinds of PR's are super helpful. Its hard as maintainer to know what will catch users off guard and what wont.

PR #831, opened 2026-01-14
fix: correct Source trait semantics and span tracking bugs

### LINKED ISSUE #691: Seek breaks span

roderickvd:
Clarify Source::current_span_len() returns total span length (not remaining), while size_hint() returns remaining samples of the entire iterator. Fix multiple bugs in span boundary detection, seeking, and iterator implementations.

Opportunistic fixes:
- fix division by zero, off-by-one error, and zero case handling
- prevent counter overflows
- optimize vector allocations

Mid-span seeking: best-effort tracking after try_seek() to detect span boundaries when seeking lands mid-span.

Fixes #691

### DIFF
--- a/src/source/zero.rs
+++ b/src/source/zero.rs
@@ -6,6 +6,8 @@ use super::SeekError;
 use crate::common::{ChannelCount, SampleRate};
 use crate::{Sample, Source};
 
+/// Error returned by [`Zero::new_samples`] when `num_samples` is not a
+/// multiple of the channel count.
 #[derive(Debug, thiserror::Error, Clone, PartialEq, Eq)]
 #[error("num_samples must be a multiple of the channel count")]
 pub struct ZeroError;


roderickvd:
I did not fix the decoders because I had already done so with the infamous #786.

src/queue.rs
```diff
@@ -135,37 +135,16 @@ fn threshold(channels: ChannelCount) -> usize {
...
-        let (lower_bound, _) = self.current.size_hint();
-        // The iterator default implementation just returns 0.
-        // That's a problematic value, so skip it.
-        if lower_bound > 0 {
-            return Some(lower_bound);
-        }
-
-        // Otherwise we use a frame-aligned threshold value.
-        Some(threshold(self.current.channels()))
+        None
```
roderickvd: I'm reworking this. A queue should never return `None` - not when it is drained without keep alive (that's `Some(0)`) and certainly not when a queued source reports `None`.

roderickvd: Done in [7538971](https://github.com/RustAudio/rodio/pull/831/commits/75389712142bd88c7812634e326e6241934797cf). Not trivial.

src/source/position.rs
```diff
@@ -88,28 +94,34 @@ where
...
-                self.set_current_span();
-            };
-        };
-        item
+        let item = self.input.next()?;
+        self.samples_counted += 1;
+
+        // Detect span boundaries by TWO mechanisms:
+        // 1. Reached end of span (by length) - handles same-parameter consecutive spans
+        // 2. Parameters changed - handles mid-stream parameter changes
```
yara-blue: 2 breaks the contract right? You are only allowed to change parameters at the end of a span. Or is 2 here in case of seeking? Should be removed from `Source` and allowed only on `FixedSource`, its seems really complex and costly to get right.

roderickvd: You are right. I can't remember why I did this. I will remove it unless I start to remember.

Have you had the chance to review the rest of the code? No problem if you haven't, but if you have, then I'll take this as hint for a final commit and get it merged.

yara-blue: will see if I can review it this weekend, gotta pace myself unfortunately

roderickvd: I remembered: it was for seeking, indeed. I think I found a way to make it work at acceptable cost.

roderickvd:
Verifying span boundary behavior in the other sources, latest commits I also added:
1. stereo and multi-channel audio support to `BltFilter`
2. mid-frame stream ending in `ChannelVolume` (akin to what we had fixed in `queue`)

w.r.t. the first, I wonder how many people use that, given it only ever worked correctly on mono sources.

roderickvd:
I want to take a final look at a couple of other sources, that may need to deal with mid-stream endings as well. In the meantime, feel free to review what we've got already, as I might apply the same patterns that we've already got.

roderickvd:
This thing is getting big but I think I've got it down now. Took me a good week's worth of evenings!

I've left the intermediate commits intact so you can see my journey. I've been going back-and-forth to defend against the fact that when source iterators end mid-frame, multi-channel audio and queuing blows up.

First I added defensive measures to "filter" sources that depend on proper frame alignment. Then, I removed it from the filters in favor of placing it with the "producer" sources: the decoders, mostly, and documenting the frame-alignment requirement.

I think it's better this way.

src/decoder/flac.rs
```diff
@@ -153,7 +165,16 @@ where
                     self.current_block_channel_len = (block.len() / block.channels()) as usize;
                     self.current_block = block.into_buffer();
                 }
-                _ => return None,
+                _ => {
+                    // Input exhausted - check if mid-frame
```
yara-blue, line 169: does this actually happen at the decoder level? Cause that feels to me like a decoder/file corruption issue. I may be wrong though.

roderickvd: Yes, it can. In the case of FLAC, blocks do not need to be frame-aligned. Generally, I/O corruption can and does occur in the wild. And then there is this: https://github.com/RustAudio/rodio/pull/684#issuecomment-2612661942

src/decoder/wav.rs
```diff
@@ -82,58 +92,83 @@ where
...
+            match next_sample {
+                Some(sample) => {
+                    self.samples_in_current_frame =
+                        (self.samples_in_current_frame + 1) % self.channels.get() as usize;
+                    return Some(sample);
                 }
-            };
-        next_sample
+                None => {
+                    // Input exhausted - check if mid-frame
```
yara-blue, line 161: strange, if a I read the wav spec correctly there is only a single header at the start of the file. So how would the parameters change... Did I miss something?

This would still protect against erroneous files I guess, where the header is wrong (claiming more samples in the file then there are). Is that why this is here?

roderickvd: This code is not to check for parameter changes, but to ensure frame-alignment. A truncated WAV file could easily cause only part of the channels to be put out.

src/source/agc.rs
```diff
@@ -505,13 +539,53 @@ where
...
-        self.input.next().map(|sample| {
-            if self.is_enabled() {
-                self.process_sample(sample)
-            } else {
-                sample
+        let current_sample_rate = self.input.sample_rate();
+        let current_channels = self.input.channels();
+        let input_span_len = self.input.current_span_len();
+
+        let (at_boundary, parameters_changed) = detect_span_boundary(
```
yara-blue: I'm kinda scared this is going to destroy performance. I guess that's fine once we introduce the fixed source as non experimental (since then we have an alternative). Would you have any numbers on that? I believe AGC is in the benchmark suite somewhere.

src/source/take.rs
```diff
@@ -106,9 +111,27 @@ where
...
+        loop {
+            // If we're padding to complete a frame, return silence.
+            if self.silence_samples_remaining > 0 {
+                self.silence_samples_remaining -= 1;
+                return Some(Sample::EQUILIBRIUM);
+            }
+
+            // Check if duration has expired.
+            if self.remaining_duration < self.duration_per_sample {
+                self.silence_samples_remaining =
```
yara-blue, line 117: this breaks the promise of take duration, should it not instead pass on a modified last span accounting for the target duration?

yara-blue: a similar thing could work for lineargainramp (there we would cut a span into two)

src/source/agc.rs
yara-blue: might be more difficult then I thought since queue should get faster from this

src/decoder/flac.rs
yara-blue: :cry:

src/source/take.rs
roderickvd: Yes, good point.

src/source/agc.rs
roderickvd: Not insignificant. `cargo bench` is 8,9% slower on my MacBook Air M3. 15,6% slower on a Raspberry Pi 4. And I don't see any other way if we need to make spans work correctly.

Actually I have one or two thoughts about that, that I'll put in with the PR discussion so it doesn't get lost here.

roderickvd:
Indeed! And the sad part is that I expect that 90% of use cases will never use it.

Two things I can think of:
1. Expedite the introduction of your new source architecture: ensure that sources output to a unified sample rate, always.
2. Adding a `stable_parameters` or similar/reciprocal method to the decoder builder, to configure eagerly reporting `None` as span length even if a decoder _could_ theoretically have its parameters changed.

Point 2 is easily done and I had already put in... well you know.

Thanks for your review.

yara-blue:
Those perf. changes are indeed not insignificant, bummer. I would like to merge this sooner rather then later though (as it's not really isolated). How about we cut a release this week and then merge this after? That gives us until the next release to provide more performance either through 1 or 2. 

I feel confident we can land the experimental source architecture within half a year, so before the next release. I'm using it right now in a side project and I'm really happy with how that code looks.

example:
https://github.com/yara-blue/mpdhaj/blob/4f781b71558add4fbe6a503991e58ee86727d475/src/player.rs#L136

roderickvd:
> Those perf. changes are indeed not insignificant, bummer. I would like to merge this sooner rather then later though (as it's not really isolated). How about we cut a release this week and then merge this after? That gives us until the next release to provide more performance either through 1 or 2.

OK for me.

src/source/take.rs
roderickvd: Actually, thinking about it more, is the contract for `TakeDuration` that get you get:
a. precisely the duration you've asked for (zero-padded if necessary)
b. _up to_ the duration you asked for, or less if the source is exhausted before that?

I think it's b and then this is good to go. This implementation even contains a test that we never end mid-frame / always emit a full frame.

src/source/blt.rs
```diff
@@ -116,66 +123,399 @@ where
...
+impl<I> BltInner<I>
+where
+    I: Source<Item = Sample>,
+{
+    fn new(input: I, formula: BltFormula, channels: ChannelCount) -> Self {
+        let channels_count = channels.get() as usize;
+
+        let sample_rate = input.sample_rate();
+        let applier = formula.to_applier(sample_rate.get());
+
```
yara-blue, line 203: TODO(yara): Copy the new multichannel logic to the fixed sources.

src/source/channel_volume.rs
yara-blue, line 34: Unrelated to the PR. Is this not wrong?

A frame starts at channel zero right? So this should pick the channel_volume for channel zero.

src/source/linear_ramp.rs
```diff
@@ -18,14 +18,21 @@ where
...
         total: duration,
         start_gain,
         end_gain,
         clamp_end,
-        sample_idx: 0u64,
+        sample_idx: 0,
+        samples_counted: 0,
+        cached_span_len: None,
+        last_sample_rate: sample_rate,
+        last_channels: channels,
```
yara-blue: these params (`cached_span_len`, `last_sample_rate` and `last_channels`) seem to be shared by most sources now. Is this something we should extract to a struct? 

like:
```rust
struct PrevParams {
    span_len,
    sample_rate,
    channel_count,
}
```

yara-blue: maybe even params, then the detect_span_boundary could maybe take, prev and current both of type Params.

Going even further,  a private extension trait implemented for everything Source could provide a `.params(&self) -> Params` method

src/source/mod.rs
```diff
@@ -830,3 +850,141 @@ source_pointer_impl!(Source for Box<dyn Source + Send>);
 source_pointer_impl!(Source for Box<dyn Source + Send + Sync>);
 
 source_pointer_impl!(<'a, Src> Source for &'a mut Src where Src: Source,);
+
+/// Detects if we're at a span boundary using dual-mode tracking.
```
yara-blue: this line does not add much for me (I dont know what dual-mode tracking means until I've read the function body)

src/source/mod.rs
```diff
@@ -830,3 +850,141 @@ source_pointer_impl!(Source for Box<dyn Source + Send>);
...
+/// Returns a tuple indicating whether we're at a span boundary and if parameters changed.
```
yara-blue: how about we return a struct with fields named: `at_span_boundary` and `parameters_changed` or something along does lines.

src/source/mod.rs
```diff
@@ -830,3 +850,141 @@ source_pointer_impl!(Source for Box<dyn Source + Send>);
...
+    } else {
+        0
+    }
+}
+
+#[cfg(test)]
+pub(crate) mod test_utils {
+    use super::*;
+
+    /// Test helper source that can end mid-frame for testing incomplete frame handling.
```
yara-blue: but we do not support ending mid-frame? 
how this is different from `SamplesBuffer`?

src/source/position.rs
```diff
@@ -1,20 +1,25 @@
 use std::time::Duration;
```
yara-blue, line 1: done till here. Might continue tomorrow

yara-blue: status update: since this overhauls almost every source I'm taking this one slow. I'm adding the source to rodio-experiments before I handle it, that way I can get read up on the original impl before I see this (makes an in depth review easier)

src/source/take.rs
yara-blue: I would say b, since you can compose a by applying `TakeDuration` on the rodio equivalent of source.chain(silence).

src/source/zero.rs
yara-blue, line 1: mhm this behavior was always redundant (inf zero + take dur). Especially now we are size hinting all the other generators as infinite we should consider bringing this in line.

src/source/zero.rs
```diff
@@ -29,13 +29,23 @@ impl Zero {
...
+    /// # Panics
+    ///
+    /// Panics if `num_samples` is not a multiple of `channels`.
     #[inline]
     pub fn new_samples(
         channels: ChannelCount,
         sample_rate: SampleRate,
         num_samples: usize,
     ) -> Self {
+        assert_eq!(
```
yara-blue: Rather have this return a Result instead, it should then just have a single struct as error type.

src/source/zero.rs
yara-blue: out of scope for this PR though

src/mixer.rs
```diff
@@ -32,9 +32,10 @@ pub fn mixer(channels: ChannelCount, sample_rate: SampleRate) -> (Mixer, MixerSo
     }));
 
     let output = MixerSource {
-        current_sources: Vec::with_capacity(16),
```
yara-blue, line 35: we ideally never allocate here. The 16 sources give us some headway. Its not the best solution, that would probably be tracking capacity and allocating a new Vec in the MixerHandle and then taking that new allocation together with the new Source.

But that is definitely out of the scope of this PR!! 

For now maybe leave the 16 sources pre-allocated.

src/mixer.rs
```diff
@@ -32,9 +32,10 @@ pub fn mixer(channels: ChannelCount, sample_rate: SampleRate) -> (Mixer, MixerSo
...
+        current_sources: Vec::new(),
         input: input.clone(),
-        sample_count: 0,
+        current_channel: 0,
+        still_pending: Vec::new(),
         pending_rx: rx,
```
yara-blue, line 38: this changed on main (the still pending list got removed entirely)

src/mixer.rs
```diff
@@ -140,8 +173,16 @@ impl MixerSource {
...
+            // to ensure correct channel alignment
+            let in_step = self.current_channel == 0;
+
+            if in_step {
+                self.current_sources.push(source);
+            } else {
+                self.still_pending.push(source);
+            }
         }
     }
```
yara-blue, line 147: take care of on main as well

src/queue.rs
```diff
@@ -423,109 +383,4 @@ mod tests {
             );
         }
     }
-
-    #[test]
-    fn span_ending_mid_frame() {
```
yara-blue, line 428: why is this test gone? It seems quiet useful

tests/total_duration.rs
```diff
@@ -77,15 +72,14 @@ fn get_music(format: &str) -> Decoder<impl Read + Seek> {
 
 #[cfg(any(
     feature = "claxon",
-    feature = "minimp3",
```
yara-blue, line 80: why is minimp3 removed here?

src/source/channel_volume.rs
roderickvd: I think I did that to trigger an initial reload in `next`. I agree it could either use a comment or have refactor the reload out into a separate function. Let me know if you want me to take something on.

src/source/linear_ramp.rs
roderickvd: Sounds all good. I don’t know precisely where you are with per-source merging so again let me know if you’d like me to take something on.

src/source/mod.rs
roderickvd: Yeah that’s be richer info. Hypothetically, sources may need to know one but not the other.

src/source/mod.rs
roderickvd: Yes we should clarify the modes or actually states.

src/source/mod.rs
roderickvd: On mobile, but it: does? `SamplesBuffer` should now enforce it is frame aligned. This source doesn’t.

src/source/position.rs
roderickvd: I understand so just tell me what you want me to pick up, if anything.

src/source/take.rs
roderickvd: OK, this should be good then as-is, unless I’m misunderstanding.

src/source/zero.rs
roderickvd: Also we shouldn’t call it zero. That’s not equilibrium for unsigned integers. Yes it’ll produce silence but at full DC, burning your drivers.

src/source/zero.rs
roderickvd: Agreed. Did this to keep the API stable. But we don’t need to.

src/mixer.rs
roderickvd: OK, no strong opinions here.

src/mixer.rs
roderickvd: OK so I’m assuming no changes needed here: will happen as you merge source by source.

src/queue.rs
roderickvd: With the new source contract, spans can no longer end mid-frame.

tests/total_duration.rs
roderickvd: The minimp3 implementation here does not support returning a duration.

roderickvd:
Sorry I as processing from top to bottom until I saw your last comment. So you do want me to make the changes here, not as-you-go? Just confirming.

src/source/linear_ramp.rs
yara-blue: My plan was to get this and the re-sampler merged first and then slowly bring that in. Don't worry there will be a ton to review :sob:.

src/source/mod.rs
yara-blue: Hypothetically, sources may need to know one but not the other. We trust the compiler to rip the unused part out :)

src/source/mod.rs
yara-blue: Okay, but the incomplete frame alignment behavior is not used anywhere with this PR applied right?

src/source/position.rs
yara-blue: quiet soon an insane amount of reviewing (but don't worry we will split it into chunks of around 600 lines, it's pretty modular) .

I'm almost done refactoring for the 6th time :)

src/source/zero.rs
yara-blue: I've called it Silence for the new FixedSource and ConstSource. As part of that overhaul this will be removed. 

Since generators never change their parameters they have no span so it does not make senses to have a generator implement Source if it can instead implement the more strict FixedSource and ConstSource.

ConstSource can be turned into a FixedSource easily and for free the same is true for turning a FixedSource into a Source.

src/source/zero.rs
yara-blue: Actually this will be removed soon so I don't think it matters, that is if you agree that having generators implement Source is redundant if they already implement FixedSource and ConstSource.

src/mixer.rs
yara-blue: mhm I thought I pushed an elegant change, turns out I am wrong. Sorry for the confusion

yara-blue:
> Sorry I as processing from top to bottom until I saw your last comment. So you do want me to make the changes here, not as-you-go? Just confirming.

Yes confirmed, we will merge this to main after you make the changes here on this PR. Let me know when everything is ready then I'll give it a second pass and merge.

Nothing else will be merged to main until that is done to make sure we get no further conflicts. So take your time do not feel rushed!

roderickvd:
@yara-blue rebased and addressed your review comments.

roderickvd:
I wonder if we should replace this `ZeroError` with something more universally applicable like an `AlignmentError` or a `SourceError::NotAligned`.

yara-blue:
> I wonder if we should replace this `ZeroError` with something more universally applicable like an `AlignmentError` or a `SourceError::NotAligned`.

We could, though I would propose we resolve by removing it entirely and making the effects strictly mono. Then offer two ways to create multi channel effects:
1. as mixing multiple mono sources into stereo/multi channel audio
1. simply repeating the mono sample (channel convertor) 

1) would be a new feature I think? So we can implement that later while 2) will be  provided by `FixedSource::channel_count_count`.

I think this is neat since it makes the overall effects API simpler and moves the channel count decision into the audio pipeline.

yara-blue:
Anyway, this PR looks great. The `SpanTracker` you introduced nicely abstracts the messy reality.

I know it has taken quiet some effort to get this all sorted. I thought about doing all this myself at some point and gave up. Thank you for persevering through!

roderickvd:
> > I wonder if we should replace this `ZeroError` with something more universally applicable like an `AlignmentError` or a `SourceError::NotAligned`.
> 
> We could, though I would propose we resolve by removing it entirely and making the effects strictly mono. 

That’s a good idea. Then we don’t even need `Zero::new_samples` anymore. Just a combination of `Zero` (or better: `Silence` in your experiments) with `TakeDuration`.

yara-blue:
> > > I wonder if we should replace this `ZeroError` with something more universally applicable like an `AlignmentError` or a `SourceError::NotAligned`.
> > 
> > 
> > We could, though I would propose we resolve by removing it entirely and making the effects strictly mono.
> 
> That’s a good idea. Then we don’t even need `Zero::new_samples` anymore. Just a combination of `Zero` (or better: `Silence` in your experiments) with `TakeDuration`.

I could not resist
image

the macro's that take care of the implementation are straight up evil. The user has to explicitly set the output channel count (the 24 in the example) because we can not yet have const expressions in const argument positions. I did add a nice error as well in case channel counts do not line up:

image

Note this happens compile time (we panic in a const function). Getting that error formatted was .... a bit of work. Const string formatting isn't in yet so this all happens by copying bytes around in cons functions (and some log10's to get the digits for the channel counts).

roderickvd:
You scare me. 

😉

PR #833, opened 2026-01-15
fix: `SymphoniaDecoder::current_span_length` now returns `None` instead of `Some(0)`

### LINKED ISSUE #832: `SymphoniaDecoder::current_span_length` returns `Some(0)`
I have an issue where I create a source with `Decoder::new_vorbis(file)` and play it later with `mixer.add(source)`. It isn't playing anything at all. I verified that a sinewave will play instead.

Then I checked what the `Decoder` returns. It does return samples if I just iterate over it, but it returns `Some(0)` for `current_span_length` - which you must only return if there are no more samples.

Going through the call stack `SymphoniaDecoder::current_span_length` returns `Some(self.buffer.len())`. I think the decoder would need to decode a packet if `self.buffer.is_empty()` to report an accurate span length.

jgraef:
Fixes #832 

My first idea was to make `current_span_len` decode a new packet if the buffer is empty, but `current_span_len` only gets an immutable borrow of `self` and thus can't do that. As a fix I just added a check that returns `None` instead of `Some(0)`.

After I applied this patch the music started playing :notes:

### DIFF
--- a/src/decoder/symphonia.rs
+++ b/src/decoder/symphonia.rs
@@ -113,8 +113,8 @@ impl SymphoniaDecoder {
                 continue;
             }
 
-            match decoder.decode(&current_span) {
-                Ok(decoded) => break decoded,
+            let decoded = match decoder.decode(&current_span) {
+                Ok(decoded) => decoded,
                 Err(e) => match e {
                     Error::DecodeError(_) => {
                         // Decode errors are intentionally ignored with no retry limit.
@@ -124,6 +124,15 @@ impl SymphoniaDecoder {
                     }
                     _ => return Err(e),
                 },
+            };
+
+            // Loop until we get a packet with audio frames. This is necessary because some
+            // formats can have packets with only metadata, particularly when rewinding, in
+            // which case the iterator would otherwise end with `None`.
+            // Note: checking `decoded.frames()` is more reliable than `packet.dur()`, which
+            // can resturn non-zero durations for packets without audio frames.
+            if decoded.frames() > 0 {
+                break decoded;
             }
         };
         let spec = decoded.spec().to_owned();


roderickvd:
Thank you for the PR to go with your issue 🚀 

While this fix works, I rather suggest that in `next()` we loop / advance until we got a packet with actual samples or the source is exhausted. That will be more correct than masking `Some(0)`.

Also if you would add a changelog entry?

jgraef:
I briefly thought about this solution too, but saw an issue with it and wanted a quick fix.

Checking again, I think I don't see any issues with this solution anymore. So after we get a sample in `next`, we'd check if the buffer is now exhausted, and if it is, we'd fetch the next?

This would almost make the loop before we get the sample unnecessary, except when the decoder is initialized with an empty buffer (and maybe other edge-cases I'm unaware of).

Aha! I remember the issue I saw. Your suggestion would not fix the issue, since it would still return the incorrect `current_span_length` if you have never called `next` on the decoder. We'd also have to run that decode loop when the decoder is initialized and really anywhere where we could end up with an empty buffer.

I can work on this, but I would first need to read and understand the code more thoroughly.

roderickvd:
I see your point. A complete fix is in #786  but was rejected.
See: https://github.com/RustAudio/rodio/blob/feat/comprehensive-seeking-and-bit-depth/src/decoder/symphonia.rs

The issue with returning `None` is that once you have, filters later in the chain may never check span boundaries again (because they start assuming they don't have to).

jgraef:
Did you link the right PR? It seems unrelated.

> The issue with returning None is that once you have, filters later in the chain may never check span boundaries again (because they start assuming they don't have to).

Imho it's better than returning `Some(0)` ;)

roderickvd:
> Did you link the right PR? It seems unrelated.

Corrected: #786

jgraef:
Having 2 separate loops that decode the next relevant packet is the problem here. It leads to the bug #775 and I think I spotted another one.

Bug #775 is now fixed by properly skipping to the next non-empty frame in the loop in `init` (like `Iterator::next` does).

The other bug: `init` checks if the packet belongs to the selected track, while `next` doesn't. I added a check for this in `next`.

I'd like to merge both loops into a single function/method, but they have slightly different error handling.

jgraef:
I also wanted to write a test for this specific bug. Does rodio have an established way of generating small audio files for this?

roderickvd:
We love tests, but don't have any standard way of generating audio files.

yara-blue:
Is this still in progress?

jgraef:
I haven't figured out a way to make a test or it. Otherwise the fix is ready to be merged, if you don't find any issues with it.

yara-blue:
In that case lets merge it, and hope nothing explodes :) 

Thanks for helping out!

PR #834, opened 2026-01-15
Fix compiler warnings for unused code when compiling without decoder

jgraef:
I use `--no-default-features -F vorbis -F playback` which causes an unused code warning. While fixing this I found more warnings that you get when compiling without any decoder feature enabled. This PR fixes these warnings

### DIFF
--- a/src/decoder/builder.rs
+++ b/src/decoder/builder.rs
@@ -302,7 +302,10 @@ impl<R: Read + Seek + Send + Sync + 'static> DecoderBuilder<R> {
         }
 
         #[cfg(not(feature = "symphonia"))]
-        Err(DecoderError::UnrecognizedFormat)
+        {
+            let _ = data;
+            Err(DecoderError::UnrecognizedFormat)
+        }
     }
 
     /// Creates a new decoder with previously configured settings.
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -101,6 +101,7 @@ pub struct Decoder<R: Read + Seek>(DecoderImpl<R>);
 pub struct LoopedDecoder<R: Read + Seek> {
     /// The underlying decoder implementation.
     inner: Option<DecoderImpl<R>>,
+
     /// Configuration settings for the decoder.
     settings: Settings,
 }
@@ -249,7 +250,10 @@ impl<R: Read + Seek> DecoderImpl<R> {
             DecoderImpl::Mp3(source) => source.try_seek(pos),
             #[cfg(feature = "symphonia")]
             DecoderImpl::Symphonia(source, PhantomData) => source.try_seek(pos),
-            DecoderImpl::None(_, _) => unreachable!(),
+            DecoderImpl::None(_, _) => {
+                let _ = pos;
+                unreachable!()
+            }
         }
     }
 }
@@ -622,6 +626,7 @@ where
 
             // Take ownership of the decoder to reset it
             let decoder = self.inner.take()?;
+
             let (new_decoder, sample) = match decoder {
                 #[cfg(all(feature = "hound", not(feature = "symphonia-wav")))]
                 DecoderImpl::Wav(source) => {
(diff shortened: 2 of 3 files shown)


src/source/mod.rs
```diff
@@ -231,7 +231,7 @@ pub trait Source: Iterator<Item = Sample> {
     #[cfg(feature = "dither")]
     #[cfg_attr(docsrs, doc(cfg(feature = "dither")))]
     #[inline]
-    fn dither(self, target_bits: BitDepth, algorithm: DitherAlgorithm) -> Dither<Self>
+    fn dither(self, target_bits: crate::BitDepth, algorithm: DitherAlgorithm) -> Dither<Self>
```
roderickvd, line 234: What is your rationale here?

src/decoder/builder.rs
```diff
@@ -302,7 +302,10 @@ impl<R: Read + Seek + Send + Sync + 'static> DecoderBuilder<R> {
         }
 
         #[cfg(not(feature = "symphonia"))]
-        Err(DecoderError::UnrecognizedFormat)
+        {
+            let _ = data;
+            Err(DecoderError::UnrecognizedFormat)
```
roderickvd, line 307: Tagging `#[allow(unused_variables)]` may convey intent more clearly.

src/decoder/mod.rs
```diff
@@ -249,7 +250,10 @@ impl<R: Read + Seek> DecoderImpl<R> {
             DecoderImpl::Mp3(source) => source.try_seek(pos),
             #[cfg(feature = "symphonia")]
             DecoderImpl::Symphonia(source, PhantomData) => source.try_seek(pos),
-            DecoderImpl::None(_, _) => unreachable!(),
+            DecoderImpl::None(_, _) => {
+                let _ = pos;
```
roderickvd, line 254: Trick here is to make the argument `_pos` in the function signature.

src/decoder/mod.rs
```diff
@@ -667,9 +672,16 @@ where
                     let sample = source.next();
                     (DecoderImpl::Symphonia(source, PhantomData), sample)
                 }
+                DecoderImpl::None(unreachable, phantom) => {
```
roderickvd, line 675: Or `DecoderImpl::None(_, _) => unreachable!()`?

roderickvd (commented):
Some ideas to make it more idiomatic and convey clearer intent.

src/source/mod.rs
jgraef: If you import `crate::BitDepth` and then just use `BitDepth` you'll get a warning about the unused import when the `dither` feature is not enabled.

We can also put the import behind a feature flag, but then we end up with feature-gated stuff spread more out and it's just easier to accidentally mess up.

roderickvd: 👍

src/decoder/builder.rs
jgraef: I don't think so. It would just allow unused variables (generally) and I think it would need to go on the function scope, and you'd want to only disable the lint for `not(feature = "symphonia")`.

In my opinion the last block - whether symphonia is enabled, or not - should use the data somehow, and that's what this does.

It's late and I think I'm failing to describe why I think using `let _ = data;` is better. Maybe tomorrow it will be clearer for either of us.

src/decoder/mod.rs
jgraef: Starting a variable with underscore is used for unused variables. `pos` is not generally unused.

src/decoder/mod.rs
jgraef: Then the code after the match statement is dead code and causes a warning.

src/decoder/mod.rs
roderickvd: You can use it as `_pos` too: `source.try_seek(_pos)`.

jgraef: Yes, but naming it `_pos` communicates the intent of it being unused - which it is not generally.

yara-blue:
Sorry for sleeping on this PR! Was not feeling well for a bit but back now. Thanks for helping us out with this!

PR #838, opened 2026-01-29
add get_ref to SampleRateConverter

MarinPostma:
allows inspecting underlying iterator with mutable access


roderickvd:
Huh, everywhere else we're implementing this and call it `inner`. Weird that we didn't do that here. Would you rename it to `inner` and add a changelog entry?

MarinPostma:
@roderickvd done

PR #842, opened 2026-02-06
Fixed Player::skip_one not updating player's length immediately.

### LINKED ISSUE #497: `Sink::skip_one` doesn't modify the sink's length accordingly
I have this problem even after #494 .

```
#[cfg(test)]
mod tests {
    use rodio::{source::SineWave, OutputStream, Sink, Source};
    use std::time::Duration;

    #[test]
    fn skip_one_test() {
        let (_stream, stream_handle) = OutputStream::try_default().unwrap();
        let sink = Sink::try_new(&stream_handle).unwrap();
        let source = SineWave::new(440.0)
            .take_duration(Duration::from_secs_f32(5.))
            .amplify(0.20);
        let source2 = SineWave::new(440.0)
            .take_duration(Duration::from_secs_f32(5.))
            .amplify(0.20);

        sink.append(source);
        sink.append(source2);

        assert_eq!(sink.len(), 2); // OK as expected

        sink.skip_one();

        assert_eq!(sink.len(), 1); // shouln't this be true? I am getting sink.len() == 2 instead
    }
}
```

nieboczek:
Fixes #497.
Needed to add `Done::should_decrement` to achieve this fix.

Instead of updating `sound_count` in the periodic access source, I made it update directly in the `skip_one` function (`clear` affected too).
When `controls.to_clear > 0` in periodic access, it will use `Done::should_decrement` to temporarily disable decrementing `sound_count`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+- Breaking: `Done` now calls a callback instead of decrementing an `Arc<AtomicUsize>`.
+- Added `Skippable::skipped` function to check if the inner source was skipped.
+- Fixed `Player::skip_one` not decreasing the player's length immediately.
 
 ## Version [0.22.2] (2026-02-22)
 
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -6,7 +6,10 @@ The list below only contains required code changes. For a complete list of
 changes and new features, see [CHANGELOG.md](CHANGELOG.md).
 
 # rodio 0.22 to current github version
-Nothing yet!
+- `Done` now calls a callback instead of decrementing an `Arc<AtomicUsize>`.
+  - To retain old behavior replace the `Arc<AtomicUsize>` argument in `Done::new` with
+    `move |_| { number.fetch_sub(1, std::sync::atomic::Ordering::Relaxed) }`.
+  - `Done` has now two generics instead of one: `<I: Source, F: FnMut(&mut I)>`.
 
 # rodio 0.21.1 to 0.22
 - _Sink_ terms are replaced with _Player_ and _Stream_ terms replaced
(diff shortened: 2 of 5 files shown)


nieboczek:
Something very bad seems to be happening with the Player, adding a simple `assert_eq` hangs the test on `Player::clear`:
```
        // ...
        assert_eq!(player.len(), 1);

        assert_eq!(source.next(), Some(1.0));

        player.clear();
        // ...
```
~~Could I get some help on this?~~ Fixed.

nieboczek:
I've hopefully fixed that. Also saw a queue issue where it wouldn't play the first audio appended in my own app with the version from the PR, but that is working correctly after my latest commit.

yara-blue:
Hii,

Thanks for working on this! Sorry I could only get to this review now (was sick for a bit but all is well now). 

I'm a little hesitant, I'm not sure `should_decrement` is useful outside fixing the `Player` and once we add a new API it's hard to remove it. Maybe we should rethink `Done` and see if we can make it more useful in general. That might allow a cleaner fix, even though it would involve a breaking change. I think a breaking change is fine if it enables new use cases. 

An idea:
We could make `Done::new` take a closure as argument and run that when the source finishes. Something like:

```rust
pub struct Done<I, F: FnMut()> {
    input: I,
    callback: F,
    signal_sent: bool,
}

    pub fn new(input: I, callback: F) -> Done<I, F> {
        Done {
            input,
            callback,
            signal_sent: false,
        }
    }
```

Then users could get the current behavior by calling: 
```rust
let mut signal_sent = false;
Done::new(source, || 
        if !signal_sent {
            signal.fetch_sub(1, Ordering::Relaxed);
            signal_sent = true;
        }
)
```
To stress: this is just an idea I'm not sure if this is a better solution but I'd love to have your thoughts.

So what do you think:
- Would this allow a cleaner solution for this issue?
- Is such a `Done` source more useful then the one we have here?

nieboczek:
I removed `should_decrement` as I also think it wouldn't be very useful.
`Done` now accepts a callback instead of an `Arc<AtomicUsize>` (breaking).
Also added `Skippable::skipped` that returns whether the skip has happened to achieve this fix easier.
`skipped` seems more useful than `should_decrement` though :)

yara-blue:
skipped seems pretty useful yeah! :)

yara-blue:
thanks for the great PR (great to have an upgrade guide entry too!)

PR #843, opened 2026-02-09
feat: high-quality resampler

roderickvd:
_Q: How hard could it be to add a high-quality `Resample` source using Rubato?
A: Almost four weeks of evening work, that's how hard!_

It's been surprisingly hard to get the details down. Way more than "just" creating a Rubato instance and processing some buffer. Very grateful for the test cases that were already there - without them, I'd never have gotten this down.

# Features
- **Builder pattern:**
```rust
// Quick presets
let config = ResampleConfig::fast();
let config = ResampleConfig::balanced();  // same as default()
let config = ResampleConfig::accurate();

// Full customization
let config = ResampleConfig::sinc()
    .sinc_len(nz!(256))
    .interpolation(Sinc::Cubic)
    .window(WindowFunction::BlackmanHarris2)
    .chunk_size(nz!(512))
    .build();

source.resample(target_rate, config);
```

- **Polynomial and sinc interpolation:** including support for FFT-based resampling for fixed ratios.

- **Fixed-ratio optimization:** auto-detects and switches sinc resamplers to FFT-based or nearest-neighbor processing when the ratio allows. This lowers CPU usage while providing the highest quality.

- **Span boundary handing:** specialized implementation of #831 to recreate the resampler on the fly when parameters change at span boundaries, even after seeking.

# Performance

Through refactoring of `UniformSourceIterator` and friends, the performance hit on `cargo bench` is now 7-8% - which is about the same as the span boundary detection from #831. Considering this PR _also_ adds span boundary detection for the resampler, that means the improved resampling is virtually free.

Initially I had made `SampleRateConverter` a simple wrapper around the new `Resample` source. This caused a 20-25% performance hit, and led to my refactoring described next.

# Changes

- The `UniformSourceIterator` now is optimized for passthrough, channel count conversion, sample rate conversion, or both.  It also inlines the former `Take` struct (private; not to be confused with `TakeDuration`) and works with `Resample` directly.

- For `UniformSourceIterator` to work with `Resample` directly, it needs to convert an `Iterator` into a `Source`. That's `from_iter` right? Wrong in v0.21: there `FromIter` is actually something that chains sources together. That seemed counter-intuitive to me. So to bring it in line with Rust idioms:
  - `FromIter` was renamed to `Chain` (chains sources from an iterator)
  - `FromFactory` was renamed to `FromFn` (chains sources from a function)
  - `FromIter` now wraps an `Iterator<Item = Sample>` into a `Source`

- `SourcesQueueOutput` was copied from #831 because it contains peeking fixes necessary to make `UniformSourceIterator` detect the sample rate correctly when transitioning out of silence.

- `SampleRateConverter` is tagged as deprecated.

- `ChannelCountConverter` now implements `Source` while keeping a minimum surface for just `I: Iterator`.

- Minor opportunistic refactoring.

# Questions

- Should we move and rename `Resample` from `src/source/resample.rs` to `SampleRateConverter` in `src/conversions/sample_rate.rs` and basically replace it?
  - This would be breaking due to requiring that `I: Source` instead of `I: Iterator`.
  - `Resample::new` could live besides `new_poly`, `new_sinc` and the builder.

# Related

Supersedes #670


yara-blue:
> _Q: How hard could it be to add a high-quality `Resample` source using Rubato? A: Almost four weeks of evening work, that's how hard!_

Oops, I've had this for a while in rodio experiments https://github.com/RustAudio/rodio-experiments/blob/main/src/conversions/resampler/variable_input.rs

but hey that's nice we can compare :)

and its probably not as extensive (for example I said hell no to seeking :))

roderickvd:
Yes the hard work started when I thought I was done and started merging the tests from the existing `SampleRateConverter` 😆 

Beyond padding blocks to satisfy Rubato’s chunks there is also output delay skipping, cutting off of mirror images after the last sample (too many samples) and flushing the last signal energy from Rubato’s filter (not enough samples).

This complexity required some degree of complexity to address it - if you see opportunities to simplify then do let me know. The test suite quickly reveals what you cannot take out…

roderickvd:
Split into different files now.

examples/resample.rs
```diff
@@ -0,0 +1,118 @@
...
+use rodio::{Decoder, Player};
+use std::error::Error;
+use std::fs::File;
+use std::io::BufReader;
+use std::time::Instant;
+
+fn main() -> Result<(), Box<dyn Error>> {
+    #[cfg(debug_assertions)]
+    {
+        eprintln!("WARNING: Running in debug mode. Audio may be choppy, especially with");
```
yara-blue, line 13: I wonder if we should not print this when building the Resampler. Seems like a footgun similar to dropping the Sink.

examples/resample.rs
```diff
@@ -0,0 +1,118 @@
...
+        eprintln!("         sinc resampling of non-integer ratios (async resampling).");
+        eprintln!("         For best results, compile with --release");
+        eprintln!();
+    }
+
+    let args: Vec<String> = std::env::args().collect();
+
+    if args.len() < 2 {
```
yara-blue, line 21: Try using [Clap](https://docs.rs/clap/latest/clap/) instead.

examples/resample.rs
```diff
@@ -0,0 +1,118 @@
...
+
+    println!("Playing resampled audio...");
+    println!("Press Ctrl+C to stop");
+
+    let playback_start = Instant::now();
+    player.append(resampled);
+    player.sleep_until_end();
+
+    let playback_time = playback_start.elapsed();
+    println!("\nPlayback finished in {:.2}s", playback_time.as_secs_f32());
```
yara-blue, line 94: I'd say for an example its fine to use the Debug fmt implementation for Duration

examples/resample.rs
```diff
@@ -0,0 +1,118 @@
...
+
+    Ok(())
+}
+
+/// Parse the resampling quality from a string argument
+fn parse_quality(method: &str) -> Result<ResampleConfig, Box<dyn Error>> {
```
yara-blue, line 100: If it is worth adding these in the example it might be worth add factories for them on ResampleConfig?

examples/resample.rs
```diff
@@ -0,0 +1,118 @@
...
+        "nearest" => ResampleConfig::poly().degree(Poly::Nearest).build(),
+        "linear" => ResampleConfig::poly().degree(Poly::Linear).build(),
+        "cubic" => ResampleConfig::poly().degree(Poly::Cubic).build(),
+        "quintic" => ResampleConfig::poly().degree(Poly::Quintic).build(),
+        "septic" => ResampleConfig::poly().degree(Poly::Septic).build(),
+        "very_fast" => ResampleConfig::very_fast(),
+        "fast" => ResampleConfig::fast(),
+        "balanced" => ResampleConfig::balanced(),
+        "accurate" => ResampleConfig::accurate(),
+        _ => return Err(format!(
```
yara-blue, line 111: This should be part of arg parsing, (don't worry clap does that for you if you make method an enum)

src/conversions/sample_rate.rs
```diff
@@ -1,388 +1,119 @@
...
+use crate::source::{resample::Poly, FromIter, Resample, ResampleConfig};
+use crate::{Sample, Source};
+
+/// Iterator that converts from one sample rate to another.
+#[deprecated(
+    since = "0.22.0",
+    note = "Use `Resample` with `FromIter` (or `from_iter` function) directly"
+)]
+#[derive(Debug)]
+#[allow(deprecated)]
```
yara-blue, line 11: what does the allow(deprecated) do here? Are we still using this internally?

src/source/resample/builder.rs
```diff
@@ -0,0 +1,452 @@
...
+    /// Linear interpolation between two nearest points.
+    ///
+    /// Relatively fast, but needs a large number of intermediate points to push the resampling
+    /// artefacts below the noise floor.
+    #[default]
+    Linear,
+
+    /// Quadratic interpolation using three nearest points.
+    ///
+    /// The computation time lies approximately halfway between that of linear and quadratic
```
yara-blue, line 68: I think you mean `and cubic`?

src/source/resample/builder.rs
```diff
@@ -0,0 +1,452 @@
...
+    /// No interpolation - picks nearest intermediate point.
+    ///
+    /// Optimal when upsampling by exact ratios (e.g., 48kHz and 96kHz) and the oversampling factor
+    /// is equal to the ratio. In these cases, no unnecessary computations are performed and the
+    /// result is equivalent to that of synchronous resampling.
+    Nearest,
+
+    /// Linear interpolation between two nearest points.
+    ///
+    /// Relatively fast, but needs a large number of intermediate points to push the resampling
```
yara-blue, line 61: This does not make sense to me, why do more points help? They are still all on a line right?

src/source/resample/mod.rs
```diff
@@ -0,0 +1,759 @@
...
+                resampler.resampler.output_frames_next(),
+            ),
+        };
+
+        let ratio = Ratio::new(self.sample_rate().get(), input_sample_rate.get());
+        if ratio.is_integer() {
+            // Integer upsampling (2x, 3x, etc.) - always exact and frame-aligned
+            input_span_len.map(|len| *ratio.numer() as usize * len)
+        } else {
+            // When the ratio contains a fraction, we cannot choose the floor or ceiling
```
yara-blue, line 319: 😿  😿  😿 

(great work, just feeling empathy for all you had to figure out)

src/source/resample/mod.rs
```diff
@@ -0,0 +1,759 @@
...
+    #[inline]
+    fn sample_rate(&self) -> SampleRate {
+        self.target_rate
+    }
+
+    #[inline]
+    fn channels(&self) -> ChannelCount {
+        match self.inner.as_ref().unwrap() {
+            ResampleInner::Passthrough { source, .. } => source.channels(),
+            ResampleInner::Poly(resampler) => resampler.channels,
```
yara-blue, line 344: I'd suggest using `source.channels()` on the `resampler`

src/source/resample/mod.rs
```diff
@@ -0,0 +1,759 @@
...
+            ResampleInner::Sinc(resampler) => resampler.channels,
+            #[cfg(feature = "rubato-fft")]
+            ResampleInner::Fft(resampler) => resampler.channels,
+        }
+    }
+
+    #[inline]
+    fn total_duration(&self) -> Option<Duration> {
+        match self.inner.as_ref().unwrap() {
+            ResampleInner::Passthrough { source, .. } => source.total_duration(),
```
yara-blue, line 354: I'would again use source directly

src/source/resample/mod.rs
```diff
@@ -0,0 +1,759 @@
...
+/// Maximum for optimized fixed-ratio resampling: 44.1 and 384 kHz (147:1280).
+const MAX_FIXED_RATIO: u32 = 1280;
+
+/// Resamples an audio source to a target sample rate using Rubato.
+#[derive(Debug)]
+pub struct Resample<I>
+where
+    I: Source,
+{
+    inner: Option<ResampleInner<I>>,
```
yara-blue, line 109: `// We keep this in an Option so we can easily swap it out

src/source/resample/mod.rs
```diff
@@ -0,0 +1,759 @@
...
+            ResampleInner::Passthrough { source, .. } => source.next()?,
+            ResampleInner::Poly(resampler) => resampler.next_sample()?,
+            ResampleInner::Sinc(resampler) => resampler.next_sample()?,
+            #[cfg(feature = "rubato-fft")]
+            ResampleInner::Fft(resampler) => resampler.next_sample()?,
+        };
+
+        // If input reports no span length, parameters are stable by contract
+        let input_span_len = self.inner.as_ref().unwrap().input().current_span_len();
+        if input_span_len.is_some() {
```
yara-blue, line 432: Early return would make this nicer

```rust
if input_span_len.is_none() {
    return Some(sample)
}
```

src/source/resample/mod.rs
```diff
@@ -0,0 +1,759 @@
...
+        let sample = match self.inner.as_mut().unwrap() {
+            ResampleInner::Passthrough { source, .. } => source.next()?,
+            ResampleInner::Poly(resampler) => resampler.next_sample()?,
+            ResampleInner::Sinc(resampler) => resampler.next_sample()?,
+            #[cfg(feature = "rubato-fft")]
+            ResampleInner::Fft(resampler) => resampler.next_sample()?,
+        };
+
+        // If input reports no span length, parameters are stable by contract
+        let input_span_len = self.inner.as_ref().unwrap().input().current_span_len();
```
yara-blue, line 431: lots of inner.as_ref().unwrap() etc. Might be worth having a self.inner() and self.inner_mut(), of course we can not have those because we already use them to return the underlying Source. Instead maybe `self.resampler`?

Let me know what you think.

src/source/resample/mod.rs
```diff
@@ -0,0 +1,759 @@
...
+                        (*channels, *source_rate, *input_samples_consumed)
+                    }
+                    ResampleInner::Poly(r) | ResampleInner::Sinc(r) => {
+                        (r.channels, r.source_rate, r.input_samples_consumed)
+                    }
+                    #[cfg(feature = "rubato-fft")]
+                    ResampleInner::Fft(r) => (r.channels, r.source_rate, r.input_samples_consumed),
+                };
+
+            // Get current parameters from input
```
yara-blue, line 451: Probably worth it to extract the boundary detection to a function. Could be called `at_boundary` return bool and take an `parameters_changed` as &mut. Or go the enum route and return a bespoke:
```rust
enum AtBoundary {
    Yes,
    ParametersChanged,
    No,
}
```

Or return two bools...

As you see I'm a little conflicted on what to do, still think it's worth it to extract, the logic here is non trivial and anything that makes it easier to parse would help future bugfixing/improvements etc.

src/source/resample/mod.rs
```diff
@@ -0,0 +1,759 @@
...
+    }
+
+    impl std::ops::Deref for TestChannelCount {
+        type Target = ChannelCount;
+        fn deref(&self) -> &Self::Target {
+            &self.0
+        }
+    }
+
+    struct TestSource {
```
yara-blue, line 584: Is this not `SamplesBuffer`?

src/source/resample/rubato.rs
```diff
@@ -0,0 +1,455 @@
...
+    pub resampler: R,
+
+    pub input_buffer: Box<[Sample]>,
+    pub input_frame_count: usize,
+
+    pub output_buffer: Box<[Sample]>,
+    pub output_buffer_pos: usize,
+    pub output_buffer_len: usize,
+
+    pub channels: ChannelCount,
```
yara-blue, line 82: I do not really see this field being used (except from resample/mod.rs).

src/source/resample/mod.rs
yara-blue, line 1: the `<module_name>/mod.rs` style we have throughout rodio is there for historical reasons. There is no reason to keep it, feel free to name the file resample.rs and have it in the source folder

src/source/resample/rubato.rs
yara-blue: the same is true for source_rate.

src/source/resample/rubato.rs
```diff
@@ -0,0 +1,455 @@
...
+        let num_channels = self.channels.get() as usize;
+        loop {
+            // If we have buffered output, return it
+            if self.output_buffer_pos < self.output_buffer_len {
+                let sample = self.output_buffer[self.output_buffer_pos];
+                self.output_buffer_pos += 1;
+                self.total_output_samples += 1;
+
+                if self.total_output_samples > self.expected_output_samples {
+                    // Cut off filter artifacts after input is exhausted
```
yara-blue, line 135: can we not instead set `self.output_buffer_len` to `min(self.output_buffer_len, self.expected_output_samples`)`?

src/source/resample/rubato.rs
```diff
@@ -0,0 +1,455 @@
...
+                let sample_pos = self.input_frame_count * num_channels;
+                for ch in 0..num_channels {
+                    if let Some(sample) = self.input.next() {
+                        self.input_buffer[sample_pos + ch] = sample;
+                    } else {
+                        self.input_exhausted = true;
+                        break;
+                    }
+                }
+                if !self.input_exhausted {
```
yara-blue, line 163: can we not drop the `!self.input_exhausted` in the while condition then move this to the top of the loop with a break?

src/source/resample/rubato.rs
```diff
@@ -0,0 +1,455 @@
...
+        self.total_output_samples = 0;
+        self.expected_output_samples = 0;
+        self.real_frames_in_buffer = 0;
+        self.indexing.partial_len = None;
+        self.output_delay_remaining =
+            Self::calculate_delay_compensation(&self.resampler, self.channels);
+    }
+
+    pub fn next_sample(&mut self) -> Option<Sample> {
+        let num_channels = self.channels.get() as usize;
```
yara-blue, line 126: at ~125L this function is rather unwieldy. I suggest extracting some functions. fill_input_buffer (I would include flushing with zeros) seems an easy candidate. I'll be honest I'm struggling a little to figure out how to cut up the end though...

src/source/resample/rubato.rs
```diff
@@ -0,0 +1,455 @@
...
+            // We can process with fewer frames than needed using partial_len when the input is
+            // exhausted. If we don't have enough input and more is coming, wait.
+            let made_progress = self.input_frame_count > frames_before;
+            if self.input_frame_count < needed_input && !self.input_exhausted && made_progress {
+                continue;
+            }
+
+            let actual_frames = self.input_frame_count;
+
+            // Prevent stack allocations in the hot path by reusing the indexing struct
```
yara-blue, line 186: I think you mean save space on the stack by keeping the indexing struct in the RubatoResample struct. Since `Self` is not boxed it's on the stack too so this makes no difference. To me it seems like this is just extra complexity. 

Let me know if I'm wrong and this sped things up!

src/source/resample/rubato.rs
```diff
@@ -0,0 +1,455 @@
...
+            }
+            self.input_frame_count -= actual_consumed;
+
+            self.output_buffer_pos = 0;
+            self.output_buffer_len = frames_out * num_channels;
+
+            // Skip warmup delay samples
+            if self.output_delay_remaining > 0 {
+                let samples_to_skip = self.output_delay_remaining.min(self.output_buffer_len);
+                self.output_buffer_pos += samples_to_skip;
```
yara-blue, line 248: Might be worth it to abstract the `Box<[Sample]> + pos` pattern into an object. Then we could simply call `output_buffer.skip(samples_to_skip);`.

src/source/resample/rubato.rs
```diff
@@ -0,0 +1,455 @@
...
+}
+
+// Async resampler (polynomial and sinc) implementations
+impl<I: Source> RubatoAsyncResample<I> {
+    pub fn new_poly(
+        input: I,
+        target_rate: SampleRate,
+        chunk_size: usize,
+        degree: Poly,
+    ) -> Result<Self, String> {
```
yara-blue, line 262: imho library errors should always be typed. We have thiserror so this would be as simple as:
```rust
#[derive(thiserror::Error, Debug)]
#[error("Failed to create resampler")]
struct CouldNotCreateResampler(#[from] rubato::ResamplerConstructionError);
```

Thanks to `#[from]` the `map_err()` is no longer needed and `?` works directly.

You can re-use this type for `new_fft`, if it's important to have the error print which resampler the user was trying to create you could make two error variants (though I'd say they can find this out from the backtrace).

src/source/resample/rubato.rs
```diff
@@ -0,0 +1,455 @@
...
+
+        let resampler = rubato::Async::new_sinc(
+            resample_ratio as _,
+            1.0,
+            &parameters,
+            chunk_size,
+            channels.get() as usize,
+            rubato::FixedAsync::Output,
+        )
+        .map_err(|e| format!("Failed to create sinc resampler: {:?}", e))?;
```
yara-blue, line 347: same error thing as above

src/source/resample/rubato.rs
```diff
@@ -0,0 +1,455 @@
...
+
+        let resampler = rubato::Fft::new(
+            source_rate.get() as usize,
+            target_rate.get() as usize,
+            input_chunk_size,
+            sub_chunks,
+            channels.get() as usize,
+            rubato::FixedSync::Output,
+        )
+        .map_err(|e| format!("Failed to create FFT resampler: {:?}", e))?;
```
yara-blue, line 419: once more :)

src/source/chain.rs
```diff
@@ -0,0 +1,181 @@
...
+        if let Some(src) = &self.current_source {
+            if let Some(val) = src.size_hint().1 {
+                if val < THRESHOLD && val != 0 {
+                    return Some(val);
+                }
+            }
+        }
+
+        // Otherwise we use the constant value.
+        Some(THRESHOLD)
```
yara-blue, line 110: I might be wrong here (I do not have the span misery mitigations clearly in my head) but can this not get dicey when I:
- create an empty chain and put it in a queue
- put a 48khz mono source in the queue 
- queue calls current_span_en get Some(THRESHOLD) on that queue
- queue calls sample_rate get 44100 on that queue
- queue calls next on chain gets None
- queue calls next on the mono source and gets a sample intended for 48khz but playing at 44100.

Now our queue may not call things in this order but what would happen with a user created one?

An alternative design for chain would forbid the empty state by getting the first sample in the `new` factory, then failing with an error if the first source was empty.

Again im pretty unsure about all this since the span misery has completely left my brain, thanks again for going through all that!

src/queue.rs
```diff
@@ -135,74 +135,43 @@ fn threshold(channels: ChannelCount) -> usize {
...
-            // Current source exhausted, peek at next queued source
-            // This prevents wrong resampling setup in UniformSourceIterator
-            next.sample_rate()
-        } else {
-            // Queue is empty, no sources queued
-            self.current.sample_rate()
+        if self.current.is_exhausted() && self.silence_samples_remaining == 0 {
+            if let Some((next, _)) = self.input.next_sounds.lock().unwrap().front() {
+                // Current source exhausted, peek at next queued source
+                // This prevents wrong resampling setup in UniformSourceIterator
```
yara-blue, line 169: there is a subtle race condition here:
- call sample_rate, current is exhausted
- no next source yet so get self.current.sample_rate()
- new source gets appended
- next call works without silence placeholder at possibly wrong sample rate

I'd suggest "poisoning" the queue whenever the current is exhausted and we do not have something new yet.

src/queue.rs
yara-blue, line 1: For now I've already reviewed them so it would just get confusing but in the future please seperate out changes that can be unrelated to multiple PR's. In this case I believe the changes to these files:
- uniform, 
- from_fn, 
- from_factory, 
- from_iter 
- chain()
Should have been separated out, now the whole PR will be held up by their review. It also makes it harder to have the whole PR reviewed (which is part of why it took so long to get this reviewed). For me a 200 line change is something I can sneak in my day somewhere while a large PR like this needs an entire afternoon.

yara-blue (changes requested):
Thank you I see why this was sooo much work there are a lot of subtleties! 

I've found a few rough edges that are worth sanding down. It's mostly code quality enhancements.

Due to it's size it took me a bit to find the time to review it. By now the PR needs a rebase, I propose we first address all the comments and only then rebase. Otherwise we'd lose where we are.

Proper resampling is the only way to get out of the span mess making this essential to the future of Rodio. Therefore let me thank you again for figuring out how this works and getting a solid implementation done.

src/conversions/sample_rate.rs
roderickvd: It's because `SampleRateConverter` is public and deprecated but the impl block references it.

src/source/resample/mod.rs
roderickvd: Makes sense. I'll add private `resampler()` / `resampler_mut()` helpers.

src/source/resample/mod.rs
roderickvd: I'll just extract it.

src/source/resample/mod.rs
roderickvd: `TestSource::current_span_len()` returns the total length (not remaining), which SamplesBuffer may not.

src/conversions/sample_rate/rubato.rs
```diff
@@ -0,0 +1,455 @@
...
+    pub resampler: R,
+
+    pub input_buffer: Box<[Sample]>,
+    pub input_frame_count: usize,
+
+    pub output_buffer: Box<[Sample]>,
+    pub output_buffer_pos: usize,
+    pub output_buffer_len: usize,
+
+    pub channels: ChannelCount,
```
roderickvd: `channels` is used inside `next_sample()` for frame calculations; `source_rate` is the the snapshot at construction time that `mod.rs` compares against to detect parameter changes. I can add comments to explain this.

src/conversions/sample_rate/mod.rs
roderickvd: I have no strong preference. If you want me to do this, then let me know and I'll take care of it after principal approval of the PR and final rebasing.

src/source/resample/rubato.rs
roderickvd: Yes, good catch.

src/source/resample/rubato.rs
roderickvd: Yes, clearer.

src/conversions/sample_rate/rubato.rs
```diff
@@ -0,0 +1,455 @@
...
+        self.total_output_samples = 0;
+        self.expected_output_samples = 0;
+        self.real_frames_in_buffer = 0;
+        self.indexing.partial_len = None;
+        self.output_delay_remaining =
+            Self::calculate_delay_compensation(&self.resampler, self.channels);
+    }
+
+    pub fn next_sample(&mut self) -> Option<Sample> {
+        let num_channels = self.channels.get() as usize;
```
roderickvd: I'll extract what seems reasonable.

src/source/resample/builder.rs
roderickvd: This is about the `oversampling_factor`on the sinc lookup table. I'll try to find a way to document something that's a little more user-friendly than that.

src/source/chain.rs
roderickvd: Span misery seems to have no limit to edge cases... so I'll add this one to the test suite and fix it.

src/queue.rs
roderickvd: The race is real and users would get the wrong sample rate in that split moment. However, the span boundary detection mechanism will pick up the right value right at the first new frame. So chances of this occurring are vanishingly small.

With the current interface I don't know how we could _not_ return a sample rate in that split jiffy?
For posterity I'll add a test that fails and ignore it.

roderickvd:
Addressed review points in https://github.com/RustAudio/rodio/pull/843/changes/97f5701b2eb7cbf99716039ad482aa98a90f2b32 and rebased - sorry, I already force-pushed without thinking about it.

This required backporting a fix from #786.

I'll investigate a bit more why the Hydrogen SRC test shows that we're missing samples. This PR took great care to be sample-exact, so I don't yet understand. That should not hold up merging.

roderickvd:
What about the open questions from the PR body?

roderickvd:
Upon further investigation, it turns out that the sample delay depends on Rubato's sinc settings. With something as extreme as this, the sample delay is 0 as it should be:

```rust
ResampleConfig::Sinc {
    sinc_len: 2048,
    oversampling_factor: 1024,
    interpolation: Sinc::Cubic,
    window: WindowFunction::BlackmanHarris2,
    f_cutoff: 0.9945994235, // 0.5^(16/sinc_len)
    chunk_size: 2048,
    sub_chunks: 1,
}
```

Note that that `f_cutoff` is higher than what Rubato's [`calculate_cutoff`](https://docs.rs/rubato/latest/rubato/fn.calculate_cutoff.html) would return. I believe this is because Rubato [optimizes for an Fs of 20 kHz](https://github.com/HEnquist/rubato/blob/63c2107383e3c0dc3419466b18a2b1b6ad87fda2/utils/cutoff_fit_cubic.py#L58), while the HydrogenAudio SRC test compares against Fs/2 or 22.05 kHz. Higher `f_cutoff` come at a cost of greater phase shift.

Finally, I can confirm that the spectrogram is clean when compiled with the `64bit` feature.

In all, I think this is good to go.

src/conversions/sample_rate/rubato.rs
```diff
@@ -0,0 +1,448 @@
...
+        self.total_input_frames = 0;
+        self.total_output_samples = 0;
+        self.expected_output_samples = 0;
+        self.real_frames_in_buffer = 0;
+        self.output_delay_remaining =
+            Self::calculate_delay_compensation(&self.resampler, self.channels);
+    }
+
+    fn fill_input_buffer(&mut self, needed: usize, num_channels: usize) {
+        while self.input_frame_count < needed {
```
phayes: I think this can read past span edges.  There's no checking in this while loop of span boundaries, so we can read past into a new inner span with different channel counts and sample-rates.

src/conversions/sample_rate/rubato.rs
```diff
@@ -0,0 +1,448 @@
...
+                let mut output_adapter = InterleavedSlice::new_mut(
+                    self.output_buffer.as_mut_slice(),
+                    num_channels,
+                    num_frames,
+                )
+                .ok()?;
+
+                self.resampler
+                    .process_into_buffer(&input_adapter, &mut output_adapter, indexing_ref)
+                    .ok()?
```
phayes: If `tracing` feature is enabled, should this logged as an error before returing `None` to end the stream?

src/conversions/sample_rate/rubato.rs
```diff
@@ -0,0 +1,448 @@
...
+                let input_adapter =
+                    InterleavedSlice::new(&self.input_buffer, num_channels, actual_frames).ok()?;
+
+                let num_frames = self.output_buffer.capacity() / num_channels;
+                let mut output_adapter = InterleavedSlice::new_mut(
+                    self.output_buffer.as_mut_slice(),
+                    num_channels,
+                    num_frames,
+                )
+                .ok()?;
```
phayes: Same thing here, perhaps log this error if `tracing` is enabled?

src/source/resample/mod.rs
```diff
@@ -0,0 +1,784 @@
...
+                resampler.output_len(),
+                resampler.resampler.output_frames_next(),
+            ),
+            #[cfg(feature = "rubato-fft")]
+            ResampleInner::Fft(resampler) => (
+                resampler.input.current_span_len(),
+                resampler.input.sample_rate(),
+                resampler.input.is_exhausted(),
+                resampler.output_has_samples(),
+                resampler.output_len(),
```
phayes: I think perhaps this should be `resampler.output_remaining()`, not `resampler.output_len()`. 

While they might often be the same, when we're zooming through initial delay samples we use `Buffer::skip()` which advances the read marker but doesn't change the buffer length. 

Since we're never emitting those initial dummy samples, I think using `resampler.output_len()` will overreport in this situation.

phayes: Thinking more about this.  This could be a problem even outside of initial sample delay skipping. 

We use this `outout_len` value here in `current_span_len()`:

```rust
// When the ratio contains a fraction, we cannot choose the floor or ceiling
// arbitrarily, because the resampler may produce either based on its internal state
if output_has_samples {
    // Running state: we are iterating over our buffer with resampled samples
    Some(output_len)
} 
```
            
If someone queries our current_span_len() mid-chunk, I think we want the amount *remaining* in the current buffer, not the total buffer size.

phayes:
I have two more general comments on this:

_**Amortization**_

Resampling can be very bursty. A single outer next() does both of these things all at once in a single `next_sample()`:

1. Pull many samples from `self.input.next()` until we have enough to do resampling work.
2. Run one expensive `process_into_buffer(...)` call.

It might make sense to try to amortize this over more calls to `next_sample()`.   There's nothing we can do to amortize  `process_into_buffer()`, but we could try to spread the calls to inner `next()` over many `next_sample()`, which could help when inner `next()` is also occationally expensive, or can't keep up in those large bursty pulls. The idea would be to estimate the number of inner pulls we'll need before the next `process_into_buffer()` and pull those into the input buffer on every `next_sample()` (or every other `next_sample()` depending on ratio etc).  Then `fill_input_buffer()` would only need to pull the balance. 

It's quite possible that I'm way over thinking this, and buffer sizes in CPAL make careful amortization of inner `next()` calls over-engineering. But I wanted to leave the thought here for you to consider. 

_**Resampler Pool**_

I notice that we're allocating a new resampler on span changes. It might make sense to introduce the idea of a "resampler pool" where we can park used resamplers, reset them with `Resampler::reset(&mut self)`, then draw from the pool without allocating if there's an already allocated and compatible resampler available.

Check out https://github.com/phayes/ardftsrc-rs/blob/master/ardftsrc/src/realtime.rs#L615, where I've implemented this for ardfcsrc.  It's not quite the same idea since my architecture around handling span boundaries and buffers is pretty different, but it could be worth thinking about. 

Both of these ideas could be "future extensions" and are definitely niche optimizations, but I thought it worth sharing my thoughts.

src/source/resample/mod.rs
```diff
@@ -0,0 +1,784 @@
...
+            if output_has_samples {
+                // Running state: we are iterating over our buffer with resampled samples
+                Some(output_len)
+            } else if input_exhausted {
+                // End state: we are at the end of our buffer and the source is exhausted
+                Some(0)
+            } else {
+                // Initial state: our buffer is empty until the first call to next() loads it with
+                // resampled samples. Return the size of the next buffer.
+                Some(output_frames_next * self.channels().get() as usize)
```
phayes: I dont think this is correct for very short initial spans.   This assumes we will always process at least one full buffer worth before the span ends.   I think it should probably return `Some(1)` (or `Some(channels)` if assuming frame aligned) to force a recheck on the next tick.

src/source/resample/mod.rs
```diff
@@ -0,0 +1,784 @@
...
+                resampler.input.current_span_len(),
+                resampler.input.sample_rate(),
+                resampler.input.is_exhausted(),
+                resampler.output_has_samples(),
+                resampler.output_len(),
+                resampler.resampler.output_frames_next(),
+            ),
+        };
+
+        let ratio = Ratio::new(self.sample_rate().get(), input_sample_rate.get());
```
phayes: There is pretty limited use of the `num_rational` crate here.  It seems like it might be best to remove it as a dependency and just do something like this:

```rust
// Integer upsampling (2x, 3x, etc.) - always exact and frame-aligned
if output_sample_rate % input_sample_rate as usize == 0 {
    match input_span_len {
       None => None,
       Some(input_len) => {
          let output_len = input_len * output_sample_rate / input_sample_rate as usize
          Some(output_len)
       }
  }
}
```

src/source/resample/rubato.rs
```diff
@@ -0,0 +1,448 @@
...
+        input: I,
+        target_rate: SampleRate,
+        chunk_size: usize,
+        sub_chunks: usize,
+    ) -> Result<Self, ResamplerCreationError> {
+        let source_rate = input.sample_rate();
+        let channels = input.channels();
+
+        // Calculate the GCD-reduced ratio
+        let ratio = Ratio::new(target_rate.get(), source_rate.get());
```
phayes: We could remove `num_rational` as a dependency by just doing something like this:

```rust
let target = target_rate.get();
let source = source_rate.get();
let den = source / gcd(target, source);
```

with a small helper:

```rust
// Greatest Common Denominator
fn gcd(mut a: u32, mut b: u32) -> u32 {
    while b != 0 {
        let r = a % b;
        a = b;
        b = r;
    }
    a
}
```

src/source/resample/mod.rs
roderickvd: > If someone queries our current_span_len() mid-chunk, I think we want the amount _remaining_ in the current buffer, not the total buffer size.

The contract for `current_span_len` is to return the _total_ length of the span (as in: what it says on the tin) irrespective of the position of the cursor.

src/source/resample/mod.rs
roderickvd: That would not catch GCDs for ratios like 44.1:48.

roderickvd:
> Resampling can be very bursty

Indeed, this resampler will be jittery. Spawning a thread with look-ahead could be interesting, but likely non-trivial because it will advance the source (possibly exhausting it, moving across spans, triggering next items in a queue) when the output hasn't caught up yet. Solvable, just not trivial.

> I notice that we're allocating a new resampler on span changes.

This I am less worried about, because any span change will mean discontinuity in the stream anyway. A little latency with a bit of silence shouldn't matter.

src/source/resample/rubato.rs
yara-blue: I see. Is your goal to lower compile time by having fewer dependencies? Since `num_rational` is a tiny crate without any build files. I suspect it to be not add that much. 

I'm happy to go either way here, so I'll leave it up to Roderick.

yara-blue:
> Solvable, just not trivial.

I'm happy to see if we can add a "do work on another thread" mechanism to Rodio. But I'll only have time _after_ mergin in the fixedsource stuff. Is this a release blocker?

roderickvd:
> > Solvable, just not trivial.
> 
> I'm happy to see if we can add a "do work on another thread" mechanism to Rodio. But I'll only have time _after_ mergin in the fixedsource stuff. Is this a release blocker?

I wouldn't see why.

I'll address @phayes' review points (thanks!) and let you know when done.

roderickvd:
@yara-blue if you could take a look at the open questions in the PR body text?

yara-blue:
> Should we move and rename `Resample` from `src/source/resample.rs` to `SampleRateConverter` in `src/conversions/sample_rate.rs` and basically replace it?
> 
>     * This would be breaking due to requiring that `I: Source` instead of `I: Iterator`.
> 
>     * `Resample::new` could live besides `new_poly`, `new_sinc` and the builder.

Yes :)

It is breaking but we do not really have anything implemented on Iterator right? The SampleRateConverter is the odd one out here. Worth it to break that if it makes everything consistent. If users end up having a problem they can always wrap their iterator into a source.

src/conversions/sample_rate/rubato.rs
roderickvd: `fill_input_buffer` doesn't enforce boundaries itself; it's called with an already-capped `fill_target`. The cap is computed in `next_sample` before the call: `needed_input = original_needed.min(input_frame_count + remaining)` where `remaining = span_frames - already_read`. So the buffer never reads past the current span.

src/conversions/sample_rate/rubato.rs
roderickvd: Done.

src/conversions/sample_rate/rubato.rs
roderickvd: Done.

src/source/resample/mod.rs
roderickvd: Good catch! The initial state now returns (one frame, which forces a recheck on the next tick.

src/source/resample/mod.rs
roderickvd: Yes, removed. I noticed that `num_rational` pulled in a couple of other dependencies, so your low-dependency approach is best. The GCD approach from your other comment still enables 44.1/48 ratios.

src/source/resample/mod.rs
roderickvd: Another good one though... the span gift that keeps on giving. I'll introduced `RubatoResample::output_span_len` and snapshot once per chunk from `output_buffer.remaining()` after delay-skip and cap. This gives `current_span_len` a value that correctly excludes delay-skipped leading samples while satisfying the total-span-length contract.

roderickvd:
Had to rebase for the CI to run. I split everything up into separate commits, so you can track back from HEAD.
Pending your final review, this should cover all earlier points.

yara-blue:
rebased the rebased :3

_Finally_ got time, im almost done reviewing. Should go in this weekend!

Update: I'm refactoring as I go (seemed to make more sense then more back and forths). I'll push those changes once I'm done for review.

src/conversions/sample_rate/rubato.rs
yara-blue: Any error returned by InterleavedSlice is a critical bug, we should not report Error's for those but just crash, there is no way to recover anyway. I've swapped these for `expect` calls. Let me know if you disagree.

src/conversions/sample_rate/mod.rs
```diff
@@ -0,0 +1,494 @@
...
+                let (lower, upper) = resampler.input.size_hint();
+                let lower = adjusted_for_resampling(lower);
+                let upper = upper.map(adjusted_for_resampling);
+                (lower.raw(), upper.as_ref().map(OutSamples::raw))
+            }
+            #[cfg(feature = "rubato-fft")]
+            ResampleInner::Fft(resampler) => {
+                let input_hint = resampler.input.size_hint();
+                let buffered_remaining = resampler.output_remaining();
+                (input_hint, resampler.source_rate, buffered_remaining)
```
yara-blue: I don't think this is tested in CI since the test would have failed on this.
(returning a three-tuple where there used to be a two tuple)

yara-blue: nvmd that is on me, I forgot what I did and did not refactor...

yara-blue (approved):

yara-blue:
yay we're good now.

Okay so what changed:
- There is a lot of book keeping regarding the resampler's state and progress. I added a bunch of newtypes to lower the chance of anything switching up there.
- Reworked the structure of the rubato resampler
- Extracted some stuff to its own modules
- Increased the MSRV to 1.95 (released april, so we cannot release for a few months because of this _or_ we change MSRV policy). Why? to get `cold_path()` as an optimization. Let's re-evaluate at release?

@roderickvd you wanna review the changes to your PR or are we good to merge?

roderickvd:
The test suite is still there right? If so and everything passes: go right ahead. I’m on mobile and the suite was pretty stringent.

PR #844, opened 2026-02-11
Filter out zero duration

cetra3:
This adjusts the symphonia integration to filter out erroneous total lengths if the duration is set to zero when probing.

I found that with a few audio tracks I wanted to play they don't set the track duration in the metadata very well for some reason & so I have to rely on reading the track to get out the total duration.

Unfortunately this breaks the `seek` behaviour of the rodio sink, since that will clamp to the total duration.

### DIFF
--- a/src/decoder/symphonia.rs
+++ b/src/decoder/symphonia.rs
@@ -104,11 +104,12 @@ impl SymphoniaDecoder {
 
         let mut decoder = symphonia::default::get_codecs()
             .make(&track.codec_params, &DecoderOptions::default())?;
-        let total_duration = stream
+        let total_duration: Option<Duration> = track
             .codec_params
             .time_base
             .zip(stream.codec_params.n_frames)
-            .map(|(base, spans)| base.calc_time(spans).into());
+            .map(|(base, spans)| base.calc_time(spans).into())
+            .filter(|d: &Duration| !d.is_zero());
 
         let decoded = loop {
             let current_span = match probed.format.next_packet() {


roderickvd:
Good idea, thanks.

roderickvd:
By the way is this with an AAC track? For kicks this is what I have in #786:

```rust
        let total_duration: Option<Duration> = track
            .codec_params
            .time_base
            .zip(track.codec_params.n_frames)
            .and_then(|(base, spans)| {
                // Treat n_frames=0 as invalid for AAC codec (common issue with some M4A files)
                // TODO: remove when Symphonia > 0.5.4 is released, see: https://github.com/pdeljanov/Symphonia/commit/4f41954de75e2447753f3e522cc7345ec414a703#diff-d4e7073755222d37e1b948b48dd361bea2066dfcfe7f7069271ebf75b6a7e083
                if spans == 0 && track.codec_params.codec == CODEC_TYPE_AAC {
                    None
                } else {
                    Some(base.calc_time(spans).into())
                }
            });
```

Symphonia v0.5.5 still does that, by the way.

cetra3:
Yeah it is an AAC track.  I was testing out [my new audio player](https://github.com/cetra3/slix-player), and noticed that the files I am creating ([via gstreamer](https://github.com/cetra3/rustcloud)) have pretty bogus metadata.  I [analyze the track](https://github.com/cetra3/slix-player/blob/05441017fe087bf255e6921e43467f5d74c61bde/src/audio.rs#L225) anyway to get the actual length out, but when trying to seek it was always starting from the beginning...

Issue #782, opened 2025-08-20
OutputStream: expose the underlying cpal::Stream or provide wrapper for pause()/play()


martinetd:
Hello,

thanks for rodio!

I've been playing with rodio in https://github.com/martinetd/fuurin and one thing that's been annoying me is that even with all sources paused, just having an OutputStream present consumes a non-negligible amount of CPU sending silence to alsa.
(for what it's worth I tried the pipewire output provided in https://github.com/RustAudio/cpal/pull/938 and it's very slightly better, but sensibly the same)

With the current code, I see no other way around dropping the whole output stream which means dropping all the sinks I've opened first then the mixer and the stream itself, but I tried just exposing the `_stream` and using `outputstream.stream().pause()` and it reduces cpu to zero, so I think it's a much better approach to this problem.

Would you accept a patch such as this?
```diff
diff --git a/src/stream.rs b/src/stream.rs
index f6c8bcf550c0..2df5d7307b49 100644
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -44,7 +44,7 @@ pub struct OutputStream {
     config: OutputStreamConfig,
     mixer: Mixer,
     log_on_drop: bool,
-    _stream: cpal::Stream,
+    stream: cpal::Stream,
 }
 
 impl OutputStream {
@@ -53,6 +53,11 @@ pub fn mixer(&self) -> &Mixer {
         &self.mixer
     }
 
+    /// Access the output stream's underlying cpal Stream.
+    pub fn stream(&self) -> &cpal::Stream {
+        &self.stream
+    }
+
     /// Access the output stream's config.
     pub fn config(&self) -> &OutputStreamConfig {
         &self.config
@@ -487,7 +492,7 @@ fn open<E>(
         Self::init_stream(device, config, source, error_callback).and_then(|stream| {
             stream.play().map_err(StreamError::PlayStreamError)?;
             Ok(Self {
-                _stream: stream,
+                stream: stream,
                 mixer: controller,
                 config: *config,
                 log_on_drop: true,
```

Or would it be better to expose pause() and play() method directly that'd just call `self.stream.pause()`/`play()` ?

I think allowing access to the stream is more flexible ultimately, but happy either way.

I'll open a PR once there's an agreement on which is better :)

Thanks!

yara-blue:
> just having an OutputStream present consumes a non-negligible amount of CPU sending silence to alsa.

[...]

Issue #299, opened 2020-07-10
Is there a way to stop the background thread after playing a sound?


harrisonthorne:
After using `play_once` or `play_raw` on a sound and audio device (both local variables), the playback stream still shows up in `pavucontrol`:

![pavucontrol](https://user-images.githubusercontent.com/33523827/87115901-230a7100-c23a-11ea-8074-506d1a00ad3f.png)

The thread also still exists when I inspect htop, and the thread is using around 3% CPU:

![20200709-231723](https://user-images.githubusercontent.com/33523827/87116127-9f9d4f80-c23a-11ea-8ad7-62e46585812c.png)

Code I'm using currently:

```rust
fn play_new_hour_sound() {
    if let Some(audio_device) = rodio::default_output_device() {
        let cursor = Cursor::new(include_bytes!("../new_hour.wav").as_ref());
        if let Ok(sink) = rodio::play_once(&audio_device, cursor) {
            sink.sleep_until_end()
        };
    }
}
```

I would expect the thread to stop after `audio_device` and `sink` are dropped (or `audio_device` and `source` if using `play_raw`), but the thread still runs. Is there a way to stop rodio's background thread after playing a sound?

mdekstrand:
I have been investigating high CPU usage on low-power devices (e.g. Raspberry Pi Zero W), over in librespot-org/librespot#567; I suspect it may be this problem (or a related problem). At idle, Rodio consumes 12-13% of a 1GHz ARMv6 core.

asafg6:
I saw 100% cpu usage on my machine when it was loaded but idle.
Not the best solution but works for me in the mean time:
`
    pub fn sleep_and_unload(&mut self) {
        if let Some(sink) = &self.sink {
            sink.sleep_until_end();
            self.sink = None;
            self.stream_handle = None;
            self._stream = None;
        }
    }
`
Drops CPU back to 0

PR #847, opened 2026-02-18
expose pause/play in MixerDeviceSink

sofoxe1:
this PR partially addresses long standing issue of the background cpu usage even when nothing is playing
https://github.com/RustAudio/rodio/issues/299
https://github.com/RustAudio/rodio/issues/782

while there was an attempt to make a proper *fix* https://github.com/RustAudio/rodio/pull/791 it looks like the author gave up after a day

### DIFF
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -54,13 +54,19 @@ impl MixerDeviceSink {
     pub fn mixer(&self) -> &Mixer {
         &self.mixer
     }
-    /// pause underlaying audio stream
+    /// pause underlaying audio stream reducing cpu usage
     ///
-    /// this will prevent audio from playing until play() is called
+    /// will silenty fail to play audio until play() is called
+    ///
+    /// This is an experimental API it is likely to removed once a better solution is found without notice or a change log entry
+    #[cfg(feature = "experimental")]
     pub fn pause(&self) {
         let _ = self._stream.pause();
     }
-    /// resumes underlaying audio stream
+    /// resume underlaying audio stream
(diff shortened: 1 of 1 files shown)


yara-blue:
> it looks like the author gave up after a day

it's on a list somewhere I swear!

> this PR partially addresses long standing issue of the background cpu usage even when nothing is playing

My worry is that if we merge something like this as is we will break a lot of systems when we eventually add a proper fix (which we will however span-less Rodio is taking priority the coming months).

We can mitigate that by putting it under the `experimental` flag and clearly documenting this method may disappear once a better solution is implemented. Would you be open to making those changes?

src/stream.rs
```diff
@@ -54,6 +54,22 @@ impl MixerDeviceSink {
     pub fn mixer(&self) -> &Mixer {
         &self.mixer
     }
+    /// pause underlaying audio stream reducing cpu usage
+    ///
+    /// will silenty fail to play audio until play() is called
+    ///
+    /// [THIS WILL BE REPLACED IN THE FEATURE IN FAVOR OF AUTOMATIC PAUSING](https://github.com/RustAudio/rodio/pull/791)
```
yara-blue: I'd would not mention only the git issue (maybe we find another solution) instead something along the lines of: 

`This is an experimental API it is likely to removed once a better solution is found without notice or a change log entry`

^^ which is also why we do not need a change-log entry

yara-blue:
Thank you! Sorry it took a while to get it merged

PR #848, opened 2026-02-21
docs: fix information about default decoder features

aschey:
Fixes the docs to reflect the current state after the default decoder was switched to Symphonia in #753. I've also added a bit more info about licensing since that's the main reason people would likely opt for the alternative decoders.


yara-blue:
Ty! :)

PR #850, opened 2026-02-22
ci: verify docs build and resolve warnings

aschey:
Adds a step to CI to verify docs build correctly and resolves existing warnings related to invalid links (required so that we can include `-D warnings` in CI).


yara-blue:
Yay :)

PR #806, opened 2025-10-08
Symphonia custom registry support

royalmustard:
This PR add support for using a custom codec registry with the symphonia backend.

This allows easily add support for additional codecs, many of which are already supported by third party addons for symphonia (e.g. opus).

To use a custom registry, a (global) codec registry must be instantiated and all desired codecs must be registered in it.
Then, this registry can be given to a `DecoderBuilder` via the `with_codec_registry` function.
If no custom registry is given to the builder, the decoder will use symphonia's default registry, as before.

PR #851, opened 2026-02-22
feat: opus support via libopus and allow registering additional codecs

### LINKED ISSUE #38: Add an Opus decoder

aschey:
Supersedes #806 

Resolves #38 

Thanks @royalmustard for starting on this! I took the changes from the previous PR and made a few tweaks:

- Exposes a method to add a decoder to the existing registry instead. This is more difficult to misuse because it's easy to forget to re-enable the default codecs when creating a completely new registry.
- Adds opus support via the `symphonia-libopus` flag. Opus support is a common enough request that I think it makes sense to support this via a flag rather than require users to add the codec manually.
- Bumps the MSRV so CI can pass ([discussed here](https://github.com/RustAudio/rodio/pull/806#issuecomment-3904944309)).
- Adds a different third-party codec example using [symphonia-adapter-fdk-aac](https://docs.rs/symphonia-adapter-fdk-aac/latest/symphonia_adapter_fdk_aac) (FDK supports HE-AAC which is currently not supported by Symphonia's AAC implementation).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ repository = "https://github.com/RustAudio/rodio"
 documentation = "https://docs.rs/rodio"
 exclude = ["assets/**", "tests/**"]
 edition = "2021"
-rust-version = "1.87"
+rust-version = "1.89"
 
 [features]
 # Default feature set provides audio playback and common format support
@@ -49,6 +49,11 @@ noise = ["rand", "rand_distr"]
 #
 # Enable WebAssembly support for web browsers
 wasm-bindgen = ["cpal/wasm-bindgen"]
+
+# Base symphonia feature (doesn't enable any codecs)
+# A feature with this name is created implicitly due to other features that reference `symphonia/xxx`
+# but we need to make it explicit here so other features can reference it.
+symphonia = ["dep:symphonia"]
 
 # To decode an audio source with Rodio, you need to enable the appropriate features for *both* the
 # demuxer and the decoder.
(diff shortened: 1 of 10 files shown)


Cargo.toml
```diff
@@ -50,6 +50,9 @@ noise = ["rand", "rand_distr"]
 # Enable WebAssembly support for web browsers
 wasm-bindgen = ["cpal/wasm-bindgen"]
 
+# Base symphonia feature (doesn't enable any codecs)
+symphonia = ["dep:symphonia"]
```
aschey, line 56: Note that there are a lot of places that use `#[cfg(feature = "symphonia")]` even though there was no feature named `symphonia`. A feature with this name was created implicitly due to the other features that reference `symphonia/xxx`. For `symphonia-libopus`, I wanted to make sure the base symphonia feature was also enabled, so we need to make this feature explicit now.

yara-blue: That makes sense! Thanks for that explanation! I think the implicit behavior is unexpected enough that copying what you said as a comment into Cargo.toml is worth it.

src/decoder/builder.rs
```diff
@@ -231,6 +250,19 @@ impl<R: Read + Seek + Send + Sync + 'static> DecoderBuilder<R> {
         self
     }
 
+    /// Adds a third-party decoder.
+    /// If multiple decoders for the same codec are registered, the one registered last will take precedence.
+    ///
+    /// See the documentation for [CodecRegistry] for more info.
+    #[cfg(feature = "symphonia")]
+    pub fn with_decoder<D>(self) -> Self
```
yara-blue: Since this only works for symphonia decoders this should maybe be named `with_symphonia_decoder`. On the other hand I see a near future where we drop all the other decoders (I have not seen any issue with the Symphonia decoders and they have been the default for a while). If we do that the name is fine.

yara-blue:
Looks great to me! There is a trivial conflict, I think your explanation about features deserves to live in the toml, and we might wanna think about the method name for a sec. As soon as that's done LGTM.

PS: apologies for the somewhat slow review. Thanks for picking this up and keeping the ball rolling.

PPS: it's soo great to see these old issues closed :)

Cargo.toml
aschey: Sure thing, added.

src/decoder/builder.rs
aschey: I agree, that makes sense. Updated :+1:

yara-blue:
Yay :tada:

Issue #854, opened 2026-02-26
Mixer mishandles sources that are added out of phase


will3942:
If a source is added whilst the mixer is outputting on a non-zero channel the source will be added to the `still_pending` `Vec` but there is no process for dequeuing from this `Vec` and adding to the `current_sources` Vec.

We run a custom mixer so haven't ran into this issue ourselves but when migrating the performance improvements from the upstream Mixer in 0.22 we noticed this oddity.

The fix should be straightforward for `start_pending_sources` to check `still_pending` and the phase alignment on the next iteration but it would be worth adding a regression test here too to catch it in the future.

Code snippet: https://github.com/RustAudio/rodio/blob/master/src/mixer.rs#L145-L157

roderickvd:
That’s an interesting catch. @yara-blue this would not be a problem if we enforce that sources emit frame-aligned samples, see #831. Then we could remove this code altogether, unless I’m mistaken.

yara-blue:
@roderickvd its a slightly different issue I think. The fix (only add sources when we are playing channel0) was trivial and simplified the mixer. See PR #855

The mixer does not account for not frame alignment sources being added.

yara-blue:
@roderickvd I feel like this is pretty bad, we should release a 0.22.2 for this right?

Also thanks @will3942 for the report!!

roderickvd:
Indeed!

PR #855, opened 2026-02-26
fix-mixer-start-pending-source

yara-blue:
fixes https://github.com/RustAudio/rodio/issues/854


yara-blue:
mhm, maybe I should run tests before pushing...
edit: fixed

Issue #853, opened 2026-02-26
get_pos much further ahead than actual playback position


SofusA:
I recently updated my music player to rodio 0.22 from 0.21.
No issues on my Fedora 43 desktop.

But on my raspberry pi with updated raspian, audio is not playing and position jumps much faster than 1 sec/sec
I have only tested with .flac files.

I get no errors or warnings when loading files or initiating playback.

Downgrading to 0.21 fixes the issue.

I have attached a video of my player skipping in time, which is the only visual cue i have.
No audio is playing while this is happening. 

There are no warning or logs to show.

https://github.com/user-attachments/assets/a6b72aca-011d-428f-bcd9-31fe2cb0f3e6

yara-blue:
Very strange indeed. The PI is weaker so I'd expect maybe audio artifacting but not skipping. I assume you're progress bar is based on the `TrackPosition` `Source`?

The fastest way to figure out what is going on is probably a `git bisect`. Let me know if you need help with that. Every Rodio build _should_ work. You could try starting at 744f27b27816aac74f266dc0cb3c5ffb6226c630 after that we did some pref. work maybe that did something.

roderickvd:
Shouldn’t the title of the issue be different: audio playback isn’t the problem, but position reporting is?

Is it different between your Fedora desktop (which backend - ALSA direct?) and your Raspberry Pi (PulseAudio?)

yara-blue:
> Shouldn’t the title of the issue be different: audio playback isn’t the problem, but position reporting is?

Good point!

@SofusA before trying `git bisect` you can see what happens if you make you're audio chain simpler unless you are using the `Player` then there is not a lot we can try.

SofusA:
`744f27b27816aac74f266dc0cb3c5ffb6226c630`: does not work
`809ad5d5bd97485ef5ac2736ffddde0c03be881b` does work

So it must be related to `cpal`.

I am using `Player` on 0.22 and `Sink` on `0.21`

roderickvd:
OK that’s a very useful data point. 

Does it do this both on your Fedora desktop and your Raspberry Pi? If it’s different please tell me about the differences in configuration.

SofusA:
Yes there are no problems on my fedora 43 desktop. 

I am new to audio. What do you mean with configuration? 

My desktop is a very default fedora setup, and audio is playing from a usb dac with pipewire. 

[...]

PR #856, opened 2026-02-26
limit-buffersize

yara-blue:
fixes https://github.com/RustAudio/rodio/issues/853

huge cpal default buffer sizes (due to weirdly set up systems) are a footgun. 

This makes `from_device` replace `BufferSize::Default` with a fixed 4096 samples. This affects `from_default_device` & `open_default_sink`.

### DIFF
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -191,10 +191,9 @@ impl DeviceSinkBuilder {
             .with_device(device)
             .with_supported_config(&default_config);
 
-        // 50 ms of audio
-        let safe_buffer_size =
-            device.config.sample_rate().get() * device.config.channel_count().get() as u32 / 20;
-        let safe_buffer_size = safe_buffer_size.next_power_of_two();
+        // minimum 40ms of audio
+        let sample_rate = device.config.sample_rate().get();
+        let safe_buffer_size = (sample_rate / (1000 / 40)).next_power_of_two();
 
         // This is suboptimal, the builder might still change the sample rate or
         // channel count which would throw the buffer size off. We have fixed


roderickvd:
I suggest to use the sample rate and channel count to calculate an audio latency of ~100 ms. Almost everywhere there's a double-buffering scheme, so that means that one buffer should represent ~50 ms.

For bonus points the buffer sizes are rounded to a safe and performant power of two.

yara-blue:
> I suggest to use the sample rate to calculate an audio latency of ~100 ms. Almost everywhere there's a double-buffering scheme, so that means that one buffer should represent ~50 ms.
> 
> For bonus points the buffer sizes are rounded to a safe and performant power of two.

yes that makes way more sense :) I'll do that and merge. Shall we put out a new patch release after that?

roderickvd:
Yeah. 

Edit: sample rate _and channel count_

yara-blue:
overhauled the experimental new device opening API (working title: speakers), it now does this nicely. The change to the stable API is just 17 lines.

yara-blue:
opportunity for a really nice refactor appeared. I'm looking into that

src/stream.rs
```diff
@@ -187,9 +187,24 @@ impl DeviceSinkBuilder {
...
+        let mut device = Self::default()
             .with_device(device)
-            .with_supported_config(&default_config))
+            .with_supported_config(&default_config);
+
+        // 50 ms of audio
+        let safe_buffer_size = device.config.sample_rate().get() as u32
+            * device.config.channel_count().get() as u32
+            / 20;
+        let safe_buffer_size = safe_buffer_size.next_power_of_two();
```
roderickvd: I recommend a round-half-up kind of scheme. For a typical 44.1 kHz stereo stream, you get 4410 samples. Next power of two is 8192 is ~93 ms for one period and a latency of ~186 ms gets into the area of perceptible asynchrony.  4096 is then closer to the target latency.

yara-blue: Agreed 8192 is too long. Not sure what you mean with `round-half-up kind of scheme` I went with using 40ms as base that should take care of the 44100 SR ballooning into high latency. Let me know if you have a better idea.

yara-blue:
Merging this so I can easily see what it does to #853 feel free to discuss/review further here if you want too

src/stream.rs
roderickvd: I had this in mind:

```rust
fn nearest_pow2(n: u32) -> u32 {
    if n <= 1 { return 1; }
    let next = n.next_power_of_two();
    let prev = next >> 1;
    if n - prev <= next - n { prev } else { next }
}
```

roderickvd:
We probably should document that Rodio configures a default buffer with a latency near 100 ms:
- regardless of system defaults
- unless a buffer size is specified manually

Before cpal changed this behavior, users found it surprising that their system settings were ignored.

src/stream.rs
yara-blue: adressed in 338d0d81

yara-blue:
> We probably should document that Rodio configures a default buffer with a latency near 100 ms:
> 
>     * regardless of system defaults
> 
>     * unless a buffer size is specified manually
> 
> 
> Before cpal changed this behavior, users found it surprising that their system settings were ignored.

done in 1cd15f8cdcea

PR #858, opened 2026-02-27
fix: try_seek on SkipDuration not relative from skipped duration

### LINKED ISSUE #857: try_seek on SkipDuration not relative from skipped duration
In skip_duration source, total_duration returns duration subtracted by skipped duration.
But try_seek attempt to seek duration from original start of internal source.

I think seek position need to be offseted by skipped duration.

Keinsleif:
Fixes #857


yara-blue:
makes sense :+1:

PR #860, opened 2026-02-28
[update] update remaining_duration after success try_seek

### LINKED ISSUE #859: try_seek on TakeDuration causes shift of truncated duration
I think that `TakeDuration` should behave as if the source has ended at a certain time.

However, in `TakeDuration` source, when `try_seek` is called, the `remaining_duration` is not updated even if the playback position changes forward or backward, so the end position shifts forward or backward by the amount of seeking.

`remaining_duration` should be updated when `try_seek` is called, using a method similar to that used by the `TrackPosition` source.

Keinsleif:
I don't really understand the `total_duration` calculation in `TakeDuration`, so there may be something odd in the code.

Fixes #859

### DIFF
--- a/src/source/take.rs
+++ b/src/source/take.rs
@@ -176,7 +176,11 @@ where
             // This assumes that the seek implementation of the codec always
             // starts again at the beginning of a span. Which is the case with
             // symphonia.
-            self.remaining_duration = self.total_duration().unwrap_or(self.requested_duration).checked_sub(pos).unwrap_or(Duration::ZERO);
+            self.remaining_duration = self
+                .total_duration()
+                .unwrap_or(self.requested_duration)
+                .checked_sub(pos)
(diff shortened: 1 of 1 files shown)


yara-blue (approved):
Seems good to me! :tada:

src/source/take.rs
```diff
@@ -171,6 +171,18 @@ where
 
     #[inline]
     fn try_seek(&mut self, pos: Duration) -> Result<(), SeekError> {
-        self.input.try_seek(pos)
+        let result = self.input.try_seek(pos);
+        if result.is_ok() {
+            // This assumes that the seek implementation of the codec always
+            // starts again at the beginning of a span. Which is the case with
```
roderickvd, line 177: Actually not so with accurate seeking that can advance to a more exact position within a span.

PR #863, opened 2026-03-08
feat: infinite sources return Duration::MAX for total_duration

### LINKED ISSUE #702: Idea/proposal: Total_duration enum (unknown/max/guarenteed)
Idea by @The0x539 https://github.com/RustAudio/rodio/issues/626#issuecomment-2661094585. Currently infinite sources will return `None` for `total_duration`, when paired with a `TakeDuration` you still get `None`. Another place we can do better is `skippable`. Right now a `skippable` will pass on the `total_duration` of what it wraps. Because the source can be skipped at any point there is no guarentee it will actually play for that length.

To address these issues we can put extra information in the `total_duration` result:

```rust
enum SourceDuration {
  Max(Duration),
  Guarenteed(Duration),
  Infinite,
  Unknown,
}
```

We could provide methods to turn this into an option returning Duration::MAX for Infinite. We could call that method `known`. We could also provide a `map`, `is_known` and maybe an `unwrap_or`.

A far simpler alternative to this proposal would be making infinite sources return `Duration::MAX`.

Sim-hu:
## Summary
- Infinite sources (sine, square, triangle, sawtooth, signal generator, noise, repeat, looped decoder, microphone) now return `Some(Duration::MAX)` instead of `None` from `total_duration()`
- `TakeDuration` now returns `Some(requested_duration)` when wrapping unknown-duration sources, instead of `None`
- Updated `Source::total_duration()` doc to clarify that `None` means unknown and `Duration::MAX` means infinite

This is the simpler alternative discussed in #702 — using `Duration::MAX` to represent infinity rather than introducing a new enum.

## Test plan
- [x] Added `infinite_sources_return_duration_max` test
- [x] Added `take_duration_on_infinite_source` test — verifies `SineWave.take_duration(5s)` reports `Some(5s)`
- [x] Added `repeat_returns_duration_max` test
- [x] All existing tests pass (`cargo test --no-default-features --features "wav,mp3,flac,vorbis" --lib --tests`)
- [x] `cargo clippy` clean
- [x] `cargo fmt --check` clean

Fixes #702

### DIFF
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -727,12 +727,11 @@ where
 
     /// Returns the total duration of this audio source.
     ///
-    /// Always returns `None` for looped decoders since they have no fixed end point -
-    /// they will continue playing indefinitely by seeking back to the start when reaching
-    /// the end of the audio data.
+    /// Always returns `Duration::MAX` for looped decoders since they play
+    /// indefinitely by seeking back to the start when reaching the end.
     #[inline]
     fn total_duration(&self) -> Option<Duration> {
-        None
+        Some(Duration::MAX)
     }
 
     /// Attempts to seek to a specific position in the audio stream.
--- a/src/microphone.rs
+++ b/src/microphone.rs
@@ -204,7 +204,7 @@ impl Source for Microphone {
     }
 
     fn total_duration(&self) -> Option<std::time::Duration> {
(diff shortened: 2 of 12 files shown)


yara-blue:
closed as spam per https://rust.audio/community/ai/

PR #874, opened 2026-04-16
Simplify advance method on SpanTracker

Keinsleif:
Similar code appear on some sources.  
```rust
let current_sample_rate = self.input.sample_rate();
let current_channels = self.input.channels();
let input_span_len = self.input.current_span_len();
```  
I think this is not clean.

I used code similar to the `seek` method to fix this.  
Now, `advance` method receive reference for Source.  

Also, channel count and sample rate are not used on loops that are not boundaries.  
In this PR, SpanTracker retrieves them from input source only on boundaries.


Keinsleif:
I'm not very familiar with how SpanTracker works, but is this working correctly?

yara-blue (approved):
Yes this will work nicely :)

Lovely cleanup thank you for contributing it!

PR #875, opened 2026-04-19
chore(cpal): update to unified error type

roderickvd:
cpal v0.18 is introducing a unified error type. This PR updates Rodio to implement it.


yara-blue:
Thanks!

PR #878, opened 2026-04-27
refactor: add DEFAULT_SAMPLE_RATE and default to 48 kHz

roderickvd:
In cpal I've changed the default sample rate to 48 kHz for devices that support it, falling back to 44.1 kHz, then to the device's maximum support sample rate.

In Rodio we used 44.1 and 48 kHz inconsistently, so I think it's good to follow cpal here.

Second, I changed `stream::supported_output_configs` which had several smells:

- Somewhere in the transition from `SampleRate` becoming `NonZero` this function became public. I'm not sure if that was intentional or smart, because it's conflating `cpal::supported_output_configs`. I reverted it to be private now, and updated the Rustdoc of `with_supported_config` to use the cpal version instead.

- The `supported_output_configs` Rustdoc was imprecise: it did not yield _all_ supported formats, only `[max, 44_100, min]` (in that order). If `min` or `max` happened to also be `44_100`, it had duplicate values. I changed it to yield `[48_000, 44_100, max]` without duplicates.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - All sources now implement `ExactSizeIterator` when their inner source does.
 - All sources now implement `Iterator::size_hint()`.
 - `Chirp` now implements `try_seek`.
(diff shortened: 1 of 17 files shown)


yara-blue (approved):
Perfect, thanks!

PR #885, opened 2026-05-22
Use realtime sanitizer to check against malloc and similar

inkreasing:
This adds tests that use the llvm realtime sanitizer. It checks known blocking libc calls. If you do a endless loop, it won't catch it.

I use the supertrait_item_shadowing thing to overwrite the Iterator::next method, so that it gets sanitized. This should be correct, as that is all audio code. That way it also applies to user written code. It leads to errors if both Source and Iterator are imported, but this is only for tests.

This is currently broken because
1. doesn't work on nix (need help fixing this, i don't use nix)
2. currently the library isn't realtime safe, so either this gets fixed or allows need to be applied (this should probably be done by running the sanitizer in continue mode, so that multiple issues can be found in one run)
3. there should be more tests (i don't know whats useful, i just copied one example)

It could/should get extended by
1. allowing users to easily test their own stuff
2. maybe also test the examples with this

also still need to figure out if it is better to do it with the cargo feature (like it is currently) or use a env-var with a tiny build script. This has the big advantage, that user crates don't have to pass the feature upwards and can just set the env var.

@yara-blue

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -34,10 +34,54 @@ dependencies = [
 ]
 
 [[package]]
+name = "anstream"
+version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "824a212faf96e9acacdbd09febd34438f8f711fb84e09a8916013cd7815ca28d"
+dependencies = [
+ "anstyle",
+ "anstyle-parse",
+ "anstyle-query",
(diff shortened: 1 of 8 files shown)


roderickvd:
This is cool. We’ve got ASAN and TSAN set up in cpal too. Using it as a pipe cleaner for realtime allocations is useful too.

PR #886, opened 2026-05-23
Fix spatial sound inversal

### LINKED ISSUE #884: Spatial sound calculates diff incorrectly (swaps left and right)
In these formulas there is I think mistake that reverses lift/right:

https://github.com/RustAudio/rodio/blame/a352fb53846b47523d828b276b6d625f251aabb2/src/source/spatial.rs#L60-L62

```rust
        let left_diff_modifier = (((left_dist - right_dist) / max_diff + 1.0) / 4.0 + 0.5).min(1.0);
        let right_diff_modifier =
            (((right_dist - left_dist) / max_diff + 1.0) / 4.0 + 0.5).min(1.0);
```

It basically says that if "left_dist" (distance from emitter to "left" listener) is bigger - then modifier to left is bigger (farther - louder, which should be the other way around, i.e closer - louder).
And same for the right.

Essentially at close distances this results in swap in sound - left feels like right and vice versa..

I think it should be 
```rust
        let left_diff_modifier = (((right_dist - left_dist) / max_diff + 1.0) / 4.0 + 0.5).min(1.0);
        let right_diff_modifier =
            (((left_dist - right_dist) / max_diff + 1.0) / 4.0 + 0.5).min(1.0);
```

It sort of appears to be obvious mistake in that code, am I missing something here?

strowk:
Corrects the way sound is produced by Spatial when emitter is located close enough to sources to be noticeably closer to left or right ear.

Previous implementation sounded inversed as demonstrated in tests (right ear would sound louder than left one when emitter is farther from right ear and closer to left ear, and vice versa).

This is a breaking change for anyone relied on old implementation behavior if they decided to switch left and right locations to get correct sounds.

Fixes #884

### DIFF
--- a/src/source/channel_volume.rs
+++ b/src/source/channel_volume.rs
@@ -61,7 +61,7 @@ where
     }
 
     /// Gets the volume for a given channel number. Returns `None` if channel number is invalid.
-    pub fn get_volume(&self, channel: usize) -> Option<Float> {
+    pub fn volume(&self, channel: usize) -> Option<Float> {
         self.channel_volumes.get(channel).copied()
     }
 }
--- a/src/source/channel_volume.rs
+++ b/src/source/channel_volume.rs
@@ -59,6 +59,11 @@ where
     pub fn into_inner(self) -> I {
         self.input
     }
+
+    /// Gets the volume for a given channel number. Will panic if channel number is invalid.
+    pub fn get_volume(&self, channel: usize) -> Float {
+        self.channel_volumes[channel]
+    }
 }
 
 impl<I> Iterator for ChannelVolume<I>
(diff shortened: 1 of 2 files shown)


src/source/channel_volume.rs
```diff
@@ -59,6 +59,11 @@ where
     pub fn into_inner(self) -> I {
         self.input
     }
+
+    /// Gets the volume for a given channel number. Returns `None` if channel number is invalid.
+    pub fn volume(&self, channel: usize) -> Option<&Float> {
```
yara-blue, line 64: small nit: `Float` is copy. But don't worry ill fix it 👍

yara-blue (approved):
Thank you _very_ much! Those tests are so nice and readable :)

PR #887, opened 2026-05-27
chore: update cpal to v0.18 release candidate

roderickvd:
cpal has reached its v0.18 design goals, so consider this a release candidate for final testing in Rodio. I am merging this as soon as the CI passes.

@yara-blue I also went ahead to rename the `symphonia-simd` feature to just `simd` - who knows, in the future we could have SIMD support elsewhere too. I would actually propose to lift this feature to the default features, but that's out of scope for this PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: Install linux build requirements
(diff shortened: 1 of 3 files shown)


yara-blue:
thank you!

> @yara-blue I also went ahead to rename the symphonia-simd feature to just simd - who knows, in the future we could have SIMD support elsewhere too. I would actually propose to lift this feature to the default features, but that's out of scope for this PR.

agreed

PR #888, opened 2026-05-30
chore: update dependencies and improve feature documentation

### LINKED ISSUE #883: Enable cpal's pulseaudio feature by default
Going to start this with a bit of a tangent, please bear with me:

The other day I updated my phone's Kernel and ran into some audio issues, where rodio's audio was playing extremely silently, ignoring output volume level, and not showing up in my audio playback sources. I figure this is probably something wrong with my setup, but a simple `aplay` seems to work out just fine. Long story short: I ran cpal's beep example and realized I get an error by default, but it works just fine if I enable the `pipewire` or `pulseaudio` features.

Now it doesn't really matter why the alsa path isn't working, but on Linux generally pulseaudio is the standard audio playback path. It's most likely to provide a consistent and pleasant audio playback solution. What I do right now is depend on the latest `rodio` and `cpal` git, then just specify the pulse feature to make sure it's enabled automatically for rodio too. But this took some tinkering to figure out and get working.

Pulseaudio (and to a lesser degree pipewire) is the standard audio playback system for Linux, so would it make sense to add a feature for one (or both) of these and enable it by default?

This is mostly a suggestion, you might have some opinions on this already, I figured since this required upstream work in cpal it might have just fallen off the radar.

roderickvd:
- Enable `pulseaudio` and `simd` by default
- Disable `dither` by default
- Improve docs

Closes #883

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -138,31 +138,48 @@
 //! to avoid adding extra crates to your binary.
 //! See the [available feature flags](https://docs.rs/crate/rodio/latest/features) for all options.
 //!
-//! ## Optional Features
+//! ## Feature Flags
 //!
-//! Rodio provides several optional features that are guarded with feature gates.
+//! ### Core
 //!
-//! ### Feature "tracing"
+//! | Feature | Description |
+//! |---------|-------------|
+//! | `playback` **(default)** | Audio playback |
+//! | `recording` **(default)** | Microphone input |
+//! | `wav_output` | Write decoded audio to a WAV file |
+//! | `tracing` | Record stream errors as `tracing` events instead of printing to stderr |
+//! | `experimental` | Experimental APIs backed by atomic floating-point operations |
 //!
-//! The "tracing" feature replaces the print to stderr when a stream error happens with a
-//! recording an error event with tracing.
+//! ### Performance
 //!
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -138,31 +138,48 @@
...
 //!
-//! The "tracing" feature replaces the print to stderr when a stream error happens with a
-//! recording an error event with tracing.
+//! ### Performance
 //!
-//! ### Feature "noise"
+//! | Feature | Description |
+//! |---------|-------------|
+//! | `simd` **(default)** | SIMD-accelerated decoding |
+//! | `64bit` | Use `f64` instead of `f32` for all samples and internal math |
```
yara-blue: I liked the previous f64 description. It's indeed to long for a table and the table format is better! Could we maybe get it in as a footnote?

yara-blue (changes requested):
Minor suggestion regarding f64 description

And thanks for doing this!

src/lib.rs
roderickvd: Done in [859c009](https://github.com/RustAudio/rodio/pull/888/commits/859c009fb5e840bcea88c9b20d051413afb5e836).

yara-blue (approved):

PR #893, opened 2026-06-14
fix: unnecessary PulseAudio deps and clarify backend fallthrough

roderickvd:
`libpulse-dev` was not necessary because `pulseaudio-rs` is pure Rust 🥳 

This takes the opportunity to explain the fallback mechanism from PipeWire > PulseAudio > ALSA depending on what's compiled in and what's available on a system. While it creates some coupling to cpal behavior, I continue to answer questions about users about it.


yara-blue:
> pulseaudio-rs is pure Rust

Love that :)

Issue #901, opened 2026-07-20
Tracking issue for audio engine rewrite


yara-blue:
This tracks landing the changes from https://github.com/RustAudio/rodio-experiments into Rodio. This involves integrating the new rubato resampler and specializing it where possible.

## Background:
For a long time we've had mysterious bugs round spans. Spans allow a source to change parameters (sample-rate and channel count) during playback. We've concluded that spans where a mistake as correctly handling audio that at _any point_ is:
- really hard to do correctly.
- impossible to optimize well leading to bad performance. 

Most audio should be converted to the playback parameters at the very start of the audio pipeline. To make this easier we are adding _two_ new source types. Both do not allow parameters to change, one fixes this at runtime the other needs the parameters known at compile time.

## Approach 
This is a large overhaul at around 7700 lines of code (ignoring docs & comments). To keep things reviewable we'll do this in reasonably sized PRs of around 500 lines. Help is greatly appreciated!

### Phase one
Implementation in: #903 and #904 
- [x] land const source
- [x] remove experimental from fixed source
- [x] land try-into adaptors/convertors between sources (others in phase two)
- [x] note overhaul in Readme
- [x] land most basic generators

### Phase two
Implementation in: #905 and #910 
- [x] channel conversions
- [x] resampler and integrate with rubato

### Phase three: 
Implementation in: #911 
- [x] land one effect and pure_effect! macro
- [ ] land all other effect
- [ ] deprecation warning on effects in DynamicSource that are not sound

### Phase four
- [ ] add new combinators [a, b, c].mixed() etc.
- [ ] add all generators (noise etc, function gen. etc)

### Phase five
- [ ] rename current source to dynamic source
- [ ] add Source = DynamicSource type alias with deprecation warning
- [ ] Remove placeholders

### Phase six
- [ ] port all examples
- [ ] re-add controls to automatic gain control
- [ ] rewrite Rodio main docs
- [ ] update porting guide (needs to be really nice this time :3 )
- [ ] prepare and make announcements

## Small chores
- [ ] check the new queues for: https://github.com/RustAudio/rodio/pull/913

## Open Questions
- Do we deprecate Source in favor of DynamicSource?
- Do cons sources and the ConstSource trait get default generics?

yara-blue:
One point for discussion:

Should the ConstSource have a default generic? as in should it be:
```rust
trait ConstSource<SR=44100, CH=2> { ... }
```
Such that it can be used as:
```rust
[...]

PR #902, opened 2026-07-21
Phase1 of audio engine rewrite

yara-blue:

Sources are very minimal still, to get an idea of what this will all look like see: [rodio-experiments](https://github.com/RustAudio/rodio-experiments).

_Implementation notes:_
- I've moved some documentation into md files and `include_str!` those to prevent duplication. We'll be using that even more for the effects.
- For now does not rename Source but do make it available under it's new name via a pub re-export in lib.rs


yara-blue:
gonna split this in two PRs to stay at a reasonable size

yara-blue:
superseded by #903

PR #903, opened 2026-07-23
Phase1a of audio engine rewrite

yara-blue:
supersedes #902 

Very minimal still, to get an idea of what this will all look like see: [rodio-experiments](https://github.com/RustAudio/rodio-experiments).

_Implementation notes:_
- I've moved some documentation into md files and `include_str!` those to prevent duplication. We'll be using that even more for the effects.
- For now does not rename Source but do make it available under it's new name via a pub re-export in lib.rs
- There is a placeholder so the example in the docs works

### DIFF
--- a/src/common/source.rs
+++ b/src/common/source.rs
@@ -1,0 +1,20 @@
+//! Code shared between source types.
+//!
+//! Since we have three source types there is a lot of code duplication. We
+//! combat that by placing whatever can be shared here.
+//!
+//! This can be:
+//! - shared types like error enums
+//! - shared free functions
+//! - shared members / impl blocks through macro_rules
+//!
+//! # Note
+//! Effects are defined through a macro and do not need this kind of
+//! deduplication
+//!
+//! This modules structure mirrors that of what it deduplicates. For example
+//! the code shared between [fixed_source::chain] and [const_source::chain] is in
+//! common/source/chain.rs
+
+pub(crate) mod buffer;
+pub(crate) mod chain;
--- a/src/common/source/buffer.rs
+++ b/src/common/source/buffer.rs
@@ -37,7 +37,7 @@ macro_rules! source_impl {
 
             // make sure the next sample is for the right channel
             let new_pos = new_pos.next_multiple_of(self.channels().get() as usize);
-            let new_pos = new_pos - curr_channel;
+            let new_pos = new_pos + curr_channel;
 
             self.pos = new_pos;
             Ok(())
@@ -56,7 +56,7 @@ macro_rules! iter_impl {
         }
         #[inline]
         fn size_hint(&self) -> (usize, Option<usize>) {
-            let remaining = self.data.len() - self.pos;
+            let remaining = self.data.len().saturating_sub(self.pos);
             (remaining, Some(remaining))
         }
     };
--- a/src/common/source/buffer.rs
+++ b/src/common/source/buffer.rs
@@ -1,0 +1,66 @@
+macro_rules! source_impl {
+    () => {
+        /// # Panics
+        /// If the length of the buffer is larger than approximately 16 billion elements.
+        /// This is because the calculation of the duration would overflow.
+        #[inline]
+        fn total_duration(&self) -> Option<Duration> {
+            use crate::math::NANOS_PER_SEC;
+
+            let duration_ns = NANOS_PER_SEC
+                .checked_mul(self.data.len() as u64)
+                .expect("slices longer then 16 billion elements are not supported")
+                / self.sample_rate().get() as u64
+                / self.channels().get() as u64;
+            let duration = Duration::new(
+                duration_ns / NANOS_PER_SEC,
+                (duration_ns % NANOS_PER_SEC) as u32,
+            );
+
+            Some(duration)
+        }
+
+        /// This jumps in memory to the sample corresponding to `pos`.
+        #[inline]
+        fn try_seek(&mut self, pos: Duration) -> Result<(), SeekError> {
+            // This is fast because all the samples are in memory already
+            // and due to the constant sample_rate we can jump to the right
+            // sample directly.
+
+            let curr_channel = self.pos % self.channels().get() as usize;
+            let new_pos = crate::math::duration_to_float(pos)
+                * self.sample_rate().get() as crate::Float
+                * self.channels().get() as crate::Float;
+            // saturate pos at the end of the source
+            let new_pos = new_pos as usize;
+            let new_pos = new_pos.min(self.data.len());
+
+            // make sure the next sample is for the right channel
+            let new_pos = new_pos.next_multiple_of(self.channels().get() as usize);
+            let new_pos = new_pos - curr_channel;
+
+            self.pos = new_pos;
+            Ok(())
+        }
+    };
+}
+
+macro_rules! iter_impl {
+    () => {
+        type Item = Sample;
+        #[inline]
+        fn next(&mut self) -> Option<Self::Item> {
+            let sample = self.data.get(self.pos)?;
+            self.pos += 1;
+            Some(*sample)
+        }
+        #[inline]
+        fn size_hint(&self) -> (usize, Option<usize>) {
+            let remaining = self.data.len() - self.pos;
+            (remaining, Some(remaining))
+        }
+    };
+}
+
+pub(crate) use iter_impl;
+pub(crate) use source_impl;
--- a/src/common/source/chain.rs
+++ b/src/common/source/chain.rs
@@ -2,15 +2,21 @@ use crate::source::SeekError;
 
 #[derive(Debug, thiserror::Error)]
 pub enum ChainSeekError {
-    #[error("Could not get duration of first source ({ty}")]
+    #[error("Could not get duration of first source ({ty})")]
     NoTotalDurationForFirst { ty: &'static str },
-    #[error("Could not seek in first source")]
+    #[error("Could not seek in first source ({ty})")]
     FailedToSeekInFirst {
         ty: &'static str,
         #[source]
         error: SeekError,
     },
-    #[error("Could not seek in second source")]
+    #[error("Could not reset first source ({ty}) to start")]
+    FailedToResetFirst {
+        ty: &'static str,
+        #[source]
+        error: SeekError,
+    },
+    #[error("Could not seek in second source ({ty})")]
     FailedToSeekInSecond {
         ty: &'static str,
         #[source]
@@ -48,6 +54,23 @@ macro_rules! source_impl {
             };
 
             if pos < first {
+                // Reset first source to prevent a jump to the current position
+                // after the first source completes again.
+                if !self.playing_first {
+                    // FIXME(yara): implement Seekable trait for all sources and extract
+                    // this to a function. (all sources are required to impl Seekable).
+                    // Might wanna do a similar thing for other shared functionality
+                    // like total duration
+                    self.second
+                        .try_seek(std::time::Duration::ZERO)
+                        .map_err(|error| ChainSeekError::FailedToResetFirst {
+                            ty: type_name_of_val(&self.first),
+                            error,
+                        })
+                        .map_err(Arc::new)
+                        .map_err(|e| SeekError::Other(e))?;
+                }
+
                 self.first
                     .try_seek(pos)
                     .map_err(|error| ChainSeekError::FailedToSeekInFirst {
@@ -55,16 +78,20 @@ macro_rules! source_impl {
                         error,
                     })
                     .map_err(Arc::new)
-                    .map_err(|e| SeekError::Other(e))
+                    .map_err(|e| SeekError::Other(e))?;
+                self.playing_first = true;
+                Ok(())
             } else {
                 self.second
-                    .try_seek(pos)
+                    .try_seek(pos - first)
                     .map_err(|error| ChainSeekError::FailedToSeekInSecond {
                         ty: type_name_of_val(&self.second),
                         error,
                     })
                     .map_err(Arc::new)
-                    .map_err(|e| SeekError::Other(e))
+                    .map_err(|e| SeekError::Other(e))?;
+                self.playing_first = false;
+                Ok(())
             }
         }
     };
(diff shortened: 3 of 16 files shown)


src/const_source/buffer.rs
```diff
@@ -0,0 +1,40 @@
+use std::sync::Arc;
+use std::time::Duration;
+
+use crate::source::SeekError;
+use crate::ConstSource;
+use crate::Sample;
+
+/// A buffer of samples treated as a source.
+#[derive(Debug, Clone)]
+pub struct SamplesBuffer<const SR: u32, const CH: u16> {
```
roderickvd, line 10: Can we use the `SampleRate` and `ChannelCount` type aliases?

src/common/source.rs
```diff
@@ -0,0 +1,20 @@
+//! Code shared between source types.
+//!
+//! Since we have three source types there is a lot of code duplication. We
```
roderickvd, line 3: I thought this shouldn't be user-facing Rustdoc when I saw it probably isn't, because the module isn't imported as public. Still, may want to consider the Rustdoc level.

src/const_source/buffer.rs
```diff
@@ -0,0 +1,40 @@
...
+    data: Arc<[Sample]>,
+    pos: usize,
+}
+
+impl<const SR: u32, const CH: u16> SamplesBuffer<SR, CH> {
+    /// Builds a new `SamplesBuffer`.
+    ///
+    /// Note any call to total_duration will panic if the buffer is larger then
```
roderickvd: I see you moved that code from `new` to `total_duration` which probably makes sense. But I'd say it doesn't need docs here then.

src/const_source/buffer.rs
```diff
@@ -0,0 +1,40 @@
...
+    /// 16 billion elements.
+    pub fn new<D>(data: D) -> SamplesBuffer<SR, CH>
+    where
+        D: Into<Vec<Sample>>,
+    {
+        const { assert!(SR > 0) };
```
roderickvd, line 21: Not too familiar with const generics; I guess we can't use `NonZero` as we wanted?

src/common/source/chain.rs
```diff
@@ -0,0 +1,94 @@
...
+                    .try_seek(pos)
+                    .map_err(|error| ChainSeekError::FailedToSeekInFirst {
+                        ty: type_name_of_val(&self.first),
+                        error,
+                    })
+                    .map_err(Arc::new)
+                    .map_err(|e| SeekError::Other(e))
+            } else {
+                self.second
+                    .try_seek(pos)
```
roderickvd: I think this needs to be `pos - first`. As-is, chaining two 1-second sources and seeking to 1.5s puts the second source at 1.5s, which is past its own end.

src/common/source/chain.rs
```diff
@@ -0,0 +1,94 @@
...
+            self.first.sample_rate()
+        }
+
+        fn total_duration(&self) -> Option<std::time::Duration> {
+            self.first
+                .total_duration()
+                .and_then(|d| self.second.total_duration().map(|d2| d2 + d))
+        }
+
+        fn try_seek(&mut self, pos: std::time::Duration) -> Result<(), crate::source::SeekError> {
```
roderickvd, line 43: `try_seek` should probably set `playing_inner = pos < first` and seek both sources. Seeking forward to 1.5s from the start replays all of the first source, and seeking back to 0 after the chain is drained keeps reading the exhausted second.

src/common/source/buffer.rs
```diff
@@ -0,0 +1,66 @@
...
+            let new_pos = crate::math::duration_to_float(pos)
+                * self.sample_rate().get() as crate::Float
+                * self.channels().get() as crate::Float;
+            // saturate pos at the end of the source
+            let new_pos = new_pos as usize;
+            let new_pos = new_pos.min(self.data.len());
+
+            // make sure the next sample is for the right channel
+            let new_pos = new_pos.next_multiple_of(self.channels().get() as usize);
+            let new_pos = new_pos - curr_channel;
```
roderickvd: This would underflow when seeking to 0 when we are mid-frame (`new_pos == 0 && curr_channel > 0`) and happened to be right for stereo but not on multi-channel audio.

We need to _add_ the offset, not subtract it, to get from the start of a frame back to `curr_channel`. And defensively clamp against `data.len()` probably.

src/common/source/buffer.rs
```diff
@@ -0,0 +1,66 @@
...
+        type Item = Sample;
+        #[inline]
+        fn next(&mut self) -> Option<Self::Item> {
+            let sample = self.data.get(self.pos)?;
+            self.pos += 1;
+            Some(*sample)
+        }
+        #[inline]
+        fn size_hint(&self) -> (usize, Option<usize>) {
+            let remaining = self.data.len() - self.pos;
```
roderickvd: `try_seek` right now could cause `self.pos` to exceed `self.data.len()` causing an underflow here.

src/const_source/chain.rs
roderickvd, line 1: This still needs `try_seek`, `size_hint` and `ExactSizeIterator` implemented.

src/common/source/chain.rs
```diff
@@ -0,0 +1,94 @@
+use crate::source::SeekError;
+
+#[derive(Debug, thiserror::Error)]
+pub enum ChainSeekError {
+    #[error("Could not get duration of first source ({ty}")]
```
roderickvd: Add closing `)`.

src/common/source/chain.rs
```diff
@@ -0,0 +1,94 @@
+use crate::source::SeekError;
+
+#[derive(Debug, thiserror::Error)]
+pub enum ChainSeekError {
```
roderickvd, line 4: Should this be exported publicly so users can downcast to them from a `SeekError::Other(Arc<dyn Error>)`?

src/common/source/chain.rs
```diff
@@ -0,0 +1,94 @@
...
+    #[error("Could not get duration of first source ({ty}")]
+    NoTotalDurationForFirst { ty: &'static str },
+    #[error("Could not seek in first source")]
```
roderickvd: Add `({ty})` here too? And below.

src/generators/silence.rs
```diff
@@ -0,0 +1,62 @@
...
+        fn try_seek(&mut self, _: Duration) -> Result<(), SeekError> {
+            Ok(())
+        }
+    }
+
+    impl<const SR: u32> Iterator for Silence<SR> {
+        type Item = crate::Sample;
+
+        fn next(&mut self) -> Option<Self::Item> {
+            Some(0.0)
```
roderickvd: `Sample::EQUILIBRIUM`

src/const_source/buffer.rs
yara-blue: Sadly not, const params are limited to primitive types atm :(

src/const_source/buffer.rs
yara-blue: yes, pretty sad. One day we'll get rustc there. But thanks to the `const` block that assert will fire compile time.

src/common/source.rs
yara-blue: It's still useful to use doc comments instead of simple comments as rust-analyzer makes hover doc available for this when working on Rodio.

src/common/source/buffer.rs
yara-blue: Instead of defensive clamping I've changed the size hint to use `saturating_sub`. Making pos guaranteed to be between zero and len is an option but then I'd like to encode that in the type system, for now that seems overkill. 

Though I bet we have enough slice + offset parts in Rodio now to make that worth it... Some other time...

src/common/source/buffer.rs
yara-blue: thanks! will be using saturating sub next see reasoning above

src/common/source/chain.rs
yara-blue: It should be re-exported in the source implementation as wel. See the: `pub use crate::common::source::chain::ChainSeekError;` in `const_source/chain.rs`

yara-blue:
Thanks for the review! Should have addressed all of the feedback. I'll see if I can fix the Phase1b tonight.

PetrGlad:
OK, I see channel and seeking precision bugs I saw yesterday, are already fixed :) 
I do not really like use of macroses(like  `macro_rules! source_impl`) since it is harder to see usages, and it postpones some checks to final compilation. Was there a reason to use macroses instead of traits, or is it just to save typing? This implementation would do, bit I'd considered something more explicit.

yara-blue:
First of all thanks for finding the time for the review! Secondly, I did not have the time to write a short answer so you get a _very_ long one, apologies.

> Was there a reason to use macroses instead of traits, or is it just to save typing? This implementation would do, bit I'd considered something more explicit.

There are going to be even more macro calls when we get to implementing all the effects... Lets see what we can do:

### Traits
I've tried doing something with traits, separating out the repeated parts:

```rust
trait FixedSource {
  fn sample_rate(&self) -> SampleRate;
  fn channel_count(&self) -> ChannelCount;
  fn total_duration(&self) -> Duration;
  fn try_seek(&mut self, pos: Duration) -> Result<(), SeekError>;
}
trait ConstSource {
  fn sample_rate(&self) -> SampleRate;
  fn channel_count(&self) -> ChannelCount;
  fn total_duration(&self) -> Duration;
  fn try_seek(&mut self, pos: Duration) -> Result<(), SeekError>;
}
```
We would have:
```rust
trait FixedSource: TotalDuration + TrySeek {
  fn sample_rate(&self) -> SampleRate;
  fn channel_count(&self) -> ChannelCount;
}
trait ConstSource: TotalDuration + TrySeek {
  fn sample_rate(&self) -> SampleRate;
  fn channel_count(&self) -> ChannelCount;
}
trait TotalDuration {
  fn total_duration(&self) -> Duration;
}
trait TrySeek {
  fn try_seek(&mut self, pos: Duration) -> Result<(), SeekError>;
}
```

Now we can implement TotalDuration and TrySeek separately. This does not win us anything unfortunately, we still need a separate object for `ConstSource` since it will need to have const generic args.

### Marker trait
Why is `ConstSource` not just a marker trait for FixedSource?

You'd do:
```rust
impl FixedSource for Amplify {
  fn sample_rate() -> SampleRate {
     ... 
  }
  ...
}

impl<SR, CH> ConstSource for Amplify<SR, CH>;
```

Well the reason for that is we do not want any type to implement both `FixedSource`  _and_ `ConstSource`. They mean different things so they must be separate types. Still we seem to be onto something here. Lets turn it around and wrap instead of marking:

### Wrapping
```rust
struct Amplify(fixed_source::Amplify);

impl<SR, CH> ConstSource<SR, CH> for Amplify {
   fn total_duration(&self) -> Duration {
     self.0.total_duration()
   }
   ...
}
```

That seems nice, no more code duplication and no more macro's. But then we realize that S will probably wrap some existing source (like chain does or a future amplify (in phase 3a). But fixed_source::S can only be generic over a `FixedSource` and S here is a `ConstSource` so it is generic over one of those.

```rust
// Error: argument T to fixed_source::Amplify must implement FixedSource
struct Amplify<T: ConstSource>(fixed_source::Amplify<T>); 
```

Well for that we have the IntoFixedSource wrapper! So we get:

```rust
struct Amplify<T>(fixed_source::Amplify<IntoFixedSource<T>);

impl<SR, CH> ConstSource<SR, CH> for Amplify {
   fn total_duration(&self) -> Duration {
     self.0.total_duration()
   }
   ...
}
```

I don't like that this leads to a lot of IntoFixedSource's in there but those should compile out... I have to think about it. What do you all think?

PetrGlad:
Yes, this is the explanation I was looking for. Of the versions you mentioned this one I had in mind, actually:
```rust
trait FixedSource: TotalDuration + TrySeek {
  fn sample_rate(&self) -> SampleRate;
  fn channel_count(&self) -> ChannelCount;
}
trait ConstSource: TotalDuration + TrySeek {
  fn sample_rate(&self) -> SampleRate;
  fn channel_count(&self) -> ChannelCount;
}
trait TotalDuration {
  fn total_duration(&self) -> Duration;
}
trait TrySeek {
  fn try_seek(&mut self, pos: Duration) -> Result<(), SeekError>;
}
```
Regarding `ConstSource`, why not 
```rust
trait ConstSource<const SR: u32, const CH: u16>: TotalDuration + TrySeek {
  fn sample_rate(&self) -> SampleRate;
  fn channel_count(&self) -> ChannelCount;
}
```

PetrGlad:
// I understand that `FixedSource` is primarily a temporary shim to distinguish it from existing Source. But for a reader "ConstSource" may imply that it supplies a constant value for each channel, for example. A better name for what is now `ConstSource` would be actually `FixedSource`... This renaming can be done later, but in its present form I find the naming confusing.

PetrGlad:
Regarding simultaneous implementation of `ConstSource` and `FixedSource` - this does not seem to be an actual problem, but maybe there will be situations where it could be (e.g. you anticipate API that cannot be supported by both)?

yara-blue:
> I understand that FixedSource is primarily a temporary shim to distinguish it from existing Source.

Not really, for example the implementation of queue gets far simpler (and more performant) when it no longer needs to watch for parameter changes. That was the primary motivation for this rework.

> But for a reader "ConstSource" may imply that it supplies a constant value for each channel, for example. A better name for what is now ConstSource would be actually FixedSource... This renaming can be done later, but in its present form I find the naming confusing.

I would not mind _eventually_ removing `DynamicSource` (the current source) but ideally the next release does not do that yet. That way users can slowly move over. (We can eventually mark it as deprecated then later remove). Even if we rename the current `Source` to `DynamicSource` in the upcoming release I'm not sure if we should reuse the `Source` name (could be pretty confusing).

> Regarding `ConstSource`, why not
> 
> ```rust
> trait ConstSource<const SR: u32, const CH: u16>: TotalDuration + TrySeek {
>   fn sample_rate(&self) -> SampleRate;
>   fn channel_count(&self) -> ChannelCount;
> }
> ```

We could, but it would not really help us. You'd still need to implement four methods except now they are divided by between three traits. It would help us if we could simultaneously implement the source trait for one object (see next answer for why that is impossible).

> Regarding simultaneous implementation of `ConstSource` and `FixedSource` - this does not seem to be an actual problem

do you mean a struct implementing both traits? While you _can_ do that it becomes really annoying to then use the struct. You'd have to disambiguate (specify which trait you want a method from) every time you call `.amplify()` or `.take_duration()` etc. Only way to solve that is to have two separate objects each implementing only one trait.

PetrGlad:
Regarding `ConstSource`, what I meant is that the naming is confusing. Ideally, what I would prefer is to eventually rename 
ConstSource -> FixedSource (or StaticSource).
DynamicSource -> Source

yara-blue:
> Regarding `ConstSource`, what I meant is that the naming is confusing. Ideally, what I would prefer is to eventually rename ConstSource -> FixedSource (or StaticSource). DynamicSource -> Source

It's a little bike-sheddy but what are the three names you would like in case we do not remove the _current_ source trait?

Edit: we might very well keep a source with spans since it is the easiest to provide to Rodio. With that I mean if you have a wonky source of audio frames of which the parameters keep changing you provide that as a `DynamicSource` and then immediately call `into_fixed()` on it.

yara-blue:
I'm going to go ahead and merge this. We can change the name of the sources at any point in the future (before release).

Regarding implementing the Const sources as wrappers for Fixed. We cannot do that everywhere but we should where we can. I'll update #904 to do that.

PR #904, opened 2026-07-23
Phase1b of audio engine rewrite

yara-blue:
review **after** #903 (this includes it's changes)

- Expands the fixed source impl. to match const source
- de-duplicates methods between them using a new `common::source` module and macros
- makes a note in the changelog
- uses `Placeholder` to have docs and examples point to not yet landed code

### DIFF
--- a/src/common/source/chain.rs
+++ b/src/common/source/chain.rs
@@ -1,0 +1,94 @@
+use crate::source::SeekError;
+
+#[derive(Debug, thiserror::Error)]
+pub enum ChainSeekError {
+    #[error("Could not get duration of first source ({ty}")]
+    NoTotalDurationForFirst { ty: &'static str },
+    #[error("Could not seek in first source")]
+    FailedToSeekInFirst {
+        ty: &'static str,
+        #[source]
+        error: SeekError,
+    },
+    #[error("Could not seek in second source")]
+    FailedToSeekInSecond {
+        ty: &'static str,
+        #[source]
+        error: SeekError,
+    },
+}
+
+macro_rules! source_impl {
+    () => {
+        fn channels(&self) -> crate::ChannelCount {
+            self.first.channels()
+        }
+
+        fn sample_rate(&self) -> crate::SampleRate {
+            self.first.sample_rate()
+        }
+
+        fn total_duration(&self) -> Option<std::time::Duration> {
+            self.first
+                .total_duration()
+                .and_then(|d| self.second.total_duration().map(|d2| d2 + d))
+        }
+
+        fn try_seek(&mut self, pos: std::time::Duration) -> Result<(), crate::source::SeekError> {
+            use crate::source::SeekError;
+            use std::any::type_name_of_val;
+            use std::sync::Arc;
+
+            let Some(first) = self.first.total_duration() else {
+                return Err(ChainSeekError::NoTotalDurationForFirst {
+                    ty: type_name_of_val(&self.first),
+                })
+                .map_err(Arc::new)
+                .map_err(|e| SeekError::Other(e));
+            };
+
+            if pos < first {
+                self.first
+                    .try_seek(pos)
+                    .map_err(|error| ChainSeekError::FailedToSeekInFirst {
+                        ty: type_name_of_val(&self.first),
+                        error,
+                    })
+                    .map_err(Arc::new)
+                    .map_err(|e| SeekError::Other(e))
+            } else {
+                self.second
+                    .try_seek(pos)
+                    .map_err(|error| ChainSeekError::FailedToSeekInSecond {
+                        ty: type_name_of_val(&self.second),
+                        error,
+                    })
+                    .map_err(Arc::new)
+                    .map_err(|e| SeekError::Other(e))
+            }
+        }
+    };
+}
+
+macro_rules! iter_impl {
+    () => {
+        type Item = Sample;
+
+        fn next(&mut self) -> Option<Self::Item> {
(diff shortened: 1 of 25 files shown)


src/fixed_source.rs
```diff
@@ -16,4 +23,207 @@ pub trait FixedSource: Iterator<Item = Sample> {
...
+
+impl<S: FixedSource> Iterator for Placeholder<S> {
+    type Item = Sample;
+    fn next(&mut self) -> Option<Self::Item> {
+        unimplemented!("placeholder")
+    }
+}
+
+/// A [`ConstSource`] adapted from a [`FixedSource`].
+pub struct IntoConstSource<const SR: u32, const CH: u16, S: FixedSource>(S);
```
roderickvd, line 160: This one and `IntoDynamicSource` should forward `try_seek`.

src/const_source.rs
```diff
@@ -0,0 +1,214 @@
...
+    S: ConstSource<SR, CH>,
+{
+    type Item = Sample;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.inner.next()
+    }
+}
+
+impl<const SR: u32, const CH: u16, S> FixedSource for IntoFixedSource<SR, CH, S>
```
roderickvd, line 199: Should forward `try_seek`.

src/fixed_source/chain.rs
```diff
@@ -0,0 +1,61 @@
...
+/// Two chained sources, the one played after the other.
+#[derive(Clone)]
+pub struct SourceChain<S1, S2> {
+    first: S1,
+    second: S2,
+    playing_inner: bool,
+}
+
+#[derive(Debug, Clone, Copy, thiserror::Error, PartialEq, Eq)]
+pub struct ParamsMismatch {
```
roderickvd, line 18: Consider exporting this one so users can downcast to it.

src/common/source/chain.rs
```diff
@@ -0,0 +1,94 @@
...
+                        error,
+                    })
+                    .map_err(Arc::new)
+                    .map_err(|e| SeekError::Other(e))
+            }
+        }
+    };
+}
+
+macro_rules! iter_impl {
```
roderickvd, line 100: Should re-add `size_hint()`.

src/docs/collect_into_buffer.md
roderickvd, line 1: This seems misplaced or having the wrong name.

src/generators/silence.rs
roderickvd, line 1: As per #913 we should add the ability to create one with an arbitrary channel count. Over there we targeted it for phase 4, so just mentioning it here.

roderickvd (commented):
I see this is built on top of #903 so I didn't duplicate any of my review points there to here.

src/docs/collect_into_buffer.md
yara-blue: how so? it is the documentation for `FixedSource::collect_into_buffer` and `ConstSource::collect_into_buffer`.

I'll expand the text a little to make that more clear.

src/generators/silence.rs
yara-blue: For silence the idea is to always put it through a channel convertor. My hope being that the optimizer can figure it out. If that is false and perf. becomes an issue we could specialize it.

src/docs/collect_into_buffer.md
roderickvd: I thought it was out of place because the title was about building a `SamplesBuffer`, not about collecting into a buffer, and went on to describe a panic that wouldn't be occurring in `new` (the "builder") anymore, but in `total_duration`.

src/generators/silence.rs
roderickvd: Automatic optimization would be great, but how would that work?

yara-blue: the compiler knows that `Silence::next` returns a constant value (namely: `Some(0.0)`).

Lets take this fragment of the channel convertor:
```
        let result = match self.next_output_sample_pos {
            0 => {
                let value = self.input.next();
                self.sample_repeat = value;
                value
            }
            x if x < self.from.get() => self.input.next(),
            1 => self.sample_repeat,
            _ => Some(Sample::EQUILIBRIUM),
        };
```

Every `self.input.next()` gets replaced with `Some(0)`:
```
        let result = match self.next_output_sample_pos {
            0 => {
                self.sample_repeat = 0.0;
                0.0
            }
            x if x < self.from.get() => 0.0,
            1 => self.sample_repeat,
            _ => Some(0.0),
        };
```

Now its simple to see that self.sample_repeat is also constant. So everywhere we see it replace with a `0.0`.
```
        let result = match self.next_output_sample_pos {
            0 => {
                0.0
            }
            x if x < self.from.get() => 0.0,
            1 => 0.0,
            _ => Some(0.0),
        };
```
And this is now just a match statement returning `0.0` always so `result` can be set to `0.0`.

There is some other stuff like `self.from.get()` and `self.next_output_sample_pos()` but that has also become dead code so the optimizer throws that out as well. And now we've gotten to the point where ChannelCountConvertor::next() can be replaced with just `0.0`. This _should_ domino up any audio pipeline.

Now if you put silence in a box and cast it to a Box<dyn FixedSource>... then this stops working.

Whether this all happens? It probably does, these are basic optimizations. You can never know for sure, sometimes a suggestion here and there is needed (a little `#[inline]` usually).

roderickvd: Ah, thanks for that write-up, that makes sense. As long as it doesn't get fed into Rubato or something, because then all bets would be off I imagine.

PR #905, opened 2026-07-23
Phase2a of audio engine rewrite

yara-blue:
review **after** #904 (this includes it's changes)

- Deduplicates all into_inner etc logic 
- Adds channel conversion to both FixedSource and ConstSource

### DIFF
--- a/src/const_source/buffer.rs
+++ b/src/const_source/buffer.rs
@@ -1,0 +1,80 @@
+use std::sync::Arc;
+use std::time::Duration;
+
+use crate::Sample;
+use crate::ConstSource;
+
+/// A buffer of samples treated as a source.
+#[derive(Debug, Clone)]
+pub struct SamplesBuffer<const SR: u32, const CH: u16> {
+    data: Arc<[Sample]>,
+    pos: usize,
+}
+
+impl<const SR: u32, const CH: u16> SamplesBuffer<SR, CH> {
+    /// Builds a new `SamplesBuffer`.
+    ///
+    /// Note any call to total_duration will panic if the buffer is larger then
+    /// 16 billion elements.
+    pub fn new<D>(data: D) -> SamplesBuffer<SR, CH>
+    where
+        D: Into<Vec<Sample>>,
+    {
+        const { assert!(SR > 0) };
+        const { assert!(CH > 0) };
+
+        SamplesBuffer {
+            data: data.into().into(),
+            pos: 0,
+        }
+    }
+}
+
+impl<const SR: u32, const CH: u16> ConstSource<SR, CH> for SamplesBuffer<SR, CH> {
+    /// # Panics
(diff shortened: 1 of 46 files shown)


src/const_source/conversions/channel_count.rs
```diff
@@ -0,0 +1,78 @@
+use std::time::Duration;
+
+use crate::source::SeekError;
+use crate::{ConstSource, Sample};
+
+/// Converts between two channel counts, created using
+/// [with_channel_count](ConstSource::with_channel_count)
+pub struct ChannelConvertor<const SR: u32, const CH_IN: u16, const CH_OUT: u16, S> {
```
PetrGlad, line 8: As I understand this is intended as 1-to-1 rewrite, just keeping old functionality?
We had discussion a while ago regarding this source, in some cases just dropping channels is not enough, it may require some mixing to work properly. (see e.g. https://github.com/RustAudio/rodio/pull/671 - which I fancy to necromance).

src/const_source/buffer.rs
```diff
@@ -0,0 +1,40 @@
...
+use crate::Sample;
+
+/// A buffer of samples treated as a source.
+#[derive(Debug, Clone)]
+pub struct SamplesBuffer<const SR: u32, const CH: u16> {
+    data: Arc<[Sample]>,
+    pos: usize,
+}
+
+impl<const SR: u32, const CH: u16> SamplesBuffer<SR, CH> {
```
PetrGlad, line 15: Maybe it was discussed already,, but why not `SampleRate`?

src/const_source/conversions/channel_count.rs
PetrGlad, line 8: Was this intended as just refactoring (keeping old behavior)?
I just wanted to note that we had discussion a while ago regarding this, and in some cases properly converting number of channels requires some mixing (se e.g. https://github.com/RustAudio/rodio/pull/656 which I fancy to necromance).

src/fixed_source.rs
```diff
@@ -16,4 +25,215 @@ pub trait FixedSource: Iterator<Item = Sample> {
...
+    sample_rate: SampleRate,
+    channel_count: ChannelCount,
+}
+
+impl<const SR: u32, const CH: u16> std::error::Error for ParameterMismatch<SR, CH> {}
+
+impl<const SR: u32, const CH: u16> std::fmt::Display for ParameterMismatch<SR, CH> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        if self.sample_rate.get() == SR && self.channel_count.get() == CH {
+            unreachable!("ParameterMismatch error can only occur when params mismatch");
```
PetrGlad, line 201: // Well, it is reachable, just not expected, `assert`, maybe?

PetrGlad (approved):

Issue #367, opened 2021-05-18
Some vorbis files distorted over USB audio on macOS Big Sur


herrernst:
I have an Ogg Vorbis file that sounds distorted when played back with rodio over an USB audio device (have tried 2 different ones), but works over the headphone jack (macOS 11.3.1 on a Mac mini 2018; no problem on MacBook Pro 2015 with macOS 10.15.7).

I can reproduce on the current master, I just change the path to my file in examples/music_ogg.rs and run `cargo run --example music_ogg`. If I `oggdec` the file and use it with examples/music_ogg.wav, it works. Somehow, there must be a problem with decoding vorbis data under some circumstances (it works in VLC, ffplay, Audacity, so the Vorbis data itself should be no problem).

Here is the track in Vorbis and as WAV: https://easyupload.io/m/crad1n

est31:
It might be some sampling rate issue? No idea.

herrernst:
I did a bisect. Before commit 3bc614fc12fae8afcb81f568efe215fe6a968993, it was crackling way more (and the .wav was also crackling), so the changes seem to have done something, but not resolved it completely. But could also be unrelated.

herrernst:
> It might be some sampling rate issue? No idea.

Seems so. The external USB interfaces were set to 48 kHz, whereas the internal one was set to 44.1 kHz. When using 44.1 kHz for the USB interface, the Vorbis file (which also is 44.1 kHz) works without crackling/distortion.

Will change the title accordingly.

PR #908, opened 2026-07-25
Don't end a resampler chunk at a same-format span boundary

### LINKED ISSUE #907: Resampler output depends on input span chunking — Vorbis-sized spans destroy the signal (regression on master, bisected to ca72dac)
## Summary

On current master, `SampleRateConverter`'s output depends on **how the input is chunked into spans**. Feeding the exact same samples, in the same order, with the same rate and channel count, produces a badly corrupted signal when the source reports many small spans instead of one large one.

This matters in practice because `VorbisDecoder::current_span_len()` returns the length of the current decoded packet, so a normal Ogg Vorbis stream arrives as hundreds of small spans. Any Vorbis file whose sample rate differs from the output device rate plays back with audible hiss / stutter / clipping.

Released `v0.22.2` is unaffected — it predates the rubato resampler. This is a regression introduced on master.

## Reproduction

Self-contained, using the existing `TestSource` in `src/conversions/sample_rate/tests.rs`. No audio device and no Vorbis file needed — the invariant is simply *"chunking the input differently must not change the output"*.

```rust
#[test]
fn resample_output_is_independent_of_span_chunking() {
    let rate = SampleRate::new(48000).unwrap();
    let target = SampleRate::new(44100).unwrap();
    let channels = ChannelCount::new(2).unwrap();

    // 0.5 s of a 440 Hz stereo sine at 48 kHz.
    let frames = 24_000usize;
    let signal: Vec<Sample> = (0..frames)
        .flat_map(|i| {
            let t = i as f32 / 48_000.0;
            let s = (std::f32::consts::TAU * 440.0 * t).sin() * 0.5;
            [s, s]
        })
        .collect();

    let resample = |source: TestSource| -> Vec<Sample> {
        SampleRateConverter::new(source, target, ResampleConfig::poly().build()).collect()
    };

    // One span holding the whole signal.
    let single = resample(TestSource::new(signal.clone(), rate, channels));

    // Same samples, same order, same format — split into packet-sized spans
    // the way a Vorbis decoder reports them.
    let packet = 1024 * channels.get() as usize;
    let mut chunks = signal.chunks(packet);
    let mut multi = TestSource::new(chunks.next().unwrap().to_vec(), rate, channels);
    for c in chunks {
        multi = multi.chain(c.to_vec(), rate, channels);
    }
    let multi = resample(multi);

    let compare = single.len().min(multi.len());
    let max_err = (0..compare)
        .map(|i| (single[i] - multi[i]).abs())
        .fold(0.0f32, f32::max);

    assert!(max_err < 0.01, "max |diff| = {max_err:.6}");
}
```

Result on `c5f1e94`:

```
single-span: 44124 samples, multi-span: 44110 samples, max |diff| = 0.494097
```

The signal amplitude is `0.5`, so a max absolute difference of `0.494` is near-total destruction at the worst point. 14 frames are also lost.

## Bisect

```
last good: 46dd8de  2026-07-12  Deduplicate sample/frame duality
first bad: ca72dac  2026-07-13  add dedicated input datastructure
```

`git bisect run` over `46dd8de..c5f1e94` with the test above.

## Mechanism

`fill_input_buffer` clamps each read to `current_span_len()`, and `resample_chunk` then flags any short read as a *partial* chunk to rubato:

```rust
partial_len: (frames_in < needed_by_resampler).then_some(frames_in.raw()),
```

`partial_len` means "short/final chunk", which rubato zero-pads. On a continuous stream this should fire once, at the true end. Instrumenting the run above shows it firing on **25 of 46** chunk calls (54%) — a discontinuity injected at essentially every span boundary.

The change in `ca72dac` was to the units the clamp is computed in:

```rust
// 46dd8de (good) — span converted to FRAMES, min taken in frames
let current_span_length = self.input.current_span_len()
    .map(InSamples)
    .map(|s| s.frames(num_channels));
let frames_to_take =
    needed_by_resampler.min(current_span_length.unwrap_or(InFrameCount::MAX));

// ca72dac (bad) — min taken in SAMPLES
let current_span_length = self.input.current_span_len().map(InSamples);
let frames_to_take = needed_by_resampler
    .samples(self.output.channels)
    .min(current_span_length.unwrap_or(InSamples::MAX));
```

I haven't fully pinned down why the sample-domain comparison makes the clamp bite so much more often, so treat that last part as a lead rather than a conclusion — the bisect and the repro are the solid parts.

## Note on the existing tests

`test_span_boundary_same_format` covers this shape but only asserts the **output sample count**, not signal integrity, so it passes while the audio is destroyed. The test above is offered as a regression test if useful.

## Possibly related

- #367 — "Some vorbis files distorted when sample rate is different from audio device" (2021). Same symptom, but far predates this resampler; this regression may be a second, independent cause of the same report.
- #584 — "Crackly audio on some mp3 files (suspected bug in resampler)". MP3 decoders also report per-packet spans, so it may share this root cause.
- #694 / #712 — the span abstraction discussions. This is a concrete case where treating decoder packet boundaries as resampler chunk boundaries silently corrupts audio.

## Environment

- rodio `master` @ `c5f1e94`
- cpal 0.18.1, rubato 3.0.0
- macOS, 44.1 kHz output device, 48 kHz Ogg Vorbis source

benface:
Fixes #907.

## Problem

`fill_input_buffer` clamps each read to `current_span_len()`, and `resample_chunk` then flags any short read as `partial_len`:

```rust
partial_len: (frames_in < needed_by_resampler).then_some(frames_in.raw()),
```

`partial_len` means "short/final chunk" and rubato zero-pads it. Decoders report one span per decoded packet, so a single continuous stream arrives as many small spans — and a discontinuity was being injected at every boundary. Instrumenting a 0.5 s resample showed `partial_len` firing on 25 of 46 chunk calls.

Vorbis is the clearest case: it switches to short blocks (commonly 128 or 256 samples) around transients, so real files move in and out of the degraded regime depending on programme material. That matches the "only some files, only on some devices" character of #367 and #584.

## Fix

A span boundary only matters to the resampler when the sample rate or channel count actually *changes*. So keep pulling input across boundaries, and stop only on a genuine format change or end of stream. `partial_len` then fires only at real end-of-stream, which is what it means.

## Measurements

440 Hz stereo sine, 48 kHz → 44.1 kHz, measuring the fraction of output energy away from the fundamental (Goertzel on the left channel) while varying only the span size:

| span size | before | after |
|---|---|---|
| single span | 0.054 % | 0.054 % |
| 4096 frames | 0.054 % | 0.054 % |
| 2048 frames | 0.054 % | 0.054 % |
| 1024 frames | **1.02 %** | 0.054 % |
| 512 frames | **23.79 %** | 0.054 % |
| 256 frames | **90.96 %** | 0.054 % |
| 128 frames | **91.00 %** | 0.054 % |

Output is bit-identical to the single-span case at every span size once chunks are no longer cut at same-format boundaries. Output length also stops drifting with span size (44110 / 44156 / 44250 → a consistent 44124).

Confirmed by ear as well: the downstream game where this surfaced (48 kHz Ogg Vorbis on a 44.1 kHz output device, previously audible hiss/stutter) plays clean with this patch.

## Test

Adds `small_spans_do_not_degrade_resampled_signal`, asserting the invariant directly — output must stay spectrally pure regardless of span size. It **fails on the parent commit** (1024-frame spans already breach the threshold) and passes here.

Worth noting the existing `test_span_boundary_same_format` covers this shape but only asserts the output *sample count*, which is why it stayed green while the audio was being corrupted.

## Scope

- Bisected to ca72dac ("add dedicated input datastructure"). Released `v0.22.2` predates the rubato resampler and is unaffected.
- `cargo test` 71 passing, `cargo fmt --check` and `cargo clippy` clean.

I'm happy to adjust the approach — I realise this module is under active refactoring, so if the intended design is for `fill_input_buffer` to stay single-span and the padding to be handled downstream instead, I can rework it that way.

### DIFF
--- a/src/conversions/sample_rate/rubato.rs
+++ b/src/conversions/sample_rate/rubato.rs
@@ -193,19 +193,41 @@ impl<I: Source, R: rubato::Resampler<Sample>> RubatoResample<I, R> {
     }
 
     fn fill_input_buffer(&mut self, needed_by_resampler: InFrameCount) -> InFrameCount {
-        let current_span_length = self.input.current_span_len().map(InSamples);
-        let frames_to_take = needed_by_resampler
-            .samples(self.output.channels)
-            .min(current_span_length.unwrap_or(InSamples::MAX));
-
+        let wanted = needed_by_resampler.samples(self.output.channels);
         self.input_buffer.clear();
-        for _ in 0..frames_to_take.raw() {
-            if let Some(sample) = self.input.next() {
-                self.input_buffer.push(sample);
-            } else {
+
+        // Keep pulling across span boundaries while the format is unchanged.
+        // A span boundary only matters to the resampler when the sample rate
+        // or channel count actually changes; ending a chunk at every boundary
+        // would flag it `partial_len`, which rubato zero-pads.
+        while self.input_buffer.len() < wanted {
+            let span = self.input.current_span_len().map(InSamples);
+            let take = (wanted - self.input_buffer.len()).min(span.unwrap_or(InSamples::MAX));
+            if take == InSamples::ZERO {
+                break;
+            }
+
+            let rate_before = self.input.sample_rate();
+            let channels_before = self.input.channels();
+
+            let mut taken = 0usize;
+            for _ in 0..take.raw() {
+                if let Some(sample) = self.input.next() {
+                    self.input_buffer.push(sample);
+                    taken += 1;
+                } else {
+                    break;
+                }
+            }
+
+            if taken == 0
+                || self.input.sample_rate() != rate_before
+                || self.input.channels() != channels_before
+            {
                 break;
             }
         }
+
         self.input_buffer.len().frames(self.output.channels)
     }
 
--- a/src/conversions/sample_rate/tests.rs
+++ b/src/conversions/sample_rate/tests.rs
@@ -355,8 +355,8 @@ fn small_spans_do_not_degrade_resampled_signal() {
     // 0.5 s of a 440 Hz stereo sine at 48 kHz.
     let signal: Vec<Sample> = (0..24_000)
         .flat_map(|i| {
-            let t = i as f32 / source_rate.get() as f32;
-            let s = (std::f32::consts::TAU * 440.0 * t).sin() * 0.5;
+            let t = i as Sample / source_rate.get() as Sample;
+            let s = (std::f64::consts::TAU as Sample * 440.0 * t).sin() * 0.5;
             [s, s]
         })
         .collect();
--- a/src/conversions/sample_rate/tests.rs
+++ b/src/conversions/sample_rate/tests.rs
@@ -364,6 +364,9 @@ fn small_spans_do_not_degrade_resampled_signal() {
     // Fraction of output energy sitting at the 440 Hz fundamental, via
     // Goertzel on the left channel. A clean resample keeps this ~1.0.
     let tone_energy_fraction = |samples: &[Sample]| -> f64 {
+        // Accumulate in f64 regardless of `Sample`: the cast is a no-op
+        // under the `64bit` feature but needed without it.
+        #[allow(clippy::unnecessary_cast)]
         let left: Vec<f64> = samples.iter().step_by(2).map(|&s| s as f64).collect();
         let n = left.len();
         let bin = 440.0 * n as f64 / target_rate.get() as f64;
--- a/src/conversions/sample_rate/tests.rs
+++ b/src/conversions/sample_rate/tests.rs
@@ -338,3 +338,64 @@ fn test_span_boundary_same_format() {
         output.len()
     );
 }
+
+/// Resampling must not depend on how the input is chunked into spans.
+///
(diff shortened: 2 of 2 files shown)


yara-blue:
Closing as a low effort - likely automated - contribution. I urge you to be more respectful of open source maintainers time in the future.

benface:
Definitely automated @yara-blue, and I apologize for that. That being said, it fixes a real issue. Is the issue not clear, the solution heavy-handed, or something else?

roderickvd:
@benface we would welcome contributions that:

1. read succinctly
2. show understanding of the problem and solution
3. apply best practices

Number one I think is clear.

Yes I believe this solves a real issue, but it doesn't solve it fully. There's another pre-existing bug there layered on top with span position tracking - can you see it?

Further, Rodio already has signal generators that you could use to simplify and improve much needed readability of the test. Finally, that test does not show the case where span boundaries actually change channel count or sample rate.

I hope this helps guide your future contributions. Use AI as your tool but own the result yourself.

benface:
Appreciate it @roderickvd. I will try to find the time to do a proper fix.

PR #913, opened 2026-07-31
fix: don't re-invoke exhausted source during keep-alive silence

roderickvd:
`SourcesQueueOutput::next()` kept calling `current.next()` on the just-exhausted source every time the keep-alive silence ran out, instead of moving past it. Sources like `EmptyCallback` fired repeatedly for as long as the queue stayed empty, instead of once.

This PR tracks exhaustion explicitly and stops calling `next()` on current once observed, while still checking the queue every frame so a newly appended source starts with no added latency.

### DIFF
--- a/src/mixer.rs
+++ b/src/mixer.rs
@@ -1,7 +1,7 @@
 //! Mixer that plays multiple sounds at the same time.
 
 use crate::common::{ChannelCount, SampleRate};
-use crate::source::{SeekError, Source, UniformSourceIterator};
+use crate::source::{Empty, SeekError, Source, UniformSourceIterator};
 use crate::Sample;
 use std::sync::Arc;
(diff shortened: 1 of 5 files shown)


roderickvd:
Converting to draft because we're not done yet (see Discord: 100% busy-looping when no source has been added yet).

@yara-blue I haven't dived into the engine rewrite PRs yet; if we are to back port this (do we need to at all?) then in which "phase PR" should I be looking?

yara-blue:
> @yara-blue I haven't dived into the engine rewrite PRs yet; if we are to back port this (do we need to at all?) then in which "phase PR" should I be looking?

That should be Phase four. I'll make a note in the tracking issue to check for this since with the new traits I rewrote the queue mechanic (to prevent allocating on the audio thread).

PR #915, opened 2026-08-06
Don't let span chunking change the resampler's output

benface:
Resampler output changes depending on how the input happens to be split into spans, which it shouldn't.

`fill_input_buffer` stops at every span boundary and marks the read partial, so rubato zero-pads it as if the stream had ended. Decoders report a span per packet, so this happens constantly — a 48 kHz file on a 44.1 kHz device hisses the whole way through. It now reads across spans and only marks partial at a real end or a format change.

The resamplers were also tracking input position against span length to spot boundaries, but they read ahead into their own buffer, so that count never matched where the input actually was. Filling already stops at a format change, so the tracking is gone — the passthrough keeps its own, where it's accurate.

Tests cover the chunking, plus a sample rate change and a channel count change at a boundary. The chunking one fails on master.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed sources to detect parameter updates after mid-span seeks.
 - Fixed `Stoppable` and `Skippable` not signaling exhaustion.
 - Fixed `SpatialAudio` left and write channel swapping
+- Fixed resampled output depending on how the input is split into spans.
 
 ## Version [0.22.2] (2026-03-05)
 
--- a/src/conversions/sample_rate/mod.rs
+++ b/src/conversions/sample_rate/mod.rs
@@ -167,6 +167,28 @@ where
         }
     }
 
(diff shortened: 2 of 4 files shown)


phayes:
I found one issue on this:

in next(), you're returning `None` (end-of-stream) when `source.current_span_len()` is `None` (No span length means the format is stable by contract).   So there's some confusion between end-of-stream signalling and stable-format signalling. 

I've opened a PR to your PR with a simple fix. 

The fix boils down to this:

```rust
        // No span length means the format is stable by contract.
        // input_span_len?; // replace this with this:
        let Some(input_span_len) = input_span_len else {
            return Some(sample);
        };
```

phayes:
PR to your PR is here: https://github.com/benface/rodio/pull/1

benface:
Thank you so much @phayes! When I rewrote `next()` I turned `if input_span_len.is_none() { return Some(sample); }` into `input_span_len?;` and kept the comment saying "format is stable by contract", without noticing that the operator does the opposite. 😅 Good test too; it fails on the pre-fix commit. Merged your PR 🙏

PR #925, opened 2026-09-20
Disable log on drop by default

FireDasher:
Nobody wants this message. It's useless. You should not have to make your stream mutable just to disable this message. I've left the message in in case someone wanted it for debugging, but this should absolutely not be enabled by default, since the program will always close eventually, and it is quite obvious that audio being played by a program will stop when it's terminated.


roderickvd:
It's difficult to read from a screen, but your "Nobody wants this message. It's useless." comes across as rather harsh.

Have you read back in history or the documentation why it's there?

It's not necessary in the strictly procedural case of "create a stream, play something, sleep, terminate" like the simple terminal examples. In other cases, users do stuff like creating a `MyPlayer::new()` where they create a handle, add it to the mixer, then are surprised nothing plays because they forgot to store the _handle_ and not just the mixer in their struct.

We have put that message there exactly because it helps lots of users.

We would welcome ideas how to fix this without having to keep the handle around in the struct. Until then, this message stays.

FireDasher:
> but your "Nobody wants this message. It's useless." comes across as rather harsh.

Sorry also yeah you make a fair point

FireDasher:
Maybe you could replace the log_on_drop to instead take self directly and return self so you don't have to make the stream mutable just to disable the message if you didn't want it.

